PR #8, opened 2019-08-09
Return empty list for get_observations in BT path

shifucun:


store/pop_obs.go
```diff
@@ -286,13 +286,6 @@ func (s *store) btGetObservations(ctx context.Context, in *pb.GetObservationsReq
 		return err
 	}
 
-	// Iterate through all dcids to make sure they are present in the final result.
-	for _, dcid := range dcids {
```
Spaceenter: Similar code also exists for get_populations. 
But I guess this is probably less of a concern for get_populations.

shifucun: Get population requires exact match, so there should be okay..

PR #10, opened 2019-08-09
Convert observation stats type to snake case in BQ path

shifucun:


store/pop_obs.go
```diff
@@ -26,6 +26,7 @@ import (
 	"cloud.google.com/go/bigtable"
 	pb "github.com/datacommonsorg/mixer/proto"
 	"github.com/datacommonsorg/mixer/util"
+	"github.com/iancoleman/strcase"
```
Spaceenter: The concern for depend on this kind of library is that it belongs to a single person so it might not be reliable, s/he might delete/move this library in the future.

If have to use, usually I see people fork the library to own repo, then depend on that.

In this case, I feel we could just build our own toSnake and toCamel.

shifucun: Okay, sent a new PR that does not use 3p library

PR #14, opened 2019-08-09
Fixed bug in GetPopulations and GetObservations

ACscooter:
Cache results now match BQ if they exist.


store/pop_obs.go
```diff
@@ -127,18 +129,23 @@ func (s *store) btGetPopulations(ctx context.Context, in *pb.GetPopulationsReque
...
 			keySuffix += ("-" + p + "-" + v)
 		})
 	}
 
+	// Generate the list of all keys to query cache for
 	rowList := bigtable.RowList{}
 	for _, dcid := range dcids {
-		rowList = append(rowList, util.BtPopPrefix+dcid+keySuffix)
+		btKey := util.BtPopPrefix + dcid + keySuffix
+		log.Printf("GetPopulations: Querying cache with key \"%s\"", btKey)
```
Spaceenter: Maybe we want to avoid logs like this (in a potentially big for loop) so that we don't swarm the logs.

ACscooter: Removed logging in dfd1793

PR #22, opened 2019-08-13
Update out.pb with recent proto change and remove explicte gke cluster version

shifucun:


server/main.go
```diff
@@ -39,21 +39,24 @@ var (
 	port       = flag.String("port", ":12345", "Port on which to run the server.")
 )
 
+// Version is the release version number.
+var version = "head"
```
beets: Should this be 1.0.0 for now?

beets: oops. ignore this comment.

server/main.go
```diff
@@ -39,21 +39,24 @@ var (
 	port       = flag.String("port", ":12345", "Port on which to run the server.")
 )
 
+// Version is the release version number.
```
beets: please clarify which version this refers to (the API?) and a clear note to update this with each branch.

shifucun: Done

Spaceenter:
I think usually the version number is in the request (or even the request URL) rather than response.
https://cloud.google.com/apis/design/versioning
https://stackoverflow.com/questions/389169/best-practices-for-api-versioning

server/main.go
```diff
@@ -39,21 +39,25 @@ var (
 	port       = flag.String("port", ":12345", "Port on which to run the server.")
 )
 
+// Version is the docker image release version number.
+// This should be updated on the release branches whenever a new docker image is built and released.
```
beets: Hmm, isn't this the version number for the API that we manually configure (separate from the docker image, which is for the binary releases and should not effect the version number in the response).

beets:
> I think usually the version number is in the request (or even the request URL) rather than response.
> https://cloud.google.com/apis/design/versioning
> https://stackoverflow.com/questions/389169/best-practices-for-api-versioning
> 
> Personally I haven't seen this kind of pattern for versioning (meaning, return version from every REST response).
> @ACscooter @shifucun

Agree that it's nonstandard... but it was a workaround to avoid supporting backwards incompatible changes (still need a way to let clients react to changes in the response, but not have a contract on the server for being able to respond to past versions of the request).

server/main.go
shifucun: Updated

PR #51, opened 2019-08-19
Update BT table which includes 'location' and 'childhoodLocation' for…

Spaceenter:
… in arcs of places.


Spaceenter:
Deprecate this PR as there is newer version of BT.

PR #54, opened 2019-08-24
Switch to a new BT table with Observations being an independent cache.

Spaceenter:


store/core.go
```diff
@@ -614,31 +614,116 @@ func (s *store) GetPropertyLabels(
...
+			return err
+		}
+	}
+
+	objPlaceIDNameMap := map[string]string{}
+	if len(obsDcids) > 0 {
+		// Observation DCIDs.
+		obsRowList := bigtable.RowList{}
+		for _, dcid := range obsDcids {
+			for _, pred := range []string{"0", "1"} {
```
pradh: Comment what is "0" and "1"?

store/core.go
```diff
@@ -614,31 +614,116 @@ func (s *store) GetPropertyLabels(
...
-		}); err != nil {
-		return err
+		}
+		if err := bigTableReadRowsParallel(ctx, s.btTable, obsRowList,
+			func(btRow bigtable.Row) error {
+				// Extract DCID from row key.
+				parts := strings.Split(btRow.Key(), "-")
+				dcid := strings.TrimPrefix(parts[0], util.BtObsAncestorPrefix)
+				var pred string
+				if parts[1] == "0" {
```
ACscooter: Echoing Shanth's comment above, maybe change "0" and "1" to use constants

store/core.go
Spaceenter: Done.

store/core.go
Spaceenter: Done.

PR #66, opened 2019-08-30
Return empty list for query when there is no query result

shifucun:


examples/query/main.go
```diff
@@ -67,15 +67,8 @@ func main() {
...
-		WHERE {
-			?question typeOf SurveyItem .
-			?question name ?qcode .
-			?question hasResponseOption ?roption .
-			?roption typeOf ResponseOption .
-			?roption identifier "0"
-		}
-		LIMIT 10`,
+		SELECT ?a
+		WHERE { ?a typeof USC_RaceCodeEnum}`,
```
Spaceenter: Unnecessary empty space before "?a"?

PR #75, opened 2019-09-10
Change "-" to "," as connector for cache key components.

Spaceenter:


Spaceenter:
This won't be merged/submitted until the BT table is updated.

PR #79, opened 2019-09-13
Implement GetPopCategory method. Also fixing the string connector for…

Spaceenter:
… GetObsAncestor.


examples/run_all.sh
```diff
@@ -1,9 +1,10 @@
 #!/bin/bash
```
shifucun: can add "go run query/mani.go"

Spaceenter: Done.

PR #92, opened 2020-01-14
Add an argument to GetPropertyValuesRequest to distinguish in and out…

Spaceenter:
… arc.


proto/mixer.proto
```diff
@@ -94,6 +94,9 @@ message GetPropertyValuesRequest {
 
   // Maximum number of nodes to query for.
   int32 limit = 4;
+
+  // If choosing out arc (or in arc).
+  bool out_arc = 5;
```
pradh: If we use proto default, then it is in-arc. Do you want to make default as out? (cheaper, more natural, etc.)

Spaceenter: That's a good point. Changing.

PR #93, opened 2020-01-15
If GetPropertyValues::direction is neither 'in' nor 'out', return res…

Spaceenter:
…ults for both.

### DIFF
--- a/store/core.go
+++ b/store/core.go
@@ -86,21 +86,70 @@ func (s *store) GetPropertyValues(ctx context.Context,
 		in.Limit = util.BtCacheLimit
 	}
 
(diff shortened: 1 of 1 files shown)


store/core.go
```diff
@@ -86,21 +86,70 @@ func (s *store) GetPropertyValues(ctx context.Context,
...
+	case "in":
+		inArc = true
+		outArc = false
+	case "out":
+		inArc = false
+		outArc = true
+	default:
+		inArc = true
+		outArc = true
+	}
```
pradh: case "":
      inArc = true
      outArc = true
  default:
      return "Invalid value for 'direction'"

PR #95, opened 2020-01-21
Staging

hanlu09205:
Add support to deploy to staging cloud mixer
1) make bq_dataset, bt_table & bt_instance configurable from template_deployment
2) update gke.sh which would replace deployment with the right value for above arguments
2) depending on project_id, set different compute/zone
4) update readme 

Question, people may forget to update bt_table, which cause we lose track the current in-use bt_table. Is there a better way to achieve that?

### DIFF
--- a/deployment/template_deployment.yaml
+++ b/deployment/template_deployment.yaml
@@ -63,10 +63,10 @@ spec:
           image: IMAGE
           imagePullPolicy: Always
           args: [
-            "--bq_dataset", "google.com:datcom-store-dev.dc_v3_clustered",
+            "--bq_dataset", "BQ_DATASET",
             "--bt_table", "dc22",
             "--bt_project", "google.com:datcom-store-dev",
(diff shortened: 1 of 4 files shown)


hanlu09205:
https://screenshot.googleplex.com/RXMOswseQCO shows query result.

deployment/template_deployment.yaml
```diff
@@ -63,10 +63,11 @@ spec:
           image: IMAGE
           imagePullPolicy: Always
           args: [
-            "--bq_dataset", "google.com:datcom-store-dev.dc_v3_clustered",
-            "--bt_table", "dc22",
+            "--bq_dataset", "BQ_DATASET",
+            # Prod using dc22 and staging using dc1. Update this when bt_table change.
```
shifucun: Can you create a file to record the bt dataset. We only update that file (and that is gold standard) and here reads from that.

PR #107, opened 2020-02-04
Make getObservations return a dcid->value dict rather than an array.

mvashishtha:
Currently, getObservations returns a JSON array of Objects such as:
[{"dcid":"dc/p/053kf3t9s9bg","observation":"23.000000"}] 

This change makes the API return instead a dict like this:

{"dc/p/053kf3t9s9bg":"23.000000"}


shifucun:
Closing this as the API has been deprecated

PR #114, opened 2020-02-18
Update prod bigtable to dc32

shifucun:


Spaceenter:
This probably can be deprecated.

PR #117, opened 2020-02-21
make bq_dataset configurable through bq_dataset.txt

hanlu09205:
Add bq_dataset.txt for both staging and prod. 
Udpate gke.sh to take in bq_dataset from the test file. 
Udpate README

### DIFF
--- a/deployment/README.md
+++ b/deployment/README.md
@@ -19,6 +19,7 @@ To build a new image, run the command below. Tag number should be larger than an
 gcloud builds submit --tag gcr.io/datcom-mixer/go-grpc-mixer:<TAG> 
 
 If you need to point to a different Bigtable table or BigQuery dataset, update the corresponding bt_table.txt or bq_dataset.txt.
+Both of them have to version, prefixed with prod_ or staging_.
 
 To deploy the project, run the command below. 
 ```shell
--- a/deployment/README.md
+++ b/deployment/README.md
@@ -18,6 +18,8 @@ Create a Google Cloud Platform (GCP) project and run the following command where
 To build a new image, run the command below. Tag number should be larger than any existing one. 
 gcloud builds submit --tag gcr.io/datcom-mixer/go-grpc-mixer:<TAG> 
 
+If you need to point to a different Bigtable table or BigQuery dataset, update the corresponding bt_table.txt or bq_dataset.txt.
+
 To deploy the project, run the command below. 
 ```shell
 ./gcp.sh {PROJECT_ID}
--- a/deployment/gke.sh
+++ b/deployment/gke.sh
@@ -51,11 +51,16 @@ else
 fi
 
 # Set BQ dataset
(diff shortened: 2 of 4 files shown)


deployment/README.md
```diff
@@ -18,6 +18,9 @@ Create a Google Cloud Platform (GCP) project and run the following command where
 To build a new image, run the command below. Tag number should be larger than any existing one. 
 gcloud builds submit --tag gcr.io/datcom-mixer/go-grpc-mixer:<TAG> 
 
+If you need to point to a different Bigtable table or BigQuery dataset, update the corresponding bt_table.txt or bq_dataset.txt.
+Both of them have to version, prefixed with prod_ or staging_.
```
pradh: have two versions

deployment/gke.sh
```diff
@@ -51,11 +51,16 @@ else
 fi
 
 # Set BQ dataset
-if [ "$PROJECT_ID" == "datcom-mixer-staging" ]; then
-  perl -i -pe's/BQ_DATASET/google.com:datcom-store-dev.dc_v3_staging_clustered/g' deployment.yaml
-else 
-  perl -i -pe's/BQ_DATASET/google.com:datcom-store-dev.dc_v3_clustered/g' deployment.yaml
+if [ "$PROJECT_ID" == "datcom-mixer" ]; then
+  bq_dataset_input_file="./prod_bq_dataset.txt"
```
pradh: ./ is redundant?

deployment/gke.sh
```diff
@@ -51,11 +51,16 @@ else
...
+else
+  bq_dataset_input_file="./staging_bq_dataset.txt"
 fi
+while read -r line; do
+  BQ_DATASET=$line
+done < "$bq_dataset_input_file"
```
pradh: why not?

BQ_DATASET=$(cat $bq_dataset_input_file)

deployment/gke.sh
```diff
@@ -51,11 +51,16 @@ else
...
+export BQ_DATASET
+perl -i -pe's/BQ_DATASET/$ENV{BQ_DATASET}/g' deployment.yaml
```
pradh: I'm curious why perl was used, long time no see :)

this is fine for consistency, but an alternative would be to do:

    sed -i "s/BQ_DATASET/$BQ_DATASET/g" deployment.yaml

which doesn't need "export BQ_DATASET".

hanlu09205: I got sed: 1: "deployment.yaml": extra characters at the end of d command
Maybe if we use sed, different OS will respond different, vs perl.

deployment/gke.sh
hanlu09205: Thanks for the suggestion.

hanlu09205: I also updated how BT_TABLE get assigned.

hanlu09205:
PTAL

PR #126, opened 2020-03-14
Read cached data from gcs

shifucun:


deployment/gke.sh
```diff
@@ -67,6 +67,14 @@ perl -i -pe's/BQ_DATASET/$ENV{BQ_DATASET}/g' versioned_mapping/base.mcf
 perl -i -pe's/BQ_DATASET/$ENV{BQ_DATASET}/g' versioned_mapping/weather.mcf
 
 
+# Set GCS bucket and in memory mcf name
```
pradh: is it in memory mcf?

shifucun: updated comment

PR #128, opened 2020-03-14
Merge popobs result

shifucun:


store/populations.go
```diff
@@ -39,18 +40,44 @@ type PopObs struct {
...
+		}
 	}
-	return nil
+	resStr, err := (&jsonpb.Marshaler{}).MarshalToString(&res1)
+	if err != nil {
+		return err
+	}
+	out.Payload, err = util.ZipAndEncode(resStr)
+	return err
 }
```
pradh: If the side-cache is empty do not unmarshall and remarshall?

store/populations.go
```diff
@@ -39,18 +40,44 @@ type PopObs struct {
...
-		out.Payload, _ = util.ZipAndEncode("{}")
+		c, err := util.UnzipAndDecode(string(btRow[util.BtFamily][0].Value))
+		if err != nil {
+			log.Print(err)
+		}
+		jsonpb.UnmarshalString(string(c), &res2)
+
+		for k, v := range res2.Populations {
+			res1.Populations[k] = v
+		}
```
pradh: To clarify, if there is an overlapping pop-id, here, we are picking the BT one?

I wonder if we should pick the side-cache?

store/populations.go
shifucun: refactored to only unzip and zip when both cache exist

store/populations.go
shifucun: Changed so side cache has higher priority

PR #132, opened 2020-04-01
Cache merge for get_place_obs

shifucun:


store/populations.go
```diff
@@ -97,19 +97,58 @@ func (s *store) GetPlaceObs(ctx context.Context, in *pb.GetPlaceObsRequest,
...
 	} else {
-		out.Payload, _ = util.ZipAndEncode("{}")
+		if tmp, err := util.UnzipAndDecode(baseString); err == nil {
+			jsonpb.UnmarshalString(string(tmp), &baseData)
+		}
+		if tmp, err := util.UnzipAndDecode(sideString); err == nil {
+			jsonpb.UnmarshalString(string(tmp), &sideData)
+		}
+		dataMap := map[string]*pb.PopObsPlace{}
+		for _, data := range baseData.Places {
```
pradh, line 135: For a given place, do we for now don't merge the observations? e.g., delta cache has 2018, but base has 2012 to 2017, will it work?

shifucun: place obs has "observationDate" is argument, so no need to merge by years.

In general the "branch cache" has the full dataset, so we can almost always overwrite the base cache, in time, stats_var dimensions.

PR #133, opened 2020-04-02
Update bigtable test support and start to add test for get_populations

shifucun:


go.mod
```diff
@@ -7,8 +7,8 @@ require (
 	cloud.google.com/go/bigquery v1.0.1
 	cloud.google.com/go/bigtable v1.0.0
 	github.com/go-test/deep v1.0.3
-	github.com/golang/protobuf v1.3.4
-	github.com/google/go-cmp v0.3.1 // indirect
+	github.com/golang/protobuf v1.4.0-rc.4
```
hanlu09205, line 10: For my education, when do we update those and where to find the right version?

store/core_test.go
```diff
@@ -36,11 +36,11 @@ func TestGetPropertyLabels(t *testing.T) {
 	} {
 		jsonRaw, err := json.Marshal(d.labels)
 		if err != nil {
-			t.Fatalf("json.Marshal(%v) = %v", d.dcid, err)
+			t.Errorf("json.Marshal(%+v) = %+v", d.dcid, err)
```
hanlu09205, line 39: Nit, for err, I think %v is good enough.

store/populations_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/datacommonsorg/mixer/util"
+)
+
+func TestGetPopObs(t *testing.T) {
+	data := map[string]string{}
+	dcid := "geoId/06"
+	key := util.BtPopObsPrefix + dcid
+	btRow := `{
+	"name": "Santa Clara",
+  "populations": {
```
hanlu09205, line 31: Nit, is there any tool we could use to format those?

store/populations_test.go
```diff
@@ -0,0 +1,96 @@
...
+      "meanValue": 9.4461,
+      "measuredProp": "particulateMatter25",
+      "measurementMethod": "CDCHealthTracking",
+      "observationDate": "2014-03-20",
+      "observedNode": "geoId/06085"
+    }
+  ]
+}`
+
+	tableValue, err := util.ZipAndEncode(string(btRow))
```
hanlu09205, line 76: So most of this is copying from g3 to github ? 

Lazy question, how to do run build and test on github?

go.mod
shifucun: This is automatically updated when I run "go test" locally. I think there is a way to force update this. Can check with Wei. In general, don't really need to update the package.

store/populations_test.go
shifucun: I use visual studio code and golang formatter plugin. It usually does good job. In this case as it is a string, so seems dumb. I updated manually.

go.mod
hanlu09205: Got it. Thanks

store/populations_test.go
shifucun: Yes. We setup a cloud trigger to run the test. If you scroll down and see there are "All checks have passed", the first one will run all the tests.

PR #137, opened 2020-04-06
Add Cloud PubSub subscriber in mixer to reload cache when notified

shifucun:


store/store.go
```diff
@@ -142,6 +163,16 @@ func NewStore(
 			cacheData[parts[0]] = parts[1]
 		}
 	}
+	st.cacheData = cacheData
```
pradh, line 166: Should writes to cacheData be synchronized with accesses with a mutex or something?

(Sorry for the delayed response.)

store/store.go
```diff
@@ -175,12 +206,50 @@ func NewStore(
...
+				log.Printf("Load cache data got error %s", err)
+			}
+		})
+		if err != nil {
+			log.Fatalf("Cloud pubsub receive: %v", err)
+		}
+	}()
+
+	// Initial branch cachel load.
+	err = st.LoadBranchCache(ctx, gcsFolder)
```
pradh, line 243: What is gcsFolder at init time - the one set by flag? That could be obsolete. Should we do the "latest_branch_cache_version" file read?

shifucun: I will replace here by reading the  "latest_branch_cache_version" file when it's ready. This needs to be added in the import controller?

pradh: Yes.

store/store.go
shifucun: Use atomic.StorePointer to assign the new cache into store

store/cache.go
```diff
@@ -0,0 +1,48 @@
...
+// NewCache returns a new Cache instance.
+func NewCache() *Cache {
+	cache := Cache{}
+	cache.data = map[string]string{}
+	return &cache
+}
+
+// Read is used for key value look up for Cache.
+func (c *Cache) Read(key string) (string, bool) {
+	c.Lock()
```
pradh, line 37: RLock()?

store/cache.go
```diff
@@ -0,0 +1,48 @@
...
+	defer c.Unlock()
+	value, ok := c.data[key]
```
pradh, line 39: Any reason to not just do:

  return c.data[key]

shifucun: That gives syntax error..

store/cache.go
shifucun: Done

PR #140, opened 2020-04-08
Read latest branch cache version from gcs

shifucun:


store/store.go
```diff
@@ -235,12 +236,26 @@ func NewStore(
...
 
 	// Initial branch cachel load.
-	err = st.LoadBranchCache(ctx, gcsFolder)
+	gcsClient, err := storage.NewClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+	rc, err := gcsClient.Bucket(gcsBucket).Object(versionFile).NewReader(ctx)
+	if err != nil {
+		log.Fatalf("%s", err)
```
pradh, line 250: Are you sure we want to crash here (and below)? If so, crash in 246 too? (for consistency)

shifucun: Also made it Fatal. This is during server start, can crash here to make sure cache always loads

PR #142, opened 2020-04-09
Add Option to API request and allows query for base cache only

shifucun:


proto/mixer.proto
```diff
@@ -73,6 +80,8 @@ message TranslateResponse {
 message GetPropertyLabelsRequest {
   // The dcids of nodes.
   repeated string dcids = 1;
+
+  Option.Enum option = 2;
```
pradh, line 84: This is OK. But might be nicer to have option be Option. That way, when we add a new option, you will not need to change anything existing. Like below:

   message Option {
     enum CacheChoice {
        ...
      };
      optional CacheChoice cache_choice = 1;
   };


   message GetPropertyLabelsRequest {
      ...
     Option option = 2;
   }

PR #152, opened 2020-04-14
Add a new API GetStats to get stats value for places given statistical variable

shifucun:


store/populations.go
```diff
@@ -471,3 +473,147 @@ func iterateSortPVs(pvs []*pb.PropertyValue, action func(i int, p, v string)) {
...
+			continue
+		}
+		if obs.MeasuredProp != mprop {
+			continue
+		}
+		msg := proto.MessageReflect(obs)
+		fd := msg.Descriptor().Fields().ByJSONName(statType)
+		if msg.Has(fd) {
+			ts.Data[obs.ObservationDate] = msg.Get(fd).Float()
+		}
```
pradh, line 505: Does this work? I thought the current ObsTimeSeries proto does not have these fields. Instead mmethod, mprop are part of the key.

shifucun: As discussed, I added all the observation fields here to check. Also using the StatisticalVariable proto here.

store/populations.go
```diff
@@ -471,3 +473,163 @@ func iterateSortPVs(pvs []*pb.PropertyValue, action func(i int, p, v string)) {
 		action(i, p, pvMap[p])
 	}
 }
+
+func addObsSeries(
+	key string,
+	cacheValue string,
+	result map[string]pb.ObsTimeSeries,
```
Spaceenter, line 480: result map[string]*pb.ObsTimeSeries,
use pointer

store/populations.go
```diff
@@ -471,3 +473,163 @@ func iterateSortPVs(pvs []*pb.PropertyValue, action func(i int, p, v string)) {
...
+	statsVar pb.StatisticalVariable,
```
Spaceenter, line 481: statsVar *pb.StatisticalVariable

use pointer

store/populations.go
```diff
@@ -471,3 +473,163 @@ func iterateSortPVs(pvs []*pb.PropertyValue, action func(i int, p, v string)) {
...
+) error {
+	parts := strings.Split(key, "^")
+	dcid := strings.TrimPrefix(parts[0], util.BtObsSeriesPrefix)
+	val, err := util.UnzipAndDecode(cacheValue)
+	if err != nil {
+		return err
+	}
+	pbData := pb.PopObsPlace{}
```
Spaceenter, line 489: pbData := &pb.PopObsPlace{}
	jsonpb.UnmarshalString(string(val), pbData)

shifucun: Done

store/populations.go
shifucun: Done

PR #158, opened 2020-04-21
Add dataflow template to ingest CSV data.

sushant2603:
Dataflow template for ingesting data from GCS to cloud  BT.

This will run as  part of airflow template which will be sent in a separate PR.

The airflow template is to be triggered by cloud function after flume job is finished. Cloud function is coming up in a separate PR


sushant2603:
@pradh @Spaceenter : Can you take a look at this

cloud_automation/java/dataflow/src/main/java/org/datacommons/dataflow/CsvImport.java
```diff
@@ -0,0 +1,121 @@
...
+    @Description("The Cloud Storage path to the CSV file.")
+    ValueProvider<String> getInputFile();
+
+    @SuppressWarnings("unused")
+    void setInputFile(ValueProvider<String> location);
+  }
+
+
+  /**
+   * <p>Creates a dataflow templae with following stages: </p>
```
Spaceenter, line 81: s/templae/template

cloud_automation/java/dataflow/README.md
```diff
@@ -0,0 +1,13 @@
...
+The CSVImport dataflow template creates dataflow jobs to ingest CSV data from
+GCS to Cloud Bigtable.
+
+The template is stored at  gs://datcom-dataflow-templates/templates
+
+To deploy changes to the template run:
+
+```
+  mvn compile exec:java -Dexec.mainClass=org.datacommons.dataflow.CsvImport -Dexec.args="--runner=DataflowRunner --project=google.com:datcom-store-dev --stagingLocation=gs://datcom-dataflow-templates/staging --templateLocation=gs://datcom-dataflow-templates/templates/csv_to_bt"
+
```
pradh, line 12: What does --stagingLocation mean?

Can we call it "csv_to_bt_staging"? (just to make it clear)

cloud_automation/java/dataflow/src/main/java/org/datacommons/dataflow/CloudBigtableOptions.java
```diff
@@ -0,0 +1,26 @@
...
+
+import org.apache.beam.runners.dataflow.options.DataflowPipelineOptions;
+import org.apache.beam.sdk.options.Description;
+import org.apache.beam.sdk.options.ValueProvider;
+
+public interface CloudBigtableOptions extends DataflowPipelineOptions {
+  @Description("The Google Cloud project ID for the Cloud Bigtable instance.")
+  ValueProvider<String> getBigtableProjectId();
+
+  @SuppressWarnings("unused")
```
Spaceenter, line 11: Do we need these unused methods in the future?

cloud_automation/java/dataflow/src/main/java/org/datacommons/dataflow/CsvImport.java
```diff
@@ -0,0 +1,121 @@
...
+    BigtableIO.Write write = BigtableIO.write()
+            .withProjectId(options.getBigtableProjectId())
+            .withInstanceId(options.getBigtableInstanceId())
+            .withTableId(options.getBigtableTableId());
+    p.apply("ReadMyFile", TextIO.read().from(options.getInputFile()))
+        .apply("TransformParsingsToBigtable", ParDo.of(new CsvToBigtableFn()))
+        .apply("WriteToBigtable", write);
+
+    PipelineResult result = p.run();
+            // Wait for pipeline to finish only if it is not constructing a template.
```
pradh, line 116: Odd indent.

cloud_automation/java/dataflow/README.md
sushant2603: staginglocation is the gcs directory where additional files required for running this template are stored. This is part of the dataflow custom template deployer and not out template code. I think it is best to leave it as-is, since it would require changing the dataflow template deployer,

cloud_automation/java/dataflow/src/main/java/org/datacommons/dataflow/CsvImport.java
```diff
@@ -0,0 +1,121 @@
+package org.datacommons.dataflow;
```
pradh, line 1: Had the same question as Wei, this file seems to differ in some ways to https://github.com/GoogleCloudPlatform/cloud-bigtable-examples/blob/master/java/dataflow-connector-examples/src/main/java/com/google/cloud/bigtable/dataflow/example/CsvImport.java. What are the differences?

cloud_automation/java/dataflow/src/main/java/org/datacommons/dataflow/CsvImport.java
```diff
@@ -0,0 +1,121 @@
...
+   *  CloudBigtableIO libraries do not completely support dataflow templates.
+   *  The inspiration for this is pipeline is from cloud google ataflow examples.
+   */
+
+  public static void main(String[] args) throws IllegalArgumentException {
+    BigtableCsvOptions options =
+        PipelineOptionsFactory.fromArgs(args).withValidation().as(BigtableCsvOptions.class);
+
+    Pipeline p = Pipeline.create(options);
+
```
pradh, line 106: I'm comparing to the original CsvImport.java. should we retain the options checks?

    if (options.getInputFile().equals("")) {
      throw new IllegalArgumentException("Please provide value for inputFile.");
    }
    if (options.getHeaders().equals("")) {
      throw new IllegalArgumentException("Please provide value for headers.");
    }

cloud_automation/java/dataflow/src/main/java/org/datacommons/dataflow/CsvImport.java
```diff
@@ -0,0 +1,121 @@
...
+            .withProjectId(options.getBigtableProjectId())
+            .withInstanceId(options.getBigtableInstanceId())
+            .withTableId(options.getBigtableTableId());
+    p.apply("ReadMyFile", TextIO.read().from(options.getInputFile()))
+        .apply("TransformParsingsToBigtable", ParDo.of(new CsvToBigtableFn()))
+        .apply("WriteToBigtable", write);
+
+    PipelineResult result = p.run();
+            // Wait for pipeline to finish only if it is not constructing a template.
+    if (options.as(DataflowPipelineOptions.class).getTemplateLocation() == null) {
```
pradh, line 117: For my understanding, what will poll the job running when this is fired off as template? (Sorry if you have this in a doc, I can read that too.)

cloud_automation/java/dataflow/src/main/java/org/datacommons/dataflow/CsvImport.java
```diff
@@ -0,0 +1,121 @@
...
+   * <p>Creates a dataflow templae with following stages: </p>
+   * <ol>
+   * <li> Read a CSV from GCS.
+   * <li> Transform each line to a Bigtable Row mutation.
+   * <li> Write to Bigtable.
+   * </ol>
+   *
+   * @param args Arguments to use to configure the Dataflow Pipeline. Check README for deployment
+   * instructions.
+   * --project=[dataflow project] \\ --stagingLocation=gs://[your google storage bucket] \\
```
pradh, line 90: Are these \\ intentional?

cloud_automation/java/dataflow/src/main/java/org/datacommons/dataflow/CsvImport.java
```diff
@@ -0,0 +1,121 @@
...
+
+  private static final byte[] FAMILY = Bytes.toBytes("csv");
+  private static final Logger LOG = LoggerFactory.getLogger(CsvImport.class);
+
+  static class CsvToBigtableFn
+      extends DoFn<String, KV<ByteString, Iterable<Mutation>>> {
+    @ProcessElement
+    public void processElement(ProcessContext c) throws Exception {
+      try {
+        // TODO: Pass header as congfiguration.
```
pradh, line 37: To confirm, our CSV doesn't have header, does it?

cloud_automation/java/dataflow/src/main/java/org/datacommons/dataflow/CsvImport.java
```diff
@@ -0,0 +1,121 @@
...
+        for (int i = 0; i < headers.length; i++) {
+          Mutation.SetCell setCell = null;
+          headerBytes[i] = Bytes.toBytes(headers[i]);
+          setCell =  Mutation.SetCell.newBuilder()
+                          .setFamilyName(FAMILY)
+                          .setColumnQualifier(ByteString.copyFrom( headerBytes[i]))
+                          .setValue(ByteString.copyFrom(Bytes.toBytes(values[i])))
+                          .build();
+           mutations.add(Mutation.newBuilder().setSetCell(setCell).build());
+        }
```
pradh, line 54: Do we need to set "key" as a column if it is anyway going to be in row-key?

(sorry never really read this code before)

cloud_automation/java/dataflow/README.md
pradh: Oh I see. I misunderstood it to mean an equivalent of "staging instance".

cloud_automation/java/dataflow/src/main/java/org/datacommons/dataflow/CsvImport.java
pradh: Per offline clarification, this sets up dataflow template, and needs to use a different BT API.

cloud_automation/java/dataflow/src/main/java/org/datacommons/dataflow/CsvImport.java
sushant2603: No itdoes not

cloud_automation/java/dataflow/src/main/java/org/datacommons/dataflow/CloudBigtableOptions.java
sushant2603: Yes they are used by the templating framework to set value of parameter

cloud_automation/java/dataflow/src/main/java/org/datacommons/dataflow/CsvImport.java
sushant2603: Fixed

cloud_automation/java/dataflow/src/main/java/org/datacommons/dataflow/CsvImport.java
sushant2603: fixed

cloud_automation/java/dataflow/src/main/java/org/datacommons/dataflow/CsvImport.java
sushant2603: I wrote a airflow job and this template runs as part of the airflow job.
once the template(dataflow) jo completes, airflow writes to GCS where it suceeded or failed. A GCF function then triggers of this file and scales down bigtable.
Any more reports/emails we need can be triggered.

cloud_automation/java/dataflow/src/main/java/org/datacommons/dataflow/CsvImport.java
sushant2603: Can I do this in a follow-up. 
Doing this gives warnings in the java code, there is something about the way templates get parameters that causes the warnings. I want to be sure I understand the warning before doing this. This is why I removed this.

cloud_automation/java/dataflow/src/main/java/org/datacommons/dataflow/CsvImport.java
pradh: Thanks. Might be worth documenting the final flow in the doc you had or in a README.

cloud_automation/java/dataflow/src/main/java/org/datacommons/dataflow/CsvImport.java
sushant2603: I was not sure about the downstream impact of this, so I tried to keep the BT schema same. @Spaceenter do you know if this would work?

cloud_automation/java/dataflow/src/main/java/org/datacommons/dataflow/CsvImport.java
sushant2603: yes I will update the document and readme as well.
Sorry about this going back-and-forth. Airflow is the cleanest way to report back. I thought of polling as well.

cloud_automation/java/dataflow/src/main/java/org/datacommons/dataflow/CsvImport.java
pradh: In any case, this is for a potential follow-up change.

Feel free to submit this.

PR #163, opened 2020-04-24
Fix extra column being written to BT.

sushant2603:


Spaceenter:
Curious do we want to add /java/dataflow/target/... to gitignore? @sushant2603

PR #166, opened 2020-04-24
Exclude 'name' from constraining properties

shifucun:


store/populations.go
```diff
@@ -549,6 +549,8 @@ func (s *store) GetStats(ctx context.Context, in *pb.GetStatsRequest,
 			statsVar.StatType = t.ObjectID
 		} else if t.Predicate == "provenance" {
 			continue
+		} else if t.Predicate == "name" {
```
pradh, line 552: Just a matter of time before we could trip on other PVs, safe to exclude these as well?

http://google3/datacommons/util/dcid.cc?l=111-130&rcl=307964029

Another PR is ok.

shifucun: sg

PR #168, opened 2020-04-26
Fix get_stats branch cache cocurrent map write bug

shifucun:


examples/get_stats/main.go
```diff
@@ -53,4 +55,29 @@ func main() {
 	}
 	log.Printf("Now printing result:\n")
 	log.Printf("%s", r.GetPayload())
+
+	// Large payload
+	file, err := os.Open("geos.txt")
```
pradh, line 60: Do you need to check this in? Or have a script to generate it, or something.

shifucun: In this cl. The github ui is hard to use...

PR #172, opened 2020-04-26
Fully parallel the processing in GetStats

shifucun:


store/populations.go
```diff
@@ -581,44 +589,66 @@ func (s *store) GetStats(ctx context.Context, in *pb.GetStatsRequest,
...
-				if err != nil {
-					log.Println(err)
-				}
+				dcid := keyToDcid(rowKey, util.BtObsSeriesPrefix)
+				dcidToRaw[dcid] = branchString
 			}
 		}
 	}
 
+	// Read result from base cache if no base cache data found
```
Spaceenter, line 604: if no base cache data found --> if no branch cache data found

shifucun: done

PR #184, opened 2020-05-05
Update staging_bq_dataset.txt

sushant2603:


sushant2603:
@shifucun : Done, updated prod files here as well.

PR #200, opened 2020-06-08
Exclude censusTableId from constraining properties in get_stats_var

shifucun:


store/populations.go
```diff
@@ -592,6 +594,7 @@ func (s *store) GetStats(ctx context.Context, in *pb.GetStatsRequest,
 			keySuffix += "^" + p + "^" + v
 		})
 	}
+	log.Println(keySuffix)
```
pradh, line 597: Intend to keep it?

shifucun: Yes, keep it for easier debugging before we fix the issue.

PR #203, opened 2020-06-15
Check for constrainingProperties in stats_var triples and exclude it …

shifucun:
…from pvs

### DIFF
--- a/store/populations.go
+++ b/store/populations.go
@@ -564,6 +564,8 @@ func (s *store) GetStats(ctx context.Context, in *pb.GetStatsRequest,
 			continue
 		} else if t.Predicate == "censusACSTableId" {
 			continue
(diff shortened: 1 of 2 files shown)


store/populations.go
```diff
@@ -564,6 +564,8 @@ func (s *store) GetStats(ctx context.Context, in *pb.GetStatsRequest,
 			continue
 		} else if t.Predicate == "censusACSTableId" {
 			continue
+		} else if t.Predicate == "constrainingProperties" {
```
mvashishtha, line 567: This should be "constraintProperties". The bug says "constrainingProperties", but I re-used the existing property  "constraintProperties".

shifucun: Okay, I updated the bug title as well :)

PR #204, opened 2020-06-16
Use ChartStore cache data for GetStats API. The time series data is r…

shifucun:
…anked by import name and measurement method


store/populations.go
```diff
@@ -28,11 +28,26 @@ import (
...
+	mmethod string
+}
+
+var statsRanking = map[rankKey]int{
+	rankKey{"CensusPEP", "CensusPEPSurvey"}:                   0,
+	rankKey{"CensusACS5YearSurvey", "CensusACS5yrSurvey"}:     1,
+	rankKey{"EurostatData", "EurostatRegionalPopulationData"}: 2,
+	rankKey{"WorldDevelopmentIndicators", ""}:                 3,
+	rankKey{"BLS_LAUS", "BLSSeasonallyUnadjusted"}:            0,
+	rankKey{"EurostatData", ""}:                               1,
```
pradh, line 46: Add some comments on what the ranking means?

Also, add inline comments to demarcate
  // Population
  // Unemployment Rate
?

store/populations.go
```diff
@@ -493,38 +508,32 @@ func getObsSeries(
...
+		result := &pb.ObsTimeSeries{}
+		result.Unit = x.ObsTimeSeries.GetUnit()
+		result.PlaceName = x.ObsTimeSeries.GetPlaceName()
+		result.IsDcAggregate = x.ObsTimeSeries.GetIsDcAggregate()
+		bestScore := lowestRank
+		for _, series := range x.ObsTimeSeries.SourceSeries {
+			key := rankKey{series.GetImportName(), series.GetMeasurementMethod()}
+			score, ok := statsRanking[key]
+			if !ok {
+				score = lowestRank
```
pradh, line 524: nit: If not found, should we consider it having the highestRank, in that it will always be shown :)

pradh: Hmm, never mind. This is the correct thing to do when you have new sources for the same statvar.

store/populations.go
shifucun: Done

PR #208, opened 2020-06-20
staging release for 2020_06_19_01_03_57

hanlu09205:


hanlu09205:
I am worried that if prod is updated and cl submitted and people accidentally restart mixer prod with the staging not signed off version. 
I think two cls, one for staging one for prod is the most safe version, though it does add some work.

shifucun:
I think it's either way. For example, if the mixer code and image is updated, then accidentally restart prod mixer with new binary + old data would also have issue. I feel have one git commit (as a snapshot), which corresponds to a g3 cl that builds a mpm package is okay.

hanlu09205:
Hmm, that sounds a potential problem. 
If there is a way to tag release like we do on g3, and certain flag is tagged with release, then we could prevent this kind of error. 
Don't know whether github has the same feature though. 
I am submitting this for now.

PR #213, opened 2020-06-23
Increase mixer cluster memory to 78G

shifucun:
This is maximum memory for 4cpu configuration.


shifucun:
Thanks for the prompt review!

PR #219, opened 2020-06-28
Refactor GetStats function and server. Add e2e test against staging storage!

shifucun:
- GetStats read BT row and decompress in one goroutine.
- Add e2e test framework to test api agains staging BT data and compared with golden data.
- Start to modularize functions and add unittest.


golden_response/staging/get_stats/TotalCrimes.json
```diff
@@ -0,0 +1,34 @@
+{
```
hanlu09205, line 1: Is this based on some version of data or it's based on test data?

main.go
```diff
@@ -0,0 +1,46 @@
+// Copyright 2019 Google LLC
```
hanlu09205, line 1: Does github have anything similar to g4 branch?
Looks like this is branched from server/main.go?
Similar comments to other branched files. 

And we are moving these files one level up? I am not clear the reason behind though.

golden_response/staging/get_stats/TotalCrimes.json
shifucun: This is actual data from current version.

main.go
shifucun: i don't think there is a branch, it can only show diff if it's a "move".

The reason is to have the server folder to hold the server.go as server package. Can not have main.go in there anymore since it has to be a main package as program entrance.

golden_response/staging/get_stats/TotalCrimes.json
hanlu09205: I c. Not sure how often it changes, otherwise, might be to craft some test data.

shifucun: This is indeed a prober golden test against the staging environment. This is to choose the less changed yearly updated data.

shifucun:
> I didn't go into a lot details, wonder whether you want to wait for other LGTM, or just iterate over this. LGTM from me.

Thanks! I will let others take a look too since this has implication on the data release.

e2etest/base.mcf
```diff
@@ -0,0 +1,252 @@
+Database: google.com:datcom-store-dev.dc_v3_clustered
```
pradh, line 1: Is this a copy of an existing mapping MCF? Can we reuse that?

e2etest/get_stats_test.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+// This test runs agains staging staging bt and bq dataset.
+// This is billed to GCP project "datcom-ci"
+// It needs Application Default Credentials to run locally or need to
+// provide service account credential when running on GCP.
+const (
+	btProject   = "google.com:datcom-store-dev"
+	btInstance  = "prophet-cache"
+	bqProject   = "datcom-ci"
```
pradh, line 25: Are we putting the BQ tables now in a different project?

e2etest/get_stats_test.go
```diff
@@ -0,0 +1,142 @@
+package e2etest
```
pradh, line 1: Is there any automated way of running this?

golden_response/staging/get_stats/TotalCrimes.json
pradh: How is this test different from the staging prober? Will we now have to maintain/update one more set of golden files?

If the intent is end-to-end test of mixer, should we do something less real with fakes?

e2etest/get_stats_test.go
shifucun: This is actually the account the query is billed: https://godoc.org/cloud.google.com/go/bigquery#NewClient

e2etest/get_stats_test.go
shifucun: This can be tested via "go test". See the presubmit for this PR: https://screenshot.googleplex.com/CEeNOtwKQqg

golden_response/staging/get_stats/TotalCrimes.json
shifucun: The staging prober and this test will share exact the same queries and golden response.

This one runs as presubmit check and enforce code change to update the golden files before sumission

e2etest/base.mcf
shifucun: The existing mapping is a template of the BQ dataset. But since this does not test BQ query and probably not a good idea to test against real BQ, I just use the template mapping file. Removed the two files.

e2etest/get_stats_test.go
pradh: Oh nice, so this will run as part of presubmit?

shifucun: Yep.

PR #229, opened 2020-07-08
Use new ChartStore cache (without mmethod in the key) for Chart API

shifucun:


shifucun:
> After this rolls out can you please sync with Wei to rollback cl/319896945, so we can stop generating this cache?

Sg. I will sync after rollout

PR #232, opened 2020-07-08
Update BT/BQ paths.

Spaceenter:


deployment/staging_bq_dataset.txt
```diff
@@ -1 +1 @@
-google.com:datcom-store-dev.dc_kg_2020_06_30_01_02_51
\ No newline at end of file
+google.com:datcom-store-dev:dc_kg_2020_07_07_19_47_23
```
shifucun, line 1: the second colon should be "."

tjann: String check here? ;)

shifucun: good point. This will fail the e2e sparql test as well

shifucun:
Can you also update the image to "gcr.io/datcom-mixer/go-grpc-mixer:0.10.0"

PR #233, opened 2020-07-08
Update GetRelatedPlaces API to accomadate to data change

shifucun:
- Remove mmethod, scaling factor and unit from request


proto/mixer.proto
```diff
@@ -314,21 +314,12 @@ message GetRelatedPlacesRequest {
...
   string measurement_qualifier = 7;
 
-  // (Opitonal) Scaling factor.
-  string scaling_factor = 8;
-
-  // (Optional) Unit.
-  string unit = 9;
-
   // The statistics type, like "measured", "median", etc...
   string stat_type = 10;
```
Spaceenter, line 325: add 

reserved 5, 8, 9;

?

e2etest/get_related_places_test.go
```diff
@@ -0,0 +1,131 @@
...
+			"CriminalActivities",
+			"count",
+			"measuredValue",
+			map[string]string{
+				"crimeType": "UCR_CombinedCrime",
+			},
+		},
+	} {
+		// TODO: remove this when dataset is updated.
+		continue
```
Spaceenter, line 95: Remove this?

proto/mixer.proto
shifucun: Added them back for now

e2etest/get_related_places_test.go
shifucun: if remove test won't pass. Need to use the new dataset.

PR #234, opened 2020-07-09
Convert more examples to e2e test

shifucun:


shifucun:
> nice. though take a look at the lint errors below..

Yep, fixed test and lint errors

PR #235, opened 2020-07-09
Move schema mapping from config file to be in the code

shifucun:
The schema mapping is never changed and this simplifies the deployment structure/process


shifucun:
> Can you add a note, like our internal CLs, on testing and release-plan?

Ok, I will add the steps in the release process. Most of the current steps will be gone once the CI is setup.

shifucun:
Thanks for review!

PR #242, opened 2020-07-13
Update data version to 2020_07_12_01_01_43

shifucun:
Use approximate match for NYT covid data


shifucun:
> Will we need a similar fix in Prober too?

CEP prober triples check a different triple. We will see if it alerts diff.

PR #243, opened 2020-07-14
Fix GetPropertyValue limit bug

shifucun:
The limit function was missed during the last big refactor. Now we have tests...


shifucun:
thx for review!

PR #245, opened 2020-07-14
Add ranking to covid data; Better branch cache merging for GetStats

shifucun:
Also add mem cache to GetStats e2e test


server/stats.go
```diff
@@ -80,14 +82,22 @@ func (s *Server) GetStats(ctx context.Context, in *pb.GetStatsRequest) (
...
+		extraData, err := readStats(ctx, s.btTable, buildStatsKey(extraDcids, statsVar))
+		if err != nil {
+			return nil, err
+		}
+		for dcid := range extraData {
+			result[dcid] = extraData[dcid]
+		}
 	}
 
 	// Fill missing place data and result result
```
pradh, line 103: replace in.GetPlace() below with "placeDcids" too?

server/stats.go
```diff
@@ -80,14 +82,22 @@ func (s *Server) GetStats(ctx context.Context, in *pb.GetStatsRequest) (
 			result[dcid] = tmp[dcid].(*pb.ObsTimeSeries)
 		}
 	}
-
 	// Read data from base cache if no branch cache data is found.
```
pradh, line 85: Update the comment to reflect current code... i.e., clarify that for DCIDs missing stats in branch cache, we'll get from base cache.

server/stats.go
```diff
@@ -41,6 +41,7 @@ var statsRanking = map[rankKey]int{
 	{"WorldDevelopmentIndicators", ""}:                 3, // Population
 	{"BLS_LAUS", "BLSSeasonallyUnadjusted"}:            0, // Unemployment Rate
 	{"EurostatData", ""}:                               1, // Unemployment Rate
+	{"NYT_COVID19", "NYT_COVID19_GitHub"}:              0, // Covid
```
pradh, line 44: Maybe I'm mistaken, but don't we need to provide the other ones too, with a lower priority?

server/stats.go
shifucun: Done

server/stats.go
shifucun: Done

server/stats.go
shifucun: So by default the score is the lowest score: https://github.com/datacommonsorg/mixer/pull/245/files#diff-090a189afcd6b3645ff26daa525859f6R177-R179.

shifucun:
Thanks for quick review Prashanth!

PR #247, opened 2020-07-17
Add measurement method and unit as parameter to GetStats API

shifucun:
This allows client to get a specific time series based on measurement method and unit. Only time series with exact exact match on given parameters are returned. If there are multiple of them, the regular ranking apply and returns the best ranked one.


proto/mixer.proto
```diff
@@ -214,6 +214,12 @@ message GetStatsRequest {
   // The dcid of the StatisticalVariable.
   string stats_var = 2;
 
+  // The measurement method of the observation.
```
beets, line 217: please specify that these are optional here.

beets:
lgtm, though how will this play out to the new proposal? will the other variants call this and keep these parameters optional?

proto/mixer.proto

shifucun:
> lgtm, though how will this play out to the new proposal? will the other variants call this and keep these parameters optional?

Yes, for "Simple" user case, python api can call this and apply filtering. Will still need the "Advanced" API which return multiple time series.

e2etest/get_stats_test.go
```diff
@@ -48,27 +48,39 @@ func TestGetStats(t *testing.T) {
 	for _, c := range []struct {
 		statsVar     string
 		place        []string
+		mmethod      string
```
pradh, line 51: have a unit test?

shifucun: Done

pradh: Oh nevermind, later realized the unit isn't really implemented yet (blocked on cache).

shifucun:
Thanks for review!

PR #249, opened 2020-07-21
Add a new GetRelatedLocations API which uses the new cache keyed by stat var DCID.

Spaceenter:


server/places.go
```diff
@@ -169,6 +169,79 @@ func (s *Server) GetRelatedPlaces(ctx context.Context,
...
+	}
+	if !util.CheckValidDCIDs([]string{in.GetDcid()}) {
+		return nil, fmt.Errorf("invalid DCID")
+	}
+
+	withinPlace := in.GetWithinPlace()
+	samePlaceType := in.GetSamePlaceType()
+	isPerCapita := in.GetIsPerCapita()
+	var prefix string
+	if withinPlace == "" {
```
shifucun, line 186: wonder if it's better to have a map from <bool, bool, bool> to prefix in here?

Spaceenter: Good point - done.

server/bigtable.go
```diff
@@ -17,19 +17,31 @@ package server
...
+)
+
 // bigTableReadRowsParallel reads BigTable rows in parallel,
 // considering the size limit for RowSet is 500KB.
 func bigTableReadRowsParallel(
 	ctx context.Context,
 	btTable *bigtable.Table,
 	rowSet bigtable.RowSet,
 	action func(string, []byte) (interface{}, error),
+	tokenType TokenType,
```
shifucun, line 44: i think it does the same thing as the opts param below where i call RemoveKeyPrefix(), so can remove opts?

shifucun, line 44: do you think it better to pass the "extract token" function directly in here instead my old way of a bool or a enum here?

Spaceenter: Done.

Spaceenter: Done.

shifucun:
nice to have getToken()!

PR #250, opened 2020-07-21
Add branch cache support for GetTriples on population node

shifucun:


server/misc.go
```diff
@@ -33,7 +33,7 @@ func (s *Server) Translate(ctx context.Context,
 	}
 
 	out := pb.TranslateResponse{}
-	mappings, err := translator.ParseMapping(in.GetSchemaMapping(), s.metadata.Bq)
+	mappings, err := translator.ParseMapping(in.GetSchemaMapping(), "bq")
```
pradh, line 36: For my understanding, where is the DB used in the mapping (I tried following the func calls, it went multiple levels :)

server/triple.go
```diff
@@ -179,6 +173,37 @@ func (s *Server) GetTriples(ctx context.Context, in *pb.GetTriplesRequest) (
...
+	if err != nil {
+		return nil, err
+	}
+	return &triples, nil
+}
+
+func convertPopTriples(dcid string, jsonRaw []byte) (interface{}, error) {
+	jsonVal := string(jsonRaw)
+	parts := strings.Split(jsonVal, "^")
+	if len(parts) == 0 || len(parts)%2 != 0 {
```
pradh, line 188: nit: len(parts) % 2

?

server/triple.go
```diff
@@ -75,6 +75,14 @@ func (s *Server) GetTriples(ctx context.Context, in *pb.GetTriplesRequest) (
 			resultsMap[dcid] = applyLimit(dcid, allTriplesCache[dcid].Triples, limit)
 		}
 	}
+	// Regular triples cache from memcache for population data
+	if len(popDcids) > 0 {
+		dataMap := s.memcache.ReadParallel(
+			buildTriplesKey(popDcids), convertTriplesCache)
```
pradh, line 81: Do we have an inventory of which other cache types don't support branch cache?

main.go
```diff
@@ -29,12 +29,13 @@ import (
...
-	btProject  = flag.String("bt_project", "", "GCP project containing the BigTable instance.")
-	btInstance = flag.String("bt_instance", "", "BigTable instance.")
-	projectID  = flag.String("project_id", "", "The cloud project to run the mixer instance.")
-	port       = flag.String("port", ":12345", "Port on which to run the server.")
+	bqDataset    = flag.String("bq_dataset", "", "DataCommons BigQuery dataset.")
+	btTable      = flag.String("bt_table", "", "DataCommons Bigtable table.")
+	btProject    = flag.String("bt_project", "", "GCP project containing the BigTable instance.")
+	btInstance   = flag.String("bt_instance", "", "BigTable instance.")
+	projectID    = flag.String("project_id", "", "The cloud project to run the mixer instance.")
+	branchFolder = flag.String("branch_folder", "", "The branch cache gcs folder.")
```
pradh, line 37: Is it just for testing?

main.go
```diff
@@ -66,11 +67,15 @@ func main() {
...
 	}
 	// Memcache
-	branchCacheFolder, err := server.ReadBranchCacheFolder(
-		ctx, branchCacheBucket, branchCacheVersionFile)
-	if err != nil {
-		log.Fatalf("failed to read branch cache folder: %v", err)
+	branchCacheFolder := *branchFolder
+	if branchCacheFolder == "" {
+		branchCacheFolder, err = server.ReadBranchCacheFolder(
+			ctx, branchCacheBucket, branchCacheVersionFile)
```
pradh, line 73: Once we launch its going to be very tough to add anything to branch cache without risk of breaking prod. Not this PR, but soon, should we make separate paths for prod vs. staging and separate pubsub topics? Such that, the downstream pipeline writes to staging, and after validation, we'll have a script to: 1. copy over the branch cache folder file, 2. send prod notification.

examples/check_bt.sh
```diff
@@ -0,0 +1,26 @@
...
+export BT=$(head -1 ../deployment/bigtable.txt)
+export PROJECT=google.com:datcom-store-dev
+echo $BT
+echo $PROJECT
+
+# Check get_stats cache
+echo "Get stats for d/f/country/AUT^count^measured^^^^Person^gender^Male"
+cbt -project=$PROJECT -instance prophet-cache read ${BT} prefix="d/f/country/AUT^count^measured^^^^Person^gender^Male"
+echo "Get stats for d/f/country/AUT^count^measured^Count_Person_25To64Years^^^Person^age^Years25To64^educationalAttainment^LessThanPrimaryEducation&PrimaryEducation&LowerSecondaryEducation"
+cbt -project=$PROJECT -instance prophet-cache read ${BT} prefix="d/f/country/AUT^count^measured^Count_Person_25To64Years^^^Person^age^Years25To64^educationalAttainment^LessThanPrimaryEducation&PrimaryEducation&LowerSecondaryEducation"
```
pradh, line 26: This looks useful for debugging, should we make this an argument which defaults to this prefix?

main.go
shifucun: Yes, this is for local testing with custom branch folder

main.go
shifucun: yea, that sounds reasonable.

server/misc.go
shifucun: It's passed into here: https://github.com/datacommonsorg/mixer/blob/master/translator/mcf.go#L129

where it adds the bq to the column name prefix

server/triple.go
shifucun: This is something that need thorough examination. We have branch cache for many cache types but I am not entirely sure how are build, and how many overlap they have with base cache. This has indication of the merging. We should find a time later to go through each branch cache to figure out

1) if we need the branch cache
2) how it should be merged.

shifucun: created https://github.com/datacommonsorg/mixer/issues/251 to track

examples/check_bt.sh
shifucun: I wan to throw a bunch of different bt cache keys in here for all cache types, that might need a lot arguments? what would be a good bash for this purpose?

shifucun: ok, made the prefix as key and move the concrete ones as list comments above

pradh: What you have works. Alternatively, we can accept input on stdin (refer: https://screenshot.googleplex.com/mhkM9hFpCDs), so they can do like below.

# echo "<KEY1>" | ./check_bt.sh

# cat key_file.txt
<KEY1>
<KEY2>
...
# cat key_file.txt | ./check_bt.sh

server/triple.go
pradh: Thanks, we can follow up there.

PR #253, opened 2020-07-23
Update bt bq versions and golden responses to 2020_07_22_01_39_12

tjann:


golden_response/staging/get_stats/census_pep.json
```diff
@@ -123,6 +123,7 @@
       "2019": 328239523
     },
     "place_name": "United States",
+    "is_dc_aggregate": true,
```
tjann, line 126: @shifucun did you add this?

shifucun: hmm, this is odd, you can submit this. I will take a look

tjann: Ok

PR #254, opened 2020-07-23
GetStats: return source url

shifucun:
This will be used by client side to show data source url.

This PR also use getToken() function in memcache parallel read to align with BT read.


proto/mixer.proto
```diff
@@ -190,13 +190,15 @@ message ObsTimeSeries {
     string measurement_method = 3;
     string observation_period = 4;
     string import_name = 5;
+    string provenanceDomain = 6;
```
pradh, line 193: This deviates from the use of snake case above.

proto/mixer.proto
```diff
@@ -190,13 +190,15 @@ message ObsTimeSeries {
...
   }
   map<string, double> data = 1;  // Date to value.
   string unit = 2;
   string place_name = 3;
   bool is_dc_aggregate = 4;
   string place_dcid = 5;  // Only set if the key uses mid.
   repeated SourceSeries source_series = 6;
+  string provenanceDomain = 7;
```
pradh, line 201: Puzzled why this is needed since you have it in SourceSeries?

shifucun: Now i only return one source series, and this is popped to upper level.

proto/mixer.proto
shifucun: changed.

proto/mixer.proto
pradh: Is it temporary then? Add this comment to proto and have a TODO to remove if temp (later PR is fine).

PR #255, opened 2020-07-27
Add a new mixer method GetLandingPage for the landing page new cache.

Spaceenter:


proto/mixer.proto
```diff
@@ -817,6 +829,14 @@ service Mixer {
     };
   }
 
+  // Get landing page info for a place.
+  rpc GetLandingPage(GetLandingPageRequest) returns (GetLandingPageResponse) {
+    option (google.api.http) = {
+      post: "/node/landing-page"
```
shifucun, line 835: can you also add statsvar as paramters? So client can ask only data for certain stats vars

Spaceenter: Done.

PR #258, opened 2020-07-29
Fix OOM failures in csv to bt dataflow job

pradh:
My theory for the OOM is that our total file size went past a threshold (9TB to 13TB), and the TextIO default "desired split size" [0] apparently was small enough (64MB) to cause many splits (~200K), and blow up memory usage. This is based on the OOM stack below, where performSplitTyped() [1] splits the input file pattern into splits and serializeSplitToCloudSource serializes it (potentially doubling mem usage).

I was trying to find a way to set the "desiredBundleSizeBytes", but that wasn't exposed [2]. Instead, I found the "withHintMatchesManyFiles" hint which apparently causes the inputs to be processed differently (by first constructing PCollection of files, and then processing them) [3], and thus avoiding this issue.

Additionally, made the damn code compile and ran the instructions in the README to update the dataflow template.

Running with the fix, the Dataflow job that used to OOM in the first couple of mins before even scaling up, has been running OK for close to an hour now.

[0] - https://github.com/apache/beam/blob/6fdde4f4eab72b49b10a8bb1cb3be263c5c416b5/runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/WorkerCustomSources.java#L83
[1] - https://github.com/apache/beam/blob/6fdde4f4eab72b49b10a8bb1cb3be263c5c416b5/runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/WorkerCustomSources.java#L212
[2] -https://github.com/apache/beam/blob/6fdde4f4eab72b49b10a8bb1cb3be263c5c416b5/sdks/java/core/src/main/java/org/apache/beam/sdk/io/TextIO.java#L554-L556
[3] - https://github.com/apache/beam/blob/6fdde4f4eab72b49b10a8bb1cb3be263c5c416b5/sdks/java/core/src/main/java/org/apache/beam/sdk/io/TextIO.java#L361-L372

== OOM Stack ==
org.apache.beam.sdk.util.SerializableUtils.serializeToByteArray(SerializableUtils.java:52) com.google.cloud.dataflow.worker.WorkerCustomSources.serializeSplitToCloudSource(WorkerCustomSources.java:145) com.google.cloud.dataflow.worker.WorkerCustomSources.wrapIntoSourceSplitResponse(WorkerCustomSources.java:300) com.google.cloud.dataflow.worker.WorkerCustomSources.performSplitTyped(WorkerCustomSources.java:202) com.google.cloud.dataflow.worker.WorkerCustomSources.performSplitWithApiLimit(WorkerCustomSources.java:181) com.google.cloud.dataflow.worker.WorkerCustomSources.performSplit(WorkerCustomSources.java:160)


pradh:
Thanks.

PR #266, opened 2020-08-06
Update mixer to the newest data release: 2020_08_06_01_02_58

beets:
* Update Goldens
* Add ability to update goldens for query_test.go
* Updated README


beets:
Please check if the diffs are expected, thanks!

golden_response/staging/get_landing_page/county.json
```diff
@@ -267,6 +267,23 @@
...
+              "2011": 1762754,
+              "2012": 1788393,
+              "2013": 1812208,
+              "2014": 1841569,
+              "2015": 1868149,
+              "2016": 1885056,
+              "2017": 1911226,
+              "2018": 1922200
+            }
+          },
```
tjann, line 286: @shifucun I thought we had done some ordering and serialization to make sure these ordering diffs don't show?

beets:
Landing page API is in flux right now, so I think having a noisy test there is ok for now.

fwiw, I'm not a fan of having the style updates causing diff's. Not sure why the formatting changed.

PR #267, opened 2020-08-07
Remove GetChart api which is used by old GNI tool

shifucun:
The client code is removed in https://github.com/datacommonsorg/website/pull/258


shifucun:
thx for review!

PR #268, opened 2020-08-07
Update GetStats to use new cache with scaling factor and unit in Sour…

shifucun:
…ceSeries


golden_response/staging/get_chart_data.json
```diff
@@ -223,10 +223,42 @@
...
+      "2020-07-31": 9913,
+      "2020-08-01": 10321,
+      "2020-08-02": 10626,
+      "2020-08-03": 10794,
+      "2020-08-04": 11030
     },
     "place_name": "Santa Clara County",
-    "place_dcid": "geoId/06085"
+    "place_dcid": "geoId/06085",
+    "provenance_domain": "nytimes.com"
```
shifucun, line 261: @pradh this might be one example that we want to update golden data when there is new data (field). So for each cache, need to determine on what field should be considered as "redundant" or not for truth assertion

PR #273, opened 2020-08-12
Filter Landing Page cache's ObsTimeSeries

beets:
* Use the same logic as get_stats
* Update get_stats to use go interface instead of proto
* Update goldens


beets:
Fixed tests, ptal! Hope the fix is the right one..

golden_response/staging/get_stats/census_pep.json
```diff
@@ -332,8 +332,8 @@
       "2017": 81197,
       "2018": 83377
     },
-    "place_name": "City of Mountain View",
-    "place_dcid": "geoId/0649670",
-    "provenance_domain": "census.gov"
+    "placeName": "City of Mountain View",
```
shifucun, line 335: Note website also needs to update the field name and release with the mixer.

beets:
@tjann will the change from snake case to camel case cause issues in our new notebooks? The python API simply passes through these properties, but it might be used by callers of the API.

tjann:
No issues for notebooks

beets:
thanks! merging..

PR #276, opened 2020-08-13
Bump data version to 2020_08_13_01_04_17

shifucun:


golden_response/staging/get_landing_page/county.json
```diff
@@ -470,36 +470,372 @@
...
-        "2011": 939851,
-        "2012": 958281,
-        "2013": 971863,
-        "2014": 989909,
-        "2015": 1012212,
-        "2016": 1027805,
-        "2017": 1038698,
-        "2018": 1043641,
-        "2019": 1053723
+        "1990-01": 847597,
```
tjann, line 473: Ah... we should probably fix the obs period to avoid these diffs?

golden_response/staging/get_stats/nyt_covid_cases.json
```diff
@@ -155,52 +155,56 @@
...
+      "2020-07-18": 3722851,
+      "2020-07-19": 3785126,
+      "2020-07-20": 3845014,
+      "2020-07-21": 3910291,
+      "2020-07-22": 3980030,
+      "2020-07-23": 4050036,
+      "2020-07-24": 4123562,
+      "2020-07-25": 4190338,
+      "2020-07-26": 4244555,
+      "2020-07-27": 4303736,
```
tjann, line 192: Found this here: https://github.com/nytimes/covid-19-data/blob/e2c51040dd8d008f7fa1c6ac3ae3471dc2632c61/us.csv

The data actually changed already 😬

tjann:
LG, not sure if we should fix obs period now or later?

golden_response/staging/get_landing_page/county.json
pradh: Hmm.. sees like the returned data switched from yearly to monthly. Did any ranking change? Or is it the new cache?

shifucun: I think this is from the filter method which prefers BLS-LAUS mmethod

pradh: Never mind, I didn't read Bo's comment about it.

I think the obs period LG then.

shifucun:
Thank you all for validation!

PR #277, opened 2020-08-19
GetStatValue API that returns a single value for one place and one statVar

shifucun:
This is the simplest version of GetStat() API, that takes one place and one stat var and return a single value.
Client can provide an optional date argument, otherwise the latest stat value is returned.
The value is from preferred source and measurement method when desired date is available.


e2etest/get_stat_value_test.go
```diff
@@ -0,0 +1,116 @@
...
+		var expected *pb.GetStatValueResponse
+		file, _ := ioutil.ReadFile(goldenFile)
+		err = json.Unmarshal(file, &expected)
+		if err != nil {
+			t.Errorf("Can not Unmarshal golden file")
+			continue
+		}
+
+		if diff := cmp.Diff(resp, expected, protocmp.Transform()); diff != "" {
+			t.Errorf("payload got diff: %v", diff)
```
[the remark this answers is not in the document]
tjann: FWIW, I'm thinking it might be good to have good errors from REST that we can reuse in Python. WDYT?

server/stats.go
```diff
@@ -49,6 +50,86 @@ const lowestRank = 100
 // Limit the concurrent channels when processing in-memory cache data.
 const maxChannelSize = 50
 
+// byRank implements sort.Interface for []*SourceSeries based on
+// the rank score.
+type byRank []*SourceSeries
```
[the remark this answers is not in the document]
tjann: Nvm, this seems to be a Golang convention..

server/stats.go
```diff
@@ -176,6 +257,38 @@ func triplesToStatsVar(
 	return &statsVar, nil
 }
 
+// getValue get the stat value from ObsTimeSeries.
+// When date is given, it get the value from the highest ranked source series
+// that has the date.
+// When date is not given, it get the latest value from the highest ranked
+// source series.
+func getValue(in *ObsTimeSeries, date string) (float64, error) {
```
pradh, line 265: I'm trying to understand why this logic is separate and different from filterAndRank()?

Would it be better to have a common function for ranking series that takes optional mmethod, unit, period, date?

server/stats.go
```diff
@@ -176,6 +257,38 @@ func triplesToStatsVar(
...
+	sourceSeries := in.SourceSeries
+	sort.Sort(byRank(sourceSeries))
+	if date != "" {
+		for _, series := range sourceSeries {
+			if value, ok := series.Val[date]; ok {
+				return value, nil
+			}
+		}
+		return 0, fmt.Errorf("No data found for date %s", date)
+	}
```
pradh, line 278: By sorting first and then checking date, we miss out if the 2nd best series has the date we want right?

Goes back to the above comment on whether the ranking logic should take date into account.

shifucun: After sorting, this checks the highest ranking series first, if found the date then return it, otherwise will check the second best series and go on this. Basically if multiple series have the same date then this picks from the highest ranked series.

server/stats.go
shifucun: Right now getValue does not care the mmethod etc since it is tailored for the simple api where user does not require knowledge on these. After we have the new apis ready, no api will take mmethod as argument, instead the /stat/all will return all series then client can pick from those.

After I can combine the logic together so the ranking score only appears in one place.

e2etest/get_stat_value_test.go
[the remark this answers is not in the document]
shifucun: Oh, this is the error message in e2e test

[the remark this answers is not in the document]
shifucun: yea, resue error msg sounds good. I fixed an error issue in getValue() function

server/stats.go
```diff
@@ -49,6 +50,86 @@ const lowestRank = 100
...
+		if data == nil {
+			obsTimeSeries = nil
+		} else {
+			obsTimeSeries = data.(*ObsTimeSeries)
+		}
+	} else {
+		// If the data is missing in branch cache, fetch it from the base cache in
+		// Bigtable.
+		btData, err := readStats(
+			ctx, s.btTable, buildStatsKey([]string{place}, statVarObject))
```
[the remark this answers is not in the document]
shifucun: Done

server/stats.go
```diff
@@ -176,6 +257,38 @@ func triplesToStatsVar(
...
 }
 
+// getValue get the stat value from ObsTimeSeries.
+// When date is given, it get the value from the highest ranked source series
+// that has the date.
+// When date is not given, it get the latest value from the highest ranked
+// source series.
+func getValue(in *ObsTimeSeries, date string) (float64, error) {
+	if in == nil {
+		return 0, fmt.Errorf("Nil obs time series for getLatest()")
```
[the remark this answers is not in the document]
shifucun: Fixed

server/stats.go
pradh: SG

server/stats.go
pradh: Ah, I misread. This LG

server/stats.go
```diff
@@ -49,6 +50,85 @@ const lowestRank = 100
...
+		scorei = lowestRank
+	}
+	oj := a[j]
+	keyj := rankKey{oj.ImportName, oj.MeasurementMethod}
+	scorej, ok := statsRanking[keyj]
+	if !ok {
+		scorej = lowestRank
+	}
+	// Higher score value means lower rank.
+	return scorei < scorej
```
pradh, line 77: For determinism, should we also include the other fields in sort decision: observationPeriod, unit, scalingFactor.

server/stats.go
```diff
@@ -176,6 +256,41 @@ func triplesToStatsVar(
...
+	sort.Sort(byRank(sourceSeries))
+	if date != "" {
+		for _, series := range sourceSeries {
+			if value, ok := series.Val[date]; ok {
+				return value, nil
+			}
+		}
+		return 0, fmt.Errorf("No data found for date %s", date)
+	}
+	currDate := ""
```
pradh, line 278: nit: latestDate ?

proto/mixer.proto
```diff
@@ -589,6 +589,20 @@ message InterestingPlaceAspect {
   repeated Aspect aspects = 1;
 }
 
+// Request for GetStat service.
+message GetStatValueRequest {
+  // dcid of the place.
+  string place = 1;
+  // dcid of the stats var.
```
pradh, line 596: nit: stat

shifucun: done

server/stats.go
shifucun: done

server/stats.go
shifucun: SG. Added a TODO for this. Will fix when this set of api work is done.

shifucun:
Thanks for review! PTAL

server/stats.go
```diff
@@ -176,6 +258,41 @@ func triplesToStatsVar(
...
+	sort.Sort(byRank(sourceSeries))
+	if date != "" {
+		for _, series := range sourceSeries {
+			if value, ok := series.Val[date]; ok {
+				return value, nil
+			}
+		}
+		return 0, fmt.Errorf("No data found for date %s", date)
+	}
+	latestData := ""
```
pradh, line 280: nit: latestDate?

pradh (dismissed):

server/stats.go
pradh: I mean, s/Data/Date/

shifucun: oops. fixed.

proto/mixer.proto
```diff
@@ -589,6 +589,20 @@ message InterestingPlaceAspect {
   repeated Aspect aspects = 1;
 }
 
+// Request for GetStat service.
+message GetStatValueRequest {
+  // dcid of the place.
+  string place = 1;
```
beets, line 595: do we prefer place over dcid? just to note that every api except for GetStatRequest uses dcid or dcids, rather than place. thoughts?

proto/mixer.proto
```diff
@@ -589,6 +589,20 @@ message InterestingPlaceAspect {
...
+  // dcid of the stat var.
+  string stat_var = 2;
+  // (optional) date of the stat.
```
beets, line 598: add a note here that latest is use if unspecified.

proto/mixer.proto
```diff
@@ -589,6 +589,20 @@ message InterestingPlaceAspect {
...
+  string date = 3;
+}
+
+message GetStatValueResponse {
+  double value = 1;
```
beets, line 603: i know we only want the value back in the py api, but should we add other parameters for the value in the return object that would be useful? e.g. date the value was derived from?

PR #281, opened 2020-08-20
Add optional api parameters for GetStatValue api

shifucun:


e2etest/get_stat_value_test.go
```diff
@@ -48,42 +48,56 @@ func TestGetStatValue(t *testing.T) {
...
 		},
 		{
 			"Median_Age_Person",
 			"geoId/0649670",
 			"median_age.json",
+			"",
 			false,
 		},
 		{
 			"Amount_EconomicActivity_GrossNationalIncome_PurchasingPowerParity_PerCapita",
```
tjann, line 76: Should we eventually add other obs properties to test?

proto/mixer.proto
```diff
@@ -610,13 +618,13 @@ message GetStatSeriesRequest {
...
+  string measurement_method = 3;
   // (optional) observation period of the observation, ex: "P1Y".
-  string observation_period = 5;
+  string observation_period = 4;
   // (optional) unit of the observation.
-  string unit = 6;
+  string unit = 5;
   // (optional) scaling factor of the observation.
-  string scaling_factor = 7;
+  string scaling_factor = 6;
```
tjann, line 627: Is it better practice to have the GetStatValue and GetStatSeries Request be duped rather than combined?

e2etest/get_stat_value_test.go
shifucun: yes. will can enrich the tests later.

proto/mixer.proto
shifucun: yeah, usually each A service will have corresponding ARequest and AResponse as req and resp message names.

PR #282, opened 2020-08-20
GetStatAll which takes multiple places and statistical variables

shifucun:
Remove placeDcid from all get stat api response since it is redundant.


proto/mixer.proto
```diff
@@ -625,6 +623,20 @@ message GetStatSeriesResponse {
...
 }
 
+// Request for GetStatAll service.
+message GetStatAllRequest {
+  // dcids of the place.
+  repeated string places = 1;
+  // dcids of the stat var.
+  repeated string stat_vars = 2;
+}
+
```
beets, line 633: comments for the response proto? especially to note what the string keys are.

server/stats.go
```diff
@@ -236,6 +251,82 @@ func (s *Server) GetStatSeries(ctx context.Context, in *pb.GetStatSeriesRequest)
 	return &resp, nil
 }
 
+// GetStatAll implements API for Mixer.GetStatAll.
+func (s *Server) GetStatAll(ctx context.Context, in *pb.GetStatAllRequest) (
+	*pb.GetStatAllResponse, error) {
+	places := in.GetPlaces()
+	statVars := in.GetStatVars()
+	if len(places) == 0 || len(statVars) == 0 {
+		return nil, fmt.Errorf("Missing required arguments")
```
beets, line 260: Could we note which missing argument in the error message?

proto/mixer.proto
shifucun: done

server/stats.go
shifucun: done

server/stats.go
```diff
@@ -236,6 +251,84 @@ func (s *Server) GetStatSeries(ctx context.Context, in *pb.GetStatSeriesRequest)
...
+	for key, token := range keyTokens {
+		if result.PlaceData[token.place].StatVarData[token.statVar] == nil {
+			extraRowList = append(extraRowList, key)
+		}
+	}
+	extraData, err := readStatsPb(ctx, s.btTable, extraRowList, keyTokens)
+	for place, placeData := range extraData {
+		for statVar, data := range placeData {
+			result.PlaceData[place].StatVarData[statVar] = data
+		}
```
pradh, line 327: should we confirm if there is anything at all to read from extraRowList before calling readStatsPb?

server/stats.go
```diff
@@ -447,20 +568,69 @@ func convertToObsSeries(dcid string, jsonRaw []byte) (interface{}, error) {
...
+		} else {
+			result[place][statVar] = data.(*ObsTimeSeries)
+		}
+	}
+	return result, nil
+}
+
+// readStats reads and process BigTable rows in parallel.
+// Consider consolidate this function and bigTableReadRowsParallel.
+func readStatsPb(
```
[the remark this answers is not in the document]
shifucun: current GetStats() return a response like {"payload": ObsTimeSeriesJsonString}, it is not possible to json.unmarshall this with protobuf, so we have a golang struct to mimic the protobuf, that way this response can be unmarshalled with the golang struct.

In this case, the response itself is a proto so we don't need to use the golang struct. Hopefully when we deprecate the old api, we can remove the need for the golang struct.

server/stats.go
shifucun: done

server/stats.go
```diff
@@ -236,6 +251,84 @@ func (s *Server) GetStatSeries(ctx context.Context, in *pb.GetStatSeriesRequest)
...
+		}
+	}
+
+	// Read data from branch in-memory cache first.
+	cacheData := s.memcache.ReadParallel(
+		rowList,
+		convertToObsSeries,
+		func(rowKey string) (string, error) {
+			return keyTokens[rowKey].place + "^" + keyTokens[rowKey].statVar, nil
+		},
```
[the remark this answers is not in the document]
shifucun: Use a function to generate this function now

server/stats.go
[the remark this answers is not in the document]
beets: sorry, i didn't catch that we're using the proto now (we just went through and updated to the struct, which i thought was preferred? either way, we should stick with one (camel or snake case for keys) and not have both in use.

my vote is for camel case, which is what currently other calls return.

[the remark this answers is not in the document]
shifucun: I propose to use protobuf for these new apis. The golang struct was used to accommodate for the old {"payload": "jsonStr"} pattern is not ideal.

pradh (dismissed):

proto/mixer.proto
```diff
@@ -633,6 +631,62 @@ message GetStatSeriesResponse {
...
+// Response for GetStatAll service.
+// The response is a two level map, with the first level keyed by place dcid,
+// and the second level keyed by the stat var dcid.
+// Each leaf object contains multiple source series with <date, value> object
+// and observation metadata.
+//
```
beets, line 647: maybe it's worth converting these to camelcase as well, and noting here that the conversion happens somewhere..

beets (dismissed):
thanks for explaining the changes, esp. in context with #284

proto/mixer.proto
shifucun: done

PR #295, opened 2020-08-24
Upgrade BT import job's dataflow version from (deprecated) 2.4 to the latest

pradh:
I have run the command in the README to update the dataflow template on GCS.


pradh:
Thanks.

PR #296, opened 2020-08-25
Release 2020_08_25_01_02_13

pradh:


golden_response/staging/get_triples/place_type.json
```diff
@@ -1,5 +1,1951 @@
...
+      "objectId": "Country",
+      "objectName": "Country",
+      "objectTypes": [
+        "Class"
+      ],
+      "provenanceId": "dc/5j06ly1"
+    },
+    {
+      "subjectId": "wikidataId/Q8733",
+      "subjectName": "Qing dynasty",
```
shifucun, line 61: Is ancient country expected?

golden_response/staging/get_triples/place_type.json
```diff
@@ -495,7 +2444,7 @@
       "subjectId": "country/TCA",
       "subjectName": "Turks and Caicos Islands",
       "subjectTypes": [
-        "AdministrativeArea",
+        "AdministrativeArea1",
```
shifucun, line 2447: is this expected?

golden_response/staging/get_stat_all/result.json
```diff
@@ -1,698 +1,698 @@
 {
- "placeData": {
-  "country/USA": {
-   "statVarData": {
-    "Amount_EconomicActivity_GrossNationalIncome_PurchasingPowerParity_PerCapita": {
-     "sourceSeries": [
+ "placeData":  {
```
shifucun, line 2: @tjann I also got this formatting result from the command, can you try this locally? I don't think this should be related to vscode

golden_response/staging/get_triples/place_type.json
pradh: Yes because it is a british overseas territory as well. So AA1 is probably more accurate than AA.

golden_response/staging/get_triples/place_type.json
pradh: Nope, it is not good. All these entities are indirect subclasses of country.

I'll file a bug to drop historical stuff, that will need more pruning.

pradh: But I think we should go ahead with this anyway so we can test it out and find more bugs like this.

golden_response/staging/get_pop_obs/moscow.json
```diff
@@ -6202,7 +6202,7 @@
           "observation_period": "P1Y",
           "measurement_denominator": "Count_Person_Upto14Years",
           "stat_var_ids": [
-            "Count_Death_Upto14Years_AsAFractionOf_Count_Person_Upto14Years"
+            "dc/fef0cwm863qz5"
```
pradh, line 6205: Still debugging why we lost curated stat-vars.

golden_response/staging/get_pop_obs/moscow.json
```diff
@@ -8942,6 +8942,18 @@
...
+          "type": "Observation",
+          "measured_prop": "count",
+          "measured_value": 11979529,
+          "measurement_method": "GoogleKGHumanCurated",
+          "provenance_id": "dc/97m6fj1",
+          "observation_date": "2013",
+          "stat_var_ids": [
+            "Count_Person"
+          ]
+        },
```
pradh, line 8956: I spent half an hour trying to debug this diff. Seemingly, the content of the Obs stayed the same, and yet the DCID changed.

My best guess is that @shifucun manually edited the resolved MCF to add "measurementMethod=GoogleKGHumanCurated". So really, the obs DCID on the LHS is bogus and the RHS is correct (unit test illustrating this: cl/328421816).

For future reference, we should always edit unresolved MCFs and re-resolve.

golden_response/staging/get_triples/place_type.json
shifucun: can you file an issue with the the golden files to revert back later.

pradh: https://github.com/datacommonsorg/mixer/issues/297

pradh:
Thanks Bo.

pradh:
@tjann  I want to push this now to test out in staging. This is not going to prod. Please review the diffs in case you think I missed something.

Issue #297, opened 2020-08-25
[BUG] Revert historical geos in place_type.json in next staging refresh


pradh:
Context: https://github.com/datacommonsorg/mixer/pull/296/files#r476778097

PR #300, opened 2020-08-27
Release 2020_08_27_01_02_21 data version

pradh:
Overall these seem like expected/+ve changes.

1. Historical places are dropped now (thus, fixing https://github.com/datacommonsorg/mixer/issues/297).

2. Stale AdministrativeArea types are dropped.

3. Ranking changes are from re-running related-places flume.

4. Stat-var changes are also due to re-running related-places flume. (It should technically re-run whenever a stat-var is added, but practically maybe ok since we only use human-defined stat-vars rather than these generated ones.)


pradh:
Thanks for the review!

PR #302, opened 2020-09-01
Add new protobuf based endpoint for stat var rankings

beets:
Added a new endpoint to simplify the arguments and make it less error prone

### DIFF
--- a/golden_response/staging/get_stat_ranking/california.json
+++ b/golden_response/staging/get_stat_ranking/california.json
@@ -1,0 +1,594 @@
+{
+  "payload":  {
+    "Count_Person":  {
+      "rankAll":  {
+        "info":  [
+          {
+            "value":  10039107,
+            "placeDcid":  "geoId/06037"
+          },
+          {
+            "rank":  1,
+            "value":  3338330,
+            "placeDcid":  "geoId/06073"
+          },
+          {
+            "rank":  2,
+            "value":  3175692,
+            "placeDcid":  "geoId/06059"
+          },
+          {
+            "rank":  3,
+            "value":  2470546,
+            "placeDcid":  "geoId/06065"
+          },
+          {
+            "rank":  4,
+            "value":  2180085,
+            "placeDcid":  "geoId/06071"
+          },
+          {
+            "rank":  5,
+            "value":  1927852,
+            "placeDcid":  "geoId/06085"
+          },
+          {
+            "rank":  6,
+            "value":  1671329,
+            "placeDcid":  "geoId/06001"
+          },
+          {
+            "rank":  7,
+            "value":  1552058,
+            "placeDcid":  "geoId/06067"
+          },
+          {
+            "rank":  8,
+            "value":  1153526,
+            "placeDcid":  "geoId/06013"
+          },
+          {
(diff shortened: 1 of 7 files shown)


golden_response/staging/get_stat_ranking/california.json
```diff
@@ -0,0 +1,594 @@
+{
+  "payload":  {
+    "Count_Person":  {
+      "rankAll":  {
+        "info":  [
+          {
+            "value":  10039107,
```
shifucun, line 7: I assume the "rank": 0 is not shown because it's default proto value. Might be better to show it (with some proto field option tuning)

proto/mixer.proto
```diff
@@ -308,16 +308,31 @@ message GetRelatedLocationsRequest {
 
   repeated string stat_var_dcids = 2;
 
-  // (Optional) Wheter to require same place type for related places.
+  // (Optional) Whether to require same place type for related places.
```
shifucun, line 311: good eye!

proto/mixer.proto
```diff
@@ -308,16 +308,31 @@ message GetRelatedLocationsRequest {
...
 
+// Request to get rankings of locations for given stat var DCIDs.
+message GetLocationsRankingsRequest {
+  repeated string stat_var_dcids = 1;
+
+  // Place type to get rankings for.
+  string place_type = 2;
+
+  // (Optional) Whether to require all related places under the same ancestor.
+  string within_place = 4;
```
shifucun, line 329: nit: = 3

server/model.go
```diff
@@ -62,14 +62,26 @@ type chanData struct {
...
+	PlaceDcid string  `json:"placeDcid,omitempty"`
+}
+
+// RelatedPlaceRanking lists ranking information for the RelatedPlaces cache.
+type RelatedPlaceRanking struct {
+	Info []RelatedPlaceRankInfo `json:"info,omitempty"`
+}
+
 // RelatedPlacesInfo represents the json structure returned by the RelatedPlaces cache.
 type RelatedPlacesInfo struct {
```
shifucun, line 78: can remove these go structs?

golden_response/staging/get_stat_ranking/california.json
beets: I took a look into it... We can add EmitUnpopulated: true to the marshaller option, but that would only apply to unit tests. The rank: 0 is stored in the cache, but there's no way to differentiate it in go, or to set an option at the field level. I'm not sure if that's what we want (other unset fields will then show up, in this case relatedPlaces: [], etc).

I could go back and fix this in prophet, starting the rank values at 1? That might make more sense.

proto/mixer.proto
beets: done

server/model.go
beets: i left them for completeness, but you're right, they're not used. removed.

golden_response/staging/get_stat_ranking/california.json
shifucun: I see, right, starting with 1 in prophet sounds good.

PR #308, opened 2020-09-16
Update mixer to use data from latest prophet run: 2020_09_15_12_54_03

beets:
Updated related locations test to be similar to what we'd use on the landing pages - the test was returning non-US places of different types, which won't make sense for our charts.

Updated goldens:
* Provenance Id's are updated for wikidata places
* Rankings positions are now 1-indexed
* Updated covid data

It's really hard to make out the differences in moscow.json - are we expecting any diffs for get_pop_obs there?


golden_response/staging/get_pop_obs/moscow.json
```diff
@@ -774,6 +774,268 @@
         "age": "YearsUpto14"
       }
     },
+    "dc/p/119gy3zecnm61": {
```
tjann, line 777: @beets regarding your question -- yes we were expecting diffs here. I agree that this is very difficult to look at. I will take a closer look on staging if that's ok.

golden_response/staging/get_stat_ranking/country.json
```diff
@@ -1,1256 +1,1267 @@
...
-      "rankAll": {
-        "info": [
+  "payload":  {
+    "Count_Person":  {
+      "rankAll":  {
+        "info":  [
           {
-            "value": 1396530000,
-            "placeDcid": "country/CHN"
+            "rank":  1,
```
tjann, line 7: Nice improvement (rank 1 on first entry)

tjann (approved):

pradh (approved):

shifucun (approved):

beets:
Thanks all!

PR #314, opened 2020-10-01
Fix a few edge case bug in landing page data api

shifucun:
- use lower case for json
- allow fixed seed for testing


tjann (approved):

shifucun:
thx for review!

PR #315, opened 2020-10-01
Data release to version: 2020_09_24_01_02_12

shifucun:
Note there are substantial change in landing page cache data since all the stat var are included there, even there is no data. This is to ensure the mixer can fetch newly added stat var that is not included in Flume


golden_response/staging/get_stat_all/result.json
```diff
@@ -66,6 +66,7 @@
      ]
     },
     "Count_Person": {
+     "isDcAggregate": true,
```
pradh, line 69: @tjann should confirm. I'm surprised why this is an aggregate.

pradh (approved):
Skipped landing-page diffs. Otherwise, LGTM.

golden_response/staging/get_stat_all/result.json
tjann: yeah this is surprising

tjann: Maybe bc this is for all sources, and CensusACS5yrSurvey we aggregate for country IIRC

PR #328, opened 2020-10-08
Related place API always require same place type

shifucun:


pradh (approved):
Thanks for the cleanup Bo!

PR #329, opened 2020-10-08
Update similar places to only use cohort places

shifucun:


beets (approved):

shifucun (commented):
Updated tests, looking good

PR #330, opened 2020-10-08
Fix random golden json issue

shifucun:
protojson don't and won't make stable output: https://github.com/golang/protobuf/issues/1082. Use encoding/json to get stable output.


tjann (approved):
THANK YOU Bo!

shifucun:
> THANK YOU Bo!

Now we should be happier :)

PR #333, opened 2020-10-10
Test for rankings 

beets:
Checks if all the stat vars we use in charts have rankings at a few levels - still very coarse (doesn't test scoped rankings) but at least shows us some coverage.

Ideally should use the chart config directly instead of this munged list of stat vars.


shifucun (commented):
random thought If the code can request to https://raw.githubusercontent.com/datacommonsorg/website/master/server/chart_config.json and take out the stat vars

beets:
that's a good idea!

beets:
Reads the config from website repo now. Lots still missing, but I think the data on staging doesn't have the fix from the weekend.

tjann:
Yeah this push does not have the similar places fix. It should have ranking fix, though.

tjann (approved):

beets (commented):
Updated this to write out missing rankings to a modified chart config golden file. Ptal, hope this helps us to debug ranking issues.

tjann (approved):

PR #334, opened 2020-10-10
Remove interesting place API

shifucun:
This API is never really used by any client, removing to save space and time for cache generation. For further usage, there is a ranking api that gets all the places ranking for a particular stat var


pradh (approved):
For posterity, it might help to add some context for what the motivation for this was and why it is no longer relevant.

shifucun:
> For posterity, it might help to add some context for what the motivation for this was and why it is no longer relevant.

Added some context

PR #340, opened 2020-10-14
Prepare golden prod response files for 2020_10_13_01_01_46 prod data push.

tjann:


pradh (approved):

beets (approved):
we need to set up more goldens for prod...

PR #346, opened 2020-10-16
Push 2020_10_15_19_22_50 (with ranking fix)

pradh:
The ranking diffs generally look reasonable to me. The ranking values change a little if we pick a different source from before (but the code does so deterministically, so hopefully no future diffs, unless a source gets added).

Defer to tjann for sanity-checking landing-page changes.

No rush on looking at this.  Just wanted to make sure I figured out the ranking brokenness (which turns out was due to incompatibility with prophet after I made ranking MCFs refer to curated stat-var-id).


golden_response/staging/get_stat_ranking/california.json
```diff
@@ -442,12 +442,12 @@
           {
             "rank": 29,
             "value": 37.9,
-            "placeDcid": "geoId/06095"
+            "placeDcid": "geoId/06111"
```
pradh, line 445: Hmm... when values match, should sort by place to ensure deterministic diffs. I'll add a note to fix this.

golden_response/staging/get_related_locations/county.json
```diff
@@ -1,18 +1,18 @@
...
   },
   "Median_Income_Person": {
     "rankFromTop": 28,
-    "rankFromBottom": 3158
+    "rankFromBottom": 3157
   },
   "UnemploymentRate_Person": {
-    "rankFromTop": 675,
-    "rankFromBottom": 2512
+    "rankFromTop": 1016,
```
pradh, line 15: This seems suspicious... I'll check on this.

pradh: This is also explained by the choice of time-series, since unemployment-rate fluctuated a lot recently.  Now we prefer the series which has the latest date while computing ranking, so I think this should be more accurate.

beets (approved):

shifucun (approved):

golden_response/staging/get_stat_ranking/crime_percapita.json
```diff
@@ -10,12 +10,12 @@
           },
           {
             "rank": 2,
-            "value": 0.185145406477198,
+            "value": 0.19114636642784,
```
tjann, line 13: Crime data didn't change, perhaps the denominator we use changed sources/mmethods?

tjann (approved):
Only crime caught me by surprise.

golden_response/staging/get_stat_ranking/crime_percapita.json
pradh: Yes, as noted in the other comments, the choice of population value could have changed.

Issue #347, opened 2020-10-19
Investigate OOM issue


shifucun:
GKE pod has a very frequent restart due to OOM. Need to add logging on memory usage and investigate memory usage hotspot.

Debug pointers:

https://blog.detectify.com/2019/09/05/how-we-tracked-down-a-memory-leak-in-one-of-our-go-microservices/
https://golang.org/doc/diagnostics.html

shifucun:
Use cloud profiling: https://cloud.google.com/profiler/docs/profiling-go

shifucun:
![Uploading image.png…]()

PR #350, opened 2020-10-19
Log memory usage and use fmt.Printf to do info level log

shifucun:
Golang GCP logging level: https://cloud.google.com/functions/docs/monitoring/logging.

In longer term, we could consider using logging client. For now, separate info and error level would be good.


main.go
```diff
@@ -99,7 +105,7 @@ func main() {
 	if err != nil {
 		log.Fatalf("failed to listen on network: %v", err)
 	}
-	log.Println("mixer ready to server")
+	fmt.Println("mixer ready to server")
```
beets, line 108: nit: "mixer ready to serve" (extra r at the end)

beets (approved):

beets:
it's very subtle that log / fmt ends up logged differently. should we standardize on fmt, and specify the severity?

main.go
shifucun: Done

util/util.go
```diff
@@ -268,3 +269,19 @@ func RemoveKeyPrefix(key string) (string, error) {
 	}
 	return match[3], nil
 }
+
+// PrintMemUsage outputs the current, total and OS memory being used. As well as the number
+// of garage collection cycles completed.
```
pradh, line 274: We could also consider using https://cloud.google.com/profiler/docs/profiling-go

pradh (approved):

shifucun:
> it's very subtle that log / fmt ends up logged differently. should we standardize on fmt, and specify the severity?

Looked a bit more on https://golang.org/pkg/fmt/, so the fmt.Errorf does not log to stderr. I think long term we should use the cloud logging client. Need to give it a try and have some feeling on how much overhead it adds to the code.

util/util.go
shifucun: Good point, added to https://github.com/datacommonsorg/mixer/issues/347 and can do for the next PR

PR #352, opened 2020-10-22
Clear in-memory branch cache before reload

shifucun:
During branch cache reload, there are two instance of mem-cache which leads to OOM and server restart. Also clean up some logging for memory usage.
image


server/server.go
```diff
@@ -131,9 +130,9 @@ func (s *Server) SubscribeBranchCacheUpdate(
 	go func() {
 		err = sub.Receive(ctx, func(ctx context.Context, msg *pubsub.Message) {
 			gcsFolder := string(msg.Data)
-			fmt.Printf("Got message: %q\n", string(gcsFolder))
 			msg.Ack()
-			fmt.Println("Subscriber action: reload mem cache...")
+			fmt.Printf("Subscriber action: reload mem cache from %s\n", string(gcsFolder))
+			s.memcache.Update(map[string][]byte{})
```
shifucun, line 135: This is where memcache is cleared before reload @pradh

pradh (approved):
Nice find!

PR #353, opened 2020-10-22
Push 2020_10_22_01_02_39 with fixes for ranking bugs.

pradh:


golden_response/staging/get_landing_page_data/tha.json
```diff
@@ -2225,45 +2225,7 @@
...
-          "2013": 52.8,
-          "2014": 52.1,
-          "2015": 51.9,
-          "2016": 53.3,
-          "2017": 53.3,
-          "2018": 53.9
-        },
-        "provenanceDomain": "worldbank.org"
-      },
+      "GiniIndex_EcconomicActivity": {},
```
pradh, line 2228: This is a typo that needs to be fixed in https://github.com/datacommonsorg/website/blob/master/server/chart_config.json#L514

I'd fixed in google3 when I built cache.

beets (approved):
thanks for the ranking fixes!

pradh:
Thanks. I'll push so we can test on staging.

PR #357, opened 2020-10-24
Add a new API GetStat() that use proto field in the reponse

shifucun:
- This is an effort to remove the "payload" from API response and use actual fields.
- This API is not public, so when all internal clients are migrated to the new one, we can remove the old GetStats() API.
- Ranking functions and helper functions are based on protobuf now, this will help to add ObsCollection cache.


proto/mixer.proto
```diff
@@ -237,6 +237,37 @@ message GetStatsResponse {
   string payload = 2;
 }
 
+// GetStat is a newer version of GetStats API, that use protobuf type as return
+// field instead of a "payload" of json string.
```
pradh, line 241: nit: Its good to make the comment future proof. When we get rid of GetStats API, this will seem obsolete to future readers :)

proto/mixer.proto
```diff
@@ -237,6 +237,37 @@ message GetStatsResponse {
...
+// TODO(shifucun): Deprecate GetStats() once all clients are switched to use
+// this API.
+// Request message for GetStat.
+message GetStatRequest {
+  // The dcids of the place.
+  repeated string places = 1;
```
pradh, line 247: To confirm, are other places calling it dcids?

server/stat.go
```diff
@@ -0,0 +1,190 @@
+// Copyright 2020 Google LLC
```
pradh, line 1: I assume the logic here mimics the existing logic except for populating proto?

pradh (approved):

proto/mixer.proto
shifucun: All the new GetStat* API uses "places". Some of the old one like GetPopulation use "dcids". We decided to use "places" uniformly in the new API, since dcid is too general as there are statvar dcids in the request too.

server/stat.go
shifucun: Yes, they have the same logci.

proto/mixer.proto
shifucun: Done

shifucun:
Thanks for review!

PR #358, opened 2020-10-24
Add API to get stat for all child places of certain type given parent place

shifucun:
- Also moved ranking logic to a separate file


e2etest/get_stat_collection_test.go
```diff
@@ -0,0 +1,104 @@
...
+	}{
+		{
+			"geoId/06",
+			"County",
+			"2016",
+			[]string{"Count_Person", "Median_Age_Person"},
+			"CA_County_2016.json",
+		},
+		{
+			"country/USA",
```
shifucun, line 62: @pradh don't seem to have data for these two cases. Do we restrict on the the place types?

shifucun: per offline discussion, will add the support for not direct-linked place types in later PR

pradh (approved):

shifucun:
thanks for review!

PR #360, opened 2020-10-27
Add a flag for using ALTS authentication

shifucun:
ALTS won't work with Cloud Endpoints and should not be used by default.

We may need to create extra cluster that only holds the server container and enable ALTS there for testing prod <-> cloud


pradh (approved):

shifucun:
thanks for review

PR #361, opened 2020-10-27
Add deployment process for mixer grpc service without esp

shifucun:
Note the documentation change is meant to aid for a one-off deployment of a custom mixer, not for the DataCommons production mixer.

The current mixer deployment is bundled with esp, ie, every GKE pod has an esp and the mixer docker image and request is intercepted by esp. That said, it is an HTTP(S) service.

This PR adds documentation to bring up mixer only on GKE and exposed via GKE load balancer as a pure GRPC service.

Since the two shares some common configurations, this PR refactored the README and sample yaml files.


shifucun (commented):
Gental ping on this PR @pradh @tjann

deployment/service.yaml
```diff
@@ -18,7 +18,7 @@ metadata:
   name: mixer-service
   namespace: mixer
 spec:
-  type: NodePort
+  type: LoadBalancer
```
pradh, line 21: For my understanding, what does this do?

pradh (commented):
Sorry, I find it hard to understand what is really going on. Could you give a high-level overview of what is happening here? Specifically:

+ how is the mixer deployed now?
+ what does it mean for the mixer to be a pure grpc service with or without esp? any pointers to documentation?
+ are we saying we can deploy the mixer in either of these modes?

deployment/service.yaml
shifucun: Here are the concept for NodePort and LoadBalancer:
https://kubernetes.io/docs/concepts/services-networking/service/#nodeport
https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer

They are types of kubernetes "services". NodePort does not have load balancer functions, and only expose a uniform port. LoadBlancer provisions a load balancer.

In GRPC + ESP scenario, we also have an ingress* sits in front of the Service (NodeType), that has load balancing, in the pure GRPC scenario, we do need the Service itself has load balancing.

* https://kubernetes.io/docs/concepts/services-networking/ingress/#what-is-ingress

shifucun:
Thanks for review!

deployment/service.yaml
pradh: Thanks.

pradh (approved):
I see, so this is for one-off deployment.  Thanks for clarifying.

Typo in title s/And deployment/Add deployment/

PR #367, opened 2020-11-12
Releasing data from 2020_11_12_01_01_15

beets:
Mostly new covid data and some provenance changes
One new missing stat var ranking -- will update the config in g3 after this so next run can pick it up.


golden_response/staging/get_triples/place_type.json
```diff
@@ -364,7 +364,7 @@
       "objectTypes": [
         "Class"
       ],
-      "provenanceId": "dc/5j06ly1"
```
beets, line 367: @tjann do you understand these provenance updates?

tjann: There's some non determinism in picking which provenance to give for a geo. We have a couple sources contributing to basic geo facts. As long as it's switching between 2 reasonably authoritative sources, we are ok. Guha's just complained if the provenance looks like fbi.gov for name of California, etc.

tjann (approved):

PR #374, opened 2020-12-03
Update release to point to 2020_12_02_14_38_23

beets:
Do we know if the deletion of covid data for county is expected?


beets:
Abandoning this release due to issues in covid data...

PR #379, opened 2020-12-07
Add /place/stat-vars API (replacement for /place/stats-var)

shifucun:
Once all clients are migrated over, can remove "/place/stats-var".

Tested in Minikube and looking fine

### DIFF
--- a/proto/mixer.proto
+++ b/proto/mixer.proto
@@ -449,6 +449,24 @@ message GetPlaceStatsVarResponse {
   map<string, StatsVars> places = 1;
 }
 
+
+// StatVars represent a list of statistical variable dcids.
+message StatVars {
+  repeated string statVars = 1;
+}
+
+// Request message for GetPlaceStatVars API.
+message GetPlaceStatVarsRequest {
+  // DCIDs of the places.
+  repeated string dcids = 1;
+}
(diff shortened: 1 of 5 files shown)


beets (commented):
Is it possible to use the new request/response protos for the older API so we don't have to duplicate the handler code (and protos)?

proto/mixer.proto
```diff
@@ -449,6 +449,24 @@ message GetPlaceStatsVarResponse {
   map<string, StatsVars> places = 1;
 }
 
+
+// StatVars represent a list of statistical variable dcids.
+message StatVars {
+  repeated string statVars = 1;
```
pradh, line 455: snake case or camel case?

pradh (approved):

shifucun:
> Is it possible to use the new request/response protos for the older API so we don't have to duplicate the handler code (and protos)?

Consolidated the implementation code. The response can not be reused since the json field name changed from "statsVars" to "statVars".

proto/mixer.proto

pradh (approved):

beets (approved):
thanks!

shifucun:
Thanks for review!

PR #382, opened 2020-12-08
Add GetPlaceStatVarsUnion API

intrepiditee:
Add GetPlaceStatVarsUnion API at "/place/stat-vars/union"


shifucun (approved):
Looks good, please test locally and verify the latency and memory usage

PR #388, opened 2020-12-09
Break if we don't get results as we climb the parent tree

beets:
We might encounter ancestors with no parents (though it's a bug in the kg, we can still return something reasonable for the place pages)

For record keeping purposes: the issue we encountered was country/NLD not having an ancestor (and country/ANT instead being a country in europe).


shifucun (approved):
thx for the fix!

PR #389, opened 2020-12-10
Update data to version 2020_12_10_01_01_26

pradh:
This mainly includes new WDI data (tjann) and i18n data (hanluc).

From the diffs, besides the usual diffs from latest-covid-stats and non-determinism:

1. I'm not sure why "LifeExpectancy_Person_Female" is missing. Does it have to do with WDI change? @tjann 

2. State has label "nameWithLanguage" as expected.


golden_response/staging/get_pop_obs/moscow.json
```diff
@@ -1837,7 +1837,7 @@
           "observation_date": "2005",
           "observation_period": "P1Y",
           "stat_var_ids": [
-            "LifeExpectancy_Person_Female"
+            "dc/3bpjn5e953v0b"
```
tjann, line 1840: I've not been able to spot a visual difference between
https://datacommons.org/browser/LifeExpectancy_Person_Female
https://datacommons.org/browser/dc/3bpjn5e953v0b , and same for Male below; which are all currently in prod. Not sure why the stats got attached to the hash DCID instead of the human readable one-- does this have anything to do with import, or is it more an API behavior?

shifucun: This api directly reads from the popobs cache, so this should happened in flume. Is there any randomness in the statvar picking? But we never seen this before though.

pradh: OK, if its a flume thing, could be something I introduced. So, let me try the Tues cache...

pradh: Wait, we shouldn't be looking at the browser, rather we should look at the new cache/BQ.  The curated version has two new properties -- description, descriptionUrl.  I bet its because descriptionUrl is not listed in PropsThatAreNotConstraintProps().

To confirm, you can run the below cmd before and after cl/345715815, and you can see the DCID change from dc/3bpjn5e953v0b to dc/f3w8mmhekf9sf.

  # blaze test -c opt :statvar_checker_test --test_arg=--curated_statvar_id=LifeExpectancy_Person_Female

So, if LifeExpectancy_Person_Female and LifeExpectancy_Person_Male aren't used by any client, we can roll this out. Otherwise we should rework cl/345715815.

Sorry for missing this during review.

pradh: I'll make a change to add descriptionUrl to PropsThatAreNotConstraintProps().

tjann: Ah ok, if we make the change to PropsThatAreNotConstraintProps, there is no rework needed for cl/345715815? I'll check if it's used on surfaces now.

tjann (approved):
Life expectancy by gender is not used in place pages nor timeline

golden_response/staging/get_pop_obs/moscow.json
pradh: @shifucun -- does the mixer have such a list? Or does it rely on constraintProps?

shifucun: It depends on constraint props

pradh: Great, then just a pipeline change should suffice.

pradh:
> Life expectancy by gender is not used in place pages nor timeline

@tjann is there a chance you generated the data using stat-vars that didn't have descriptionUrl in it?

If you generated it with descriptionUrl, then we might need to re-resolve it after fix...

tjann:
> > Life expectancy by gender is not used in place pages nor timeline
> 
> @tjann is there a chance you generated the data using stat-vars that didn't have descriptionUrl in it?
> 
> If you generated it with descriptionUrl, then we might need to re-resolve it after fix...

I generated it with descriptionUrl. Ok, I can do that.

pradh:
> > > Life expectancy by gender is not used in place pages nor timeline
> > 
> > 
> > @tjann is there a chance you generated the data using stat-vars that didn't have descriptionUrl in it?
> > If you generated it with descriptionUrl, then we might need to re-resolve it after fix...
> 
> I generated it with descriptionUrl. Ok, I can do that.

In that case, I'm puzzled how the StatPop had showed up fine?  I'd expect it to have descriptionUrl as a cProp, does it?

tjann:
Sorry, you are right, there was no descriptionUrl.

The StatVars checked into g3 happened after all the resolutions, which were based on https://github.com/datacommonsorg/data/blob/ba69463d841a4a37b62b272021ca3bb4bddf052e/scripts/world_bank/output/WorldBank_StatisticalVariables_p1.mcf

PR #395, opened 2020-12-13
Use correct title for ESP

shifucun:
The title should be distinct. Use the endpoints url for it for the ease of debugging.

Currently it looks like:

image


deployment/template/endpoints.yaml.tmpl
```diff
@@ -21,7 +21,7 @@ config_version: 3
 name: "<SERVICE_NAME>"
 
 # API title to appear in the user interface (Google Cloud Console).
-title: Data Commons API (Testing)
+title: <API_TITLE>
```
beets, line 24: sorry, i don't follow where this string gets replaced (or what the final title would be). would it now also be $service_name?

shifucun: It would be replaced in https://github.com/datacommonsorg/mixer/pull/395/files#diff-1306a0330ffe59d264ca80671199aee3b508d384954c97adf36ca8374f8e58d6R41

Currently the. title looks like below. It's 1) not an API (exposed externally) and 2) not test..
image

deployment/minikube/generate_yaml.sh
```diff
@@ -38,5 +38,6 @@ yq w -i --style=double --inplace -- deployment.yaml spec.template.spec.container
 
 # ESP service configuration
 yq w --style=double ../template/endpoints.yaml.tmpl name $SERVICE_NAME > endpoints.yaml
+yq w -i endpoints.yaml title $service_name
```
beets, line 41: should this be all caps as well, like in previous line?
$SERVICE_NAME

deployment/template/endpoints.yaml.tmpl
beets: gotcha.

since it will be replaced with service_name, would it be clearer to do:
title: <SERVICE_NAME>

i also don't see name: getting updated in the script.

deployment/minikube/generate_yaml.sh
```diff
@@ -38,5 +38,6 @@ yq w -i --style=double --inplace -- deployment.yaml spec.template.spec.container
...
 
 # Deploy
```
beets, line 43: Is there a missing step after this comment? Or does another script take over after?

deployment/template/endpoints.yaml.tmpl
shifucun: good catch! 

I also realized this change is in the "mixer" repo, so it would be used by the standalone mixer as well. In that case, this title would be something like "Data Commons API - Staging(Prod)". And this title should be set in the configuration. It's still written as env.sh, which I add a todo to move to yaml file.

"website" repo import this file as sub-module. Once this PR is submitted, I will change the part in "website" repo

deployment/minikube/generate_yaml.sh
shifucun: Removed as this is stale. Another script will take over.

deployment/minikube/generate_yaml.sh
shifucun: updated.

beets (approved):

PR #413, opened 2020-12-27
Add mixer GKE setup scripts and documentation

shifucun:


gke/config.yaml
```diff
@@ -0,0 +1,21 @@
+# Copyright 2020 Google LLC
```
beets, line 1: Is this file a template? Should we rename the file to indicate that (I think we have a few already), and make it clear in the instructions to create a new copy of this file?

beets (approved):

gke/config.yaml
shifucun: Made this a template file and updated instructions.

shifucun:
thanks for review!

PR #418, opened 2020-12-29
Update prod golden files after 2020_12_28_12_49_30 release

tjann:


tjann:
We don't want to touch prod this week

Issue #412, opened 2020-12-27
Re-structure project layout


shifucun:
Current project has a flat layout that has Golang packages, cloud build files, bash scripts on the top level folder. At this point, it becomes hard to add deployment framework like [Skaffold](https://skaffold.dev/docs/quickstart/). Having a better structure would ease the code maintain burden as well.

Here propose to use a standard [Golang layout](https://github.com/golang-standards/project-layout) for the mixer repo. The repo structure would look like the following:
.
+-- build/
+-- cmd/
+-- config/
+-- deploy/
+-- docs/
+-- esp/
+-- examples/
+-- gke/
+-- internal/
|   +-- base/
|   +-- proto/
|   +-- sparql/
|   +-- translator/
|   +-- util/
+-- proto/
+-- scripts/
+-- test/

PR #419, opened 2020-12-29
Move integration (former e2e) test into a new top level folder

shifucun:
This is part of effort to re-structure the mixer folder in https://github.com/datacommonsorg/mixer/issues/412.

There is no code change other than package rename and path change.

Note the name change to integration reflects the fact that the test only covers gRPC part not the proto to json transcoding with ESP, hence change the name from "e2e" to "integration"


shifucun:
@pradh It would be easier to review this in Github UI

pradh (approved):

PR #421, opened 2020-12-30
Parallel integration tests to speed up from ~2min to 10s

shifucun:
- Add t.Parallel() to all the integration tests.
- The TestChartConfigRankings ranking test have two for loops over place type (4) and chart config (>100) which takes about 60 seconds to run. Use goroutine to parallel at config level so the O(100) request can emit concurrently. Hence needs to sort the final result by chart title and updating the golden response.

---
This change is [Reviewable](https://reviewable.io/reviews/datacommonsorg/mixer/421)

### DIFF
--- a/test/integration/place_statvar_ranking_test.go
+++ b/test/integration/place_statvar_ranking_test.go
@@ -59,7 +59,9 @@ func readChartConfig() ([]Chart, error) {
 	return config, nil
 }
 
-func getMissingStatVarRanking(client pb.MixerClient, req *pb.GetLocationsRankingsRequest) ([]string, error) {
+func getMissingStatVarRanking(
+	client pb.MixerClient,
+	req *pb.GetLocationsRankingsRequest) ([]string, error) {
 	ctx := context.Background()
 	response, err := client.GetLocationsRankings(ctx, req)
 	if err != nil {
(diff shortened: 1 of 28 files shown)


test/integration/place_statvar_ranking_test.go
```diff
@@ -114,59 +117,72 @@ func TestChartConfigRankings(t *testing.T) {
...
+					// Test main chart rankings
+					req := &pb.GetLocationsRankingsRequest{
+						PlaceType:    c.placeType,
+						WithinPlace:  c.parentPlace,
+						StatVarDcids: chart.StatsVars,
+						IsPerCapita:  len(chart.Denominator) > 0,
+					}
+					missingStatVars, err := getMissingStatVarRanking(client, req)
+					if err != nil {
+						t.Errorf("Error fetching rankings for chart %s: %s", chart.Title, c.placeType)
```
pradh, line 141: nit: is the line too long?

pradh (approved):
Nice!

test/integration/place_statvar_ranking_test.go
shifucun: Move long lines to multi-lines

shifucun:
thanks for quick review!

PR #423, opened 2020-12-31
Deploy config and scripts update. This prepares for GKE deployment to staging/prod

shifucun:


script/deploy_gke.sh
```diff
@@ -0,0 +1,56 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# Script to deploy mixer to a GKE cluster.
+# This retrives the docker images and gRPC descriptor based on git commit hash,
+# so these binaries should have been pushed to container registry and Cloud
+# Storage.
```
beets, line 20: Is there a script that needs to be run to do this? Would be good to reference that here.

beets: Might be good to also add a warning to do this from a clean repo, and to fetch the latest from upstream (or maybe the script should do that as well?)

beets: Also check that the ENV is supplied (and exit script if not specified or invalid). The header comment should include possible options.

beets (approved):

script/deploy_gke.sh
shifucun: Updated accordingly

shifucun:
thanks for the review!

PR #427, opened 2021-01-06
Should not have newline in bigquery and bigtable version file

shifucun:
These are files used to generate kubernetes config and the newline will be treated as "\n": https://stackoverflow.com/questions/56395758/what-is-the-difference-between-and-when-creating-a-configmap-from-a-file-i

Need to find a way to pre-submit this.


beets (approved):
ouch - nice find!

PR #429, opened 2021-01-07
Update data version to 2021_01_07_01_01_54

pradh:
This includes refreshes for Census ACS and WDI to include 2019 data.

For WDI, it is expected that older data may change (because WDI produces full historical data every year).


beets (approved):
I'm surprised that the data changes for even 1970 in some cases. Though the actual deltas are very small.

pradh:
> I'm surprised that the data changes for even 1970 in some cases. Though the actual deltas are very small.

Yup, per https://datahelpdesk.worldbank.org/knowledgebase/articles/114939-how-are-revisions-managed, they may apparently change estimation methods, etc.

For the 1970 case:
-          "1970": 2927.0729581373,
+         "1970": 2927.07293946175,

To confirm, this matches source:
image

test/integration/golden_response/staging/get_stat_all/result.json
```diff
@@ -121,15 +122,16 @@
...
+                "2010": 309321666,
+                "2011": 311556874,
+                "2012": 313830990,
+                "2013": 315993715,
+                "2014": 318301008,
+                "2015": 320635163,
+                "2016": 322941311,
+                "2017": 324985539,
+                "2018": 326687501,
+                "2019": 328239523
```
tjann, line 134: Wow this is surprising... we were missing 10 years before?

beets:
Sounds good - thanks for checking (and the pointer!)

tjann (approved):
LG

shifucun (approved):

test/integration/golden_response/staging/get_stat_all/result.json
pradh: The delta is showing up poorly, past 10 years just changed, but existed?

pradh (commented):
Thanks for the reviews!

test/integration/golden_response/staging/get_stat_all/result.json
tjann: Oh I see. Thanks!

PR #431, opened 2021-01-08
check for version file end of file newline

shifucun:


beets (approved):
would be nice to catch this locally as well, but this is nice check

shifucun:
> would be nice to catch this locally as well, but this is nice check

Yeah, I would keep looking for solutions to strip the newline

build/ci/cloudbuild.test.yaml
```diff
@@ -14,6 +14,26 @@
...
 # Build protobuf to go bindings
+- id: lint
+  name: golang
+  entrypoint: 'bash'
+  args:
+  - -c
+  - |
+    set -e
+    c=`tail -c 1 deploy/storage/bigtable.version`
+    if [ "$c" == "" ]; then
```
pradh, line 25: would the following be a stronger check? i.e., no newline at all in the file vs. newline not the last character of file

```
c=`grep -c '\n' deploy/storage/bigtable.version`
if ["$c" != "0" ]; then
  ...
```

pradh (approved):
Thanks for the fix!

build/ci/cloudbuild.test.yaml
shifucun: That's good idea. I tried grep '\n' but somehow it always give 0, even for files with multi line. Switched to use `cat deploy/storage/bigtable.version | wc -l`

pradh: LG.. Sorry about that, per https://tldp.org/LDP/Bash-Beginners-Guide/html/sect_04_02.html the reason is `Since newline is also a separator for the list of patterns, there is no way to match newline characters in a text.`

shifucun: thanks, good to know!

PR #432, opened 2021-01-08
Enable branch cache by default

shifucun:


cmd/main.go
```diff
@@ -95,13 +94,10 @@ func main() {
...
-		if branchCacheFolder == "" {
-			branchCacheFolder, err = server.ReadBranchCacheFolder(
-				ctx, branchCacheBucket, branchCacheVersionFile)
-			if err != nil {
-				log.Fatalf("Failed to read branch cache folder: %v", err)
-			}
+		branchCacheFolder, err := server.ReadBranchCacheFolder(
+			ctx, branchCacheBucket, branchCacheVersionFile)
+		if err != nil {
+			log.Fatalf("Failed to read branch cache folder: %v", err)
```
pradh, line 100: Should this be fatal or more graceful?

pradh (approved):
Thanks for the change!

cmd/main.go
shifucun: there is no monitoring on errors yet, for now can make it fatal so we know the branch cache load has issue?

pradh: ok. and if bad things happen we just set enableBranchCache to false...

shifucun: sg

PR #434, opened 2021-01-10
Improve Skaffold local dev process

shifucun:
- disable branch cache as the readiness probe will fail on Minikube (not sure why)
- make log working
- various fix for branch cache reload

### DIFF
--- a/deploy/overlays/dev/patch_esp.yaml
+++ b/deploy/overlays/dev/patch_esp.yaml
@@ -12,19 +12,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+  # Disable branch cache for mixer
 - op: add
   path: /spec/template/spec/containers/0/args/0
   value: --enable_branch_cache=false
+  # Do not use service for esp
 - op: remove
   path: /spec/template/spec/containers/1/args/0
-- op: remove
-  path: /spec/template/spec/containers/1/args/0
+  # Add service json path for esp
 - op: add
   path: /spec/template/spec/containers/1/args/0
-  value: --service_json_path
+  value: --service_json_path=/esp/mixer-grpc.json
+  # esp runs in non_gcp mode
 - op: add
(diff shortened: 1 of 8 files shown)


docs/developer_guide.md
```diff
@@ -41,10 +41,10 @@ which is compiled using [API Compiler](https://github.com/googleapis/api-compile
 ### Start mixer in minikube
 
 ```bash
-minikube start
+minikube start --meory 6G
```
pradh, line 44: memory?

deploy/overlays/dev/patch_esp.yaml
```diff
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+- op: add
+  path: /spec/template/spec/containers/0/args/0
```
pradh, line 16: for my understanding, what are these indexes - 0/args/0, 1/args/0 ?

pradh (approved):

deploy/overlays/dev/patch_esp.yaml
shifucun: These corresponding to the deployment.yaml json tree. For example, here refers to the first container in deployment.yaml, and add an argument in the first position: https://github.com/datacommonsorg/mixer/blob/master/deploy/base/deployment.yaml#L53

docs/developer_guide.md
shifucun: good catch!

shifucun:
thanks for review!

deploy/overlays/dev/patch_esp.yaml
pradh: Wow, seems brittle. Add a comment?

shifucun: Done. And cleaned up the args

PR #435, opened 2021-01-12
Push version 2021_01_11_01_01_38

pradh:
This includes the following changes:

- SuperEnum/Place aggregation with 2019 data from Tiffany
- Ranking regeneration with 2019 data

Many of the values in the ranking info changes to the latest value. There are a few previously missing stat-vars that not missing. I'm not entirely sure where that comes from, perhaps because of a long pending ranking regen.


tjann (approved):
Interesting @ the no longer missing rankings 🤔 🤷

beets (approved):
nice!

pradh:
Thanks!

PR #438, opened 2021-01-14
Implement /v1/stat/set/series API

shifucun:
- This is to replace /stat/series/rich API which was implemented for 1 sv * N places, but never used. Below is the prod mixer stats for the past 30 days, there is 0 request to /stat/series/rich
image

- This API takes M sv and N places, return the best ranked stat series for each [sv, place] combination
- This does not take optional argument for filtering. If user is interested in more sources, should use /v1/stat/set/series/all (to be implemented)
- Set version to /v1 
- Sample response: https://github.com/datacommonsorg/mixer/pull/438/files#diff-2e286a38eb97377f678f295dfd4d987abca0302c8fe33122e51166705eb4d8e3
- Once this is in, can migrate browser to use this API, and deprecate /bulk/stats endpoints


internal/server/stat_series.go
```diff
@@ -288,91 +288,88 @@ func (s *Server) GetStats(ctx context.Context, in *pb.GetStatsRequest) (
...
 		rowList,
 		convertToObsSeriesPb,
 		tokenFn(keyTokens),
 	)
+
 	for token, data := range cacheData {
-		place := strings.Split(token, "^")[0]
+		parts := strings.Split(token, "^")
+		place := parts[0]
+		statVar := parts[1]
```
pradh, line 348: To confirm, what cache is this that has stat-var ID as second part?

proto/stat.proto
```diff
@@ -115,34 +126,22 @@ message GetStatsResponse {
   string payload = 2;
 }
 
-// Request for GetStat API.
+// Request for GetStatSetSeries API.
 // TODO(shifucun): Deprecate GetStats() once all clients are switched to use
 // this API.
```
pradh, line 131: TODO still relevant?

proto/stat.proto
```diff
@@ -58,6 +58,17 @@ message SourceSeries {
...
 
+message Series {
+  // Map from date to stat value.
+  map<string, double> val = 1;
+  // Series metadata.
+  StatMetadata metadata = 2;
+}
+
+message SeriesMap {
+  // Map from key to series. The key could be place dcid or stat var dcid.
```
pradh, line 69: Depending on...?

proto/stat.proto
```diff
@@ -58,6 +58,17 @@ message SourceSeries {
   string provenance_url = 9;
 }
 
+message Series {
```
pradh, line 61: Add comment? Represents a time-series from a source...

pradh (approved):

internal/server/stat_series.go
shifucun: The place and statvar here is from the request. They are passed into the memcache parallel read so the read-cache can bind back to the original <place, statvar>

proto/stat.proto
shifucun: Done

proto/stat.proto
shifucun: Updates the comment to make it more generic. What is key is really depending on the response structure so not worth emphasize in here.

proto/stat.proto
shifucun: Yep, website still uses GetStats(), need to do the migration there

shifucun (commented):
thanks for review!

PR #439, opened 2021-01-15
Add a flag so mixer can run in BigQuery only mode

shifucun:
This is to allow mixer to only connect to BigQuery and run Sparql without setting up BigTable.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -45,7 +45,7 @@ var (
 	port              = flag.Int("port", 12345, "Port on which to run the server.")
 	useALTS           = flag.Bool("use_alts", false, "Whether to use ALTS server authentication")
 	enableBranchCache = flag.Bool("enable_branch_cache", true, "Whether to use branch cache")
-	sparqlOnly        = flag.Bool("sparql_only", true, "The service only serves sparql query")
+	bigqueryOnly      = flag.Bool("bigquery_only", true, "The service only serves sparql query")
 )
 
 const (
@@ -82,7 +82,7 @@ func main() {
 	}
 
 	var btTable *bigtable.Table
-	if *sparqlOnly {
+	if *bigqueryOnly {
 		btTable = nil
 	} else {
 		// BigTable.
@@ -99,7 +99,7 @@ func main() {
 	}
 
 	memcache := &server.Memcache{}
-	if !*sparqlOnly && *enableBranchCache {
+	if !*bigqueryOnly && *enableBranchCache {
 		// Memcache
 		branchCacheFolder, err := server.ReadBranchCacheFolder(
 			ctx, branchCacheBucket, branchCacheVersionFile)
@@ -117,7 +117,7 @@ func main() {
 	s := server.NewServer(bqClient, btTable, memcache, metadata)
 
 	// Subscribe to cache update
-	if !*sparqlOnly && *enableBranchCache {
+	if !*bigqueryOnly && *enableBranchCache {
 		err = s.SubscribeBranchCacheUpdate(
 			ctx, pubsubProject, branchCacheBucket, subscriberPrefix, pubsubTopic)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


pradh (approved):

pradh:
Can this be validated locally (i.e., if such a mixer receives non-sparql calls) ?

cmd/main.go
```diff
@@ -37,13 +38,14 @@ import (
...
 	bqDataset         = flag.String("bq_dataset", "", "DataCommons BigQuery dataset.")
-	btTable           = flag.String("bt_table", "", "DataCommons Bigtable table.")
+	btTableName       = flag.String("bt_table", "", "DataCommons Bigtable table.")
 	btProject         = flag.String("bt_project", "", "GCP project containing the BigTable instance.")
 	btInstance        = flag.String("bt_instance", "", "BigTable instance.")
 	projectID         = flag.String("project_id", "", "The cloud project to run the mixer instance.")
 	port              = flag.Int("port", 12345, "Port on which to run the server.")
 	useALTS           = flag.Bool("use_alts", false, "Whether to use ALTS server authentication")
 	enableBranchCache = flag.Bool("enable_branch_cache", true, "Whether to use branch cache")
+	sparqlOnly        = flag.Bool("sparql_only", true, "The service only serves sparql query")
```
pradh, line 48: Sorry one more question on the flag... should we call it sparqlOnly or bigqueryOnly?

Because, in bigqueryOnly mode, we can enable all APIs, in theory.

shifucun:
> Can this be validated locally (i.e., if such a mixer receives non-sparql calls) ?

non-sparql  would crash the server. I am debating on adding checks in a lot of files, or leave it as is since this is a custom server that the user should **only** run sparql query

pradh:
> > Can this be validated locally (i.e., if such a mixer receives non-sparql calls) ?
> 
> non-sparql would crash the server. I am debating on adding checks in a lot of files, or leave it as is since this is a custom server that the user should **only** run sparql query

Maybe have a single common function that you can call everywhere? In case someone tries our open API on it, it might be good to not crash...

cmd/main.go
shifucun: change it to bigquery_only. this is easy for client to realize it's only talking to bt, but also it can only support /query API for now

shifucun:
> > > Can this be validated locally (i.e., if such a mixer receives non-sparql calls) ?
> > 
> > 
> > non-sparql would crash the server. I am debating on adding checks in a lot of files, or leave it as is since this is a custom server that the user should **only** run sparql query
> 
> Maybe have a single common function that you can call everywhere? In case someone tries our open API on it, it might be good to not crash...

Add nil check

internal/server/bigtable.go
```diff
@@ -52,6 +52,22 @@ func bigTableReadRowsParallel(
...
+	if btTable == nil {
+		for _, key := range rowList {
+			token, err := getToken(key)
+			if err != nil {
+				return nil, err
+			}
+			result[token] = nil
+		}
+		return result, nil
+	}
```
pradh, line 69: should we add some e2e test for this mode? (later PR is ok)

internal/server/property_label.go
```diff
@@ -52,6 +52,10 @@ func (s *Server) GetPropertyLabels(ctx context.Context,
 	}
 	result := map[string]*PropLabelCache{}
 	for dcid, data := range dataMap {
+		if data == nil {
+			result[dcid] = nil
+			continue
+		}
```
pradh, line 58: So, other callsites of bigTableReadRowsParallel don't have this problem?

pradh (approved):

internal/server/property_label.go
shifucun: There are a bunch of type conversion unchecked. Fixed them all in https://github.com/datacommonsorg/mixer/pull/440

internal/server/bigtable.go
shifucun: Added unittest and e2e test

shifucun:
thanks for review!

PR #440, opened 2021-01-15
check type conversion errors

shifucun:
There are quite some instances of un-checked type conversion. These can potentially lead to server crash.

Use s, ok: = data(.DataType) can deal with a nil "data", so no need for a precedent data == nil check

---
This change is [Reviewable](https://reviewable.io/reviews/datacommonsorg/mixer/440)

### DIFF
--- a/internal/server/landing_page.go
+++ b/internal/server/landing_page.go
@@ -248,7 +248,10 @@ func fetchBtData(
 	// Populate result from landing page cache
 	result := map[string]map[string]*ObsTimeSeries{}
 	for dcid, data := range dataMap {
-		landingPageData := data.(*LandingPageData)
+		landingPageData, ok := data.(*LandingPageData)
+		if !ok {
+			continue
+		}
 		finalData := map[string]*ObsTimeSeries{}
 		for statVarDcid, obsTimeSeries := range landingPageData.Data {
 			obsTimeSeries.filterAndRank(&ObsProp{})
--- a/internal/server/property_value.go
+++ b/internal/server/property_value.go
@@ -17,6 +17,7 @@ package server
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 
 	mapset "github.com/deckarep/golang-set"
 	"google.golang.org/grpc/codes"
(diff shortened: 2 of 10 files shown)


internal/server/property_value.go
```diff
@@ -125,7 +126,10 @@ func getPropertyValuesHelper(
 			return propVals.Nodes, nil
 		}, nil)
 	for dcid := range branchNodeMap {
-		branchNodes := branchNodeMap[dcid].([]*Node)
+		branchNodes, ok := branchNodeMap[dcid].([]*Node)
+		if !ok {
+			return nil, fmt.Errorf("branch nodes invalid for %s", dcid)
```
pradh, line 131: For my understanding, why do we return error in this case, while silently putting in nil in others?

internal/server/landing_page.go
```diff
@@ -248,7 +248,10 @@ func fetchBtData(
 	// Populate result from landing page cache
 	result := map[string]map[string]*ObsTimeSeries{}
 	for dcid, data := range dataMap {
-		landingPageData := data.(*LandingPageData)
+		landingPageData, ok := data.(*LandingPageData)
```
pradh, line 251: To confirm, are these changes necessarily related to bigqueryOnly mode?

If so, why don't we return an ERROR in bigTableReadRowsParallel interface? Wouldn't that be simpler (since we already handle it)?

shifucun: This is not entirely related to that. In general, we want to check if the conversion is ok to prevent crash, and we did in some places (to check the nil pointer).

pradh (approved):
> *[Reviewable](https://reviewable.io/reviews/datacommonsorg/mixer/440)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on @shifucun)

___
*[internal/server/landing_page.go, line 251 at r1](https://reviewable.io/reviews/datacommonsorg/mixer/440#-MRkT4BoZqfICREaHxtl-r1-251:-MRkWP04CjBf3mfzAP51:b-yg3o5r) ([raw file](https://github.com/datacommonsorg/mixer/blob/60e159ee479a4dc61ea0a2f9eba01979ac673b58/internal/server/landing_page.go#L251)):*
<details><summary><i>Previously, shifucun (Bo Xu) wrote…</i></summary><blockquote>

This is not entirely related to that. In general, we want to check if the conversion is ok to prevent crash, and we did in some places (to check the nil pointer).
</blockquote></details>

Got it. Trying to understand why you return error in some cases and silently put in nil in others.

shifucun:
Close this PR as the files are too out-of dated. Created a new PR https://github.com/datacommonsorg/mixer/pull/535

PR #449, opened 2021-01-21
Pull base/branch cache merge logic out of bigtable parallel read function

shifucun:
Currently bigTableReadRowsParallel prefers branch cache data over base cache data but this is only true for stats series and collections cache. For property-value, property-labels and triples cache, this is the opposite.

This PR removes data merge logic from bigTableReadRowsParallel, so it only returns raw base and branch cache data. Merge logic is implemented at each API.

For property value and triples cache, only base cache is used to avoid expensive merge. This works for now as branch cache only contains "incremental" stats data but not schema. Merge triples cache is doable but expensive. When merge is definitely necessary, we should consider change the cache layout from plain list to a map keyed by property.

---
This change is [Reviewable](https://reviewable.io/reviews/datacommonsorg/mixer/449)


beets (commented):
Thanks Bo.

Could we add some tests to capture the intricacies of these changes? One that comes to mind is for parentPlaces (getPropertyValues for containedInPlace)

> *[Reviewable](https://reviewable.io/reviews/datacommonsorg/mixer/449#-:-MRfRb3G3fVL8JpMkScj:b3akwky)* status: 0 of 15 files reviewed, all discussions resolved (waiting on @pradh)

shifucun:
> Thanks Bo.
> 
> Could we add some tests to capture the intricacies of these changes? One that comes to mind is for parentPlaces (getPropertyValues for containedInPlace)
> 
> > _[Reviewable](https://reviewable.io/reviews/datacommonsorg/mixer/449#-:-MRfRb3G3fVL8JpMkScj:b3akwky)_ status: 0 of 15 files reviewed, all discussions resolved (waiting on @pradh)

Add test for property-values, with mock branch cache

beets (approved):
thanks for adding the test!

shifucun:
thanks for review!

PR #450, opened 2021-01-23
deploy website autopush instance on mixer commit

shifucun:


build/ci/cloudbuild.push.yaml
```diff
@@ -73,6 +73,28 @@ steps:
...
+      mkdir src
+      git clone https://github.com/datacommonsorg/website.git src
+      cd src
+      git clone https://github.com/datacommonsorg/mixer.git mixer
+      git submodule update --init --recursive
+      cd mixer
+      git remote set-url origin https://github.com/datacommonsorg/mixer.git
+      git pull origin master
+      git checkout master
+      gsutil cp gs://automation_control/latest_base_cache_version.txt deploy/storage/bigtable.version
```
pradh, line 93: wondering about bigquery version... website doesn't use it so this is ok.

but is there is a separate mixer version of this? that might need bigquery

build/ci/cloudbuild.push.yaml
```diff
@@ -73,6 +73,28 @@ steps:
       - |
         ./script/update_staging_version.sh $SHORT_SHA
 
+  # This is a trigger to push website at website/mixer code HEAD and with the
+  # latest base cache.
+  - name: 'gcr.io/datcom-ci/deploy-tool'
+    entrypoint: 'bash'
```
pradh, line 79: for my understanding, is there a way to invoke a bash-script (which then can be separately tested) instead of embedding stuff inline?

beets (approved):

hanlu09205 (approved):

pradh (approved):

build/ci/cloudbuild.push.yaml
shifucun: I think in general would be helpful to have a bigquery version file in gcs as well. Can add it here.

Mixer does not have autopush yet

build/ci/cloudbuild.push.yaml
shifucun: It can be put in a bash script. That would be helpful If shared in multiple place, like "./script/deploy_gke.sh" is used here

shifucun (commented):
thanks for review!

build/ci/cloudbuild.push.yaml
pradh: Ok, I'll make the change to write bigquery file.

build/ci/cloudbuild.push.yaml
pradh: Then I might suggest preferring shell-scripts over inline stuff... for testability.

PR #452, opened 2021-01-25
Prepare Data Release Version 2021_01_24_11_55_42

pradh:
Countries-in-Earth ranking for below-poverty is computed now. But it might be incomplete since we don't have data for all countries. Might need to do something about it...


beets (approved):
thanks for catching that ranking issue, as well as the other fixes!

test/integration/golden_response/staging/get_triples/place_type.json
```diff
@@ -4945,7 +4945,7 @@
       "objectTypes": [
         "Class"
       ],
-      "provenanceId": "dc/sm3m2w3"
+      "provenanceId": "dc/5n63hr1"
```
shifucun, line 4948: would be nice to find a way to stabilize the provenance..

shifucun (approved):

test/integration/golden_response/staging/get_triples/place_type.json
pradh: Yeah, that is a long-standing TODO... I'll look into it this week.

pradh:
Thanks for the review!

hanlu09205 (commented):
> What's the expectation from reviewers?
> I don't know whether those changes are expected?
> Is it if only I changed data, then make sure any golden change related to that looks good?
> Otherwise, reviewer don't need to worry about these value change?

Yes, in case one of us changed something. Or, since we still have noisy diffs (due to non-determinism in pipelines that need to fixed), this is also to get more pairs of eyes looking at the diffs.

For the ranking diffs, there are some cases where data is new and some non-determinism.  So I check that the ranks haven't changed very much (in which case its a red-flag).

PR #453, opened 2021-01-26
Data release version 2021_01_26_01_02_32

shifucun:


test/integration/golden_response/staging/get_stat_ranking/california.json
```diff
@@ -332,12 +332,12 @@
           {
             "rank": 7,
             "value": 50.5,
-            "placeDcid": "geoId/06057"
+            "placeDcid": "geoId/06005"
```
shifucun, line 335: should we expect the ranking to change from code update? @pradh

pradh: This is basically a swap of ranks when value is same (7<->8) due to some non-determinism in ranking that we discussed yesterday in https://github.com/datacommonsorg/mixer/pull/452

pradh (approved):
Thanks for the PR!

PR #457, opened 2021-01-27
Fix ranking test from chart config change

shifucun:
We may want to move this test to website repo? @beets


beets (approved):
I think it should stay in mixer since it's data related. Though the test can be set up better to ignore the title fields and only look for statsVars. It's an odd test really, in that we want this file to be empty..

shifucun:
> I think it should stay in mixer since it's data related. Though the test can be set up better to ignore the title fields and only look for statsVars. It's an odd test really, in that we want this file to be empty..

checking only for statvar existence sg

PR #458, opened 2021-02-05
release 2021_02_05_04_32

hanlu09205:


beets (commented):
The new covid data lgtm, but I wasn't sure we were expecting ranking updates? @pradh

pradh:
> The new covid data lgtm, but I wasn't sure we were expecting ranking updates? @pradh

We now run the ranking pipeline weekly (to pick up new changes with covid, etc). And the ranking diffs are due to some non-determinism (in my TODO list)...

pradh (approved):

beets (approved):

hanlu09205:
Thanks for the review!

PR #461, opened 2021-02-10
Add process for setup custom mixer

shifucun:


docs/custom_mixer.md
```diff
@@ -0,0 +1,41 @@
...
+
+* Remove all the .mcf files from [mapping](../deploy/mapping) directory and add
+  the custom template MCF files created above in that directory.
+
+* Change the mapping customization [file](../deploy/mapping/kustomization.yaml)
+  to use the custom MCF files.
+
+* Update the bigquery version [file](../deploy/storage/bigquery.version) to use
+  the custom dataset.
+
```
pradh, line 32: So, typically would people start by "forking" the mixer over and then making these changes?

And how will they patch new updates/bug-fixes to it?

gke/config.yaml.tpl
```diff
@@ -22,4 +22,4 @@ domain:
 api_title:
 region:
 nodes: 1
-store: google.com:datcom-store-dev
+store:
```
pradh, line 25: nit <PROJECT_NAME> ?

pradh (approved):

gke/config.yaml.tpl
shifucun: Done

docs/custom_mixer.md
shifucun: These are config changes to the particular project. If people would contribute to the common code part, should be able to follow the regular git PR process.

shifucun (commented):
thanks for review!

docs/custom_mixer.md
pradh: I understand this is specific to a particular project. My question was how those people will track the configs? If they create a fork of our repo, and make local changes, then keeping it up to date with our fixes is a pain. What's our recommendation?

shifucun: Normally forked repo owner would do "git pull dc master --rebase" to pull from the base repo. I moved the custom files all to a new folder so user won't see conflict.

deploy/overlays/custom/kustomization.yaml
```diff
@@ -0,0 +1,53 @@
...
+configMapGenerator:
+- name: mixer-configmap
+  behavior: create
+  literals:
+    - projectId=<PROJECT_ID>
+    - serviceName=<DOMAIN>
+- name: store-configmap
+  behavior: create
+  files:
+  - bigtable.version
```
pradh, line 35: Should we remove bigtable.version, since external people won't have it?

gke/README.md
```diff
@@ -18,7 +18,9 @@ Instructions of the initial setup of the GCP project and GKE clusters for deploy
   cp config.yaml.tpl config.yaml
   ```
 
-* Add the GCP project id in `config.yaml`, **project** field.
+* Add the GCP project id used for deployment in `config.yaml`, **project** field.
+
+* Add the GCP project that hosts the Bigquery table in `config.yaml`, **store** field.
```
pradh, line 23: Just curious... like the other configs in custom/ directory, should config.yaml be separated out as datacommons/ vs. custom/ ?

docs/custom_mixer.md
pradh: Makes sense. Is it worth adding a section to fork source repo and make changes locally, maybe before Setup GKE ?

pradh (approved):

deploy/overlays/custom/kustomization.yaml
shifucun: The base deployment.yaml requires "bigtable.version" config. we can separate bt and bq to two config but I don't want to make it too complicated since this is a noop for custom deploy

gke/README.md
shifucun: This folder is for general setup and config.yaml is not checked in. The specific configs for each env is stored in https://github.com/datacommonsorg/mixer/tree/master/deploy/gke

docs/custom_mixer.md
shifucun: Done

PR #462, opened 2021-02-10
Add deployment config for encode mixer

shifucun:


deploy/mapping/encode.mcf
```diff
@@ -0,0 +1,33 @@
+Node: E:EncodeBedLine->E1
```
pradh, line 1: I think we should add the BedLine mappings to base.mcf, and use that for encode KG? 

Since the tables are all the same.

pradh (commented):
Thanks for the quick PR!

deploy/mapping/encode.mcf
shifucun: This is included in main and encode kg. Note there are 3 mapping files base.mcf, weather.mcf and encode.mcf. Separate them out so it's easy to maintain.

pradh: Ah, got it...

pradh (approved):

PR #475, opened 2021-03-08
Add a mixer flag for store project and simplify other flags

shifucun:
This is to prepare for the stat var obs migration

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -37,25 +37,23 @@ import (
 )
 
 var (
-	bqDataset        = flag.String("bq_dataset", "", "DataCommons BigQuery dataset.")
-	btProject        = flag.String("bt_project", "", "GCP project containing the BigTable instance.")
-	baseBtInstance   = flag.String("base_bt_instance", "", "Base cache BigTable instance.")
-	baseTableName    = flag.String("base_table", "", "Base cache Bigtable table.")
-	branchBtInstance = flag.String("branch_bt_instance", "", "Branch cache BigTable instance.")
-	projectID        = flag.String("project_id", "", "The cloud project to run the mixer instance.")
-	port             = flag.Int("port", 12345, "Port on which to run the server.")
-	useALTS          = flag.Bool("use_alts", false, "Whether to use ALTS server authentication")
-	bigqueryOnly     = flag.Bool("bigquery_only", false, "The service only serves sparql query")
-	svobsMode        = flag.Bool("svobs_mode", false, "Use new storage that is built on top of StatVar instead of StatPop")
-	schemaPath       = flag.String("schema_path", "/translator/mapping", "The director that contains the schema mapping files")
+	mixerProject  = flag.String("mixer_project", "", "The cloud project to run the mixer instance.")
+	storeProject  = flag.String("store_project", "", "GCP project stores Bigtable and BigQuery.")
+	bqDataset     = flag.String("bq_dataset", "", "DataCommons BigQuery dataset.")
+	baseTableName = flag.String("base_table", "", "Base cache Bigtable table.")
+	port          = flag.Int("port", 12345, "Port on which to run the server.")
+	useALTS       = flag.Bool("use_alts", false, "Whether to use ALTS server authentication")
+	bigqueryOnly  = flag.Bool("bigquery_only", false, "The service only serves sparql query")
+	svobsMode     = flag.Bool("svobs_mode", false, "Use new storage that is built on top of StatVar instead of StatPop")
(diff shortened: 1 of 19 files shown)


cmd/main.go
```diff
@@ -77,8 +75,15 @@ func main() {
 		}
 	}
 
+	var branchCacheVersionBucket string
+	if *svobsMode {
+		branchCacheVersionBucket = "datcom-control"
+	} else {
+		branchCacheVersionBucket = "prophet_cache"
```
pradh, line 82: This should be "automation_control", no?

cmd/main.go
```diff
@@ -37,25 +37,23 @@ import (
...
-	schemaPath       = flag.String("schema_path", "/translator/mapping", "The director that contains the schema mapping files")
+	mixerProject  = flag.String("mixer_project", "", "The cloud project to run the mixer instance.")
+	storeProject  = flag.String("store_project", "", "GCP project stores Bigtable and BigQuery.")
+	bqDataset     = flag.String("bq_dataset", "", "DataCommons BigQuery dataset.")
+	baseTableName = flag.String("base_table", "", "Base cache Bigtable table.")
+	port          = flag.Int("port", 12345, "Port on which to run the server.")
+	useALTS       = flag.Bool("use_alts", false, "Whether to use ALTS server authentication")
+	bigqueryOnly  = flag.Bool("bigquery_only", false, "The service only serves sparql query")
+	svobsMode     = flag.Bool("svobs_mode", false, "Use new storage that is built on top of StatVar instead of StatPop")
+	schemaPath    = flag.String("schema_path", "/translator/mapping", "The director that contains the schema mapping files")
```
pradh, line 48: Would we not have different mapping files based on svobsMode ?

pradh, line 48: nit: directory

cmd/main.go
```diff
@@ -37,25 +37,23 @@ import (
...
-	svobsMode        = flag.Bool("svobs_mode", false, "Use new storage that is built on top of StatVar instead of StatPop")
-	schemaPath       = flag.String("schema_path", "/translator/mapping", "The director that contains the schema mapping files")
+	mixerProject  = flag.String("mixer_project", "", "The cloud project to run the mixer instance.")
+	storeProject  = flag.String("store_project", "", "GCP project stores Bigtable and BigQuery.")
+	bqDataset     = flag.String("bq_dataset", "", "DataCommons BigQuery dataset.")
+	baseTableName = flag.String("base_table", "", "Base cache Bigtable table.")
+	port          = flag.Int("port", 12345, "Port on which to run the server.")
+	useALTS       = flag.Bool("use_alts", false, "Whether to use ALTS server authentication")
+	bigqueryOnly  = flag.Bool("bigquery_only", false, "The service only serves sparql query")
+	svobsMode     = flag.Bool("svobs_mode", false, "Use new storage that is built on top of StatVar instead of StatPop")
```
pradh, line 47: nit: StatVar/StatVarObs instead of StatPop/Obs

docs/developer_guide.md
```diff
@@ -96,14 +96,11 @@ Run the following code to start mixer gRPC server
 # cd into repo root directory
 
 go run cmd/main.go \
+    --mixer_project=datcom-mixer-staging \
+    --store_project=google.com:datcom-store-dev \
```
pradh, line 100: nit: TODO: to switch this to datcom-store ?

cmd/main.go
shifucun: That bucket only has the base cache version files: https://pantheon.corp.google.com/storage/browser/automation_control;tab=objects?forceOnBucketsSortingFiltering=false&project=google.com:datcom-store-dev&prefix=&forceOnObjectsSortingFiltering=false

docs/developer_guide.md
shifucun: Done

cmd/main.go
shifucun: Done

shifucun: This is a mapped directory in the pod. It's the same for all deployment: https://github.com/datacommonsorg/mixer/blob/master/deploy/base/deployment.yaml#L64-L66, which is specified by the config map https://github.com/datacommonsorg/mixer/blob/master/deploy/base/deployment.yaml#L46-L47. For example, in the current deployment: https://github.com/datacommonsorg/mixer/blob/master/deploy/mapping/kustomization.yaml#L18

cmd/main.go
shifucun: Done

shifucun (commented):
thanks for review!

cmd/main.go
pradh: Ah, I see!

pradh (approved):

PR #483, opened 2021-03-19
Fix bugs of golden diff notification

shifucun:
There are two issues:

* The base cache version in SVOBS mode should be updated
* when generate_golden is True, should only update golden files but not run tests


pradh (approved):

shifucun (commented):
thanks for review

PR #488, opened 2021-04-02
Add API to read StatVarGroup for a place

shifucun:
This API reads the stat var group cache and filter out stat var and stat var group that are not relevant to this place.

A stat var group will be kept if itself or it's children stat var group has valid children stat vars for this place


chejennifer (approved):

chejennifer:
thanks for adding!

internal/server/statvar_group.go
```diff
@@ -0,0 +1,118 @@
...
+		return nil, err
+	}
+	svgResp := &pb.StatVarGroups{}
+	err = protojson.Unmarshal(jsonRaw, svgResp)
+	if err != nil {
+		return nil, err
+	}
+
+	// Read place statvars
+	if len(svgResp.StatVarGroups) > 0 {
```
pradh, line 100: nit: maybe return if len(...) == 0 {} so bulk of the logic need not be nested :)

internal/server/statvar_group.go
```diff
@@ -0,0 +1,118 @@
...
+
+func filterSVG(svgResp *pb.StatVarGroups, placeSVs []string) *pb.StatVarGroups {
+	// Build set for all the SV.
+	svSet := map[string]struct{}{}
+	for _, sv := range placeSVs {
+		svSet[sv] = struct{}{}
+	}
+
+	// All the svg with valid sv for this place
+	svgSet := map[string]struct{}{}
```
pradh, line 37: curious, why not bool if it's just a set?

internal/server/statvar_group.go
```diff
@@ -0,0 +1,118 @@
...
+	// for this place.
+	for svgID, svgData := range svgResp.StatVarGroups {
+		filteredChildren := []string{}
+		for _, child := range svgData.ChildStatVars {
+			if _, ok := svSet[child]; ok {
+				filteredChildren = append(filteredChildren, child)
+			}
+		}
+		svgData.ChildStatVars = filteredChildren
+		if len(filteredChildren) > 0 {
```
pradh, line 49: There could be grandchildren and below without children? so should we check all child SVGs are excluded before excluding?

pradh: I see. Its handled in the next loop.

internal/server/statvar_group.go
```diff
@@ -0,0 +1,118 @@
...
+			svgSet[svgID] = struct{}{}
+		}
+	}
+
+	// Another iteration to remove SVG without svg children nor sv children
+	for svgID, svgData := range svgResp.StatVarGroups {
```
pradh, line 55: I'm wondering if this loop should happen from child to parent? Otherwise is there a chance we may keep empty nodes.

internal/server/statvar_group.go
```diff
@@ -0,0 +1,118 @@
...
+		filteredChildren := []*pb.StatVarGroupNode_Child{}
+		for _, c := range svgData.ChildStatVarGroups {
+			if _, ok := svgSet[c.Id]; ok {
+				filteredChildren = append(filteredChildren, c)
+			}
+		}
+		svgData.ChildStatVarGroups = filteredChildren
+		if len(svgData.ChildStatVars) == 0 && len(svgData.ChildStatVarGroups) == 0 {
+			delete(svgResp.StatVarGroups, svgID)
+		}
```
pradh, line 65: To confirm, is it safe in golang to iterate over a map and mutate it in the same loop?

internal/server/statvar_group.go
shifucun: The empty struct has no size: https://medium.com/@l.peppoloni/how-to-improve-your-go-code-with-empty-structs-3bd0c66bc531

internal/server/statvar_group.go
shifucun: Yep: https://stackoverflow.com/questions/23229975/is-it-safe-to-remove-selected-keys-from-map-within-a-range-loop

internal/server/statvar_group.go
shifucun: We have got all the valid SVG before this loop, so this is just to check the validity for the children SVG, seems safe.

internal/server/statvar_group.go
shifucun: Done

internal/server/statvar_group.go
shifucun: Now handle multiple hops of svg children with a recursive check of validity

internal/server/statvar_group.go
```diff
@@ -28,34 +27,60 @@ import (
...
 				filteredChildren = append(filteredChildren, child)
 			}
 		}
 		svgData.ChildStatVars = filteredChildren
-		if len(filteredChildren) > 0 {
-			svgSet[svgID] = struct{}{}
+	}
+
+	// Step 2: recursively check if a stat var group is valid. A stat var group
+	// is valid if it has any grand children with non-empty stat vars.
```
pradh, line 48: nit: has any descendent stat var group with non-empty stat vars.

internal/server/statvar_group.go
```diff
@@ -28,34 +27,60 @@ import (
...
+		// Recursively check child svg, if there is any valid svg child, then this
+		// is valid too
+		for _, svgChild := range svgChildren {
+			if checkValid(svgChild.Id) {
+				validSVG[svgID] = struct{}{}
+				return true
+			}
 		}
+		return false
+	}
```
pradh, line 74: nit: move inlined function out for readability?

internal/server/statvar_group.go
```diff
@@ -28,34 +27,60 @@ import (
...
+		// If there are non-empty sv chldren, then this svg is valid
+		if len(svChildren) > 0 {
+			validSVG[svgID] = struct{}{}
+			return true
+		}
+		// Recursively check child svg, if there is any valid svg child, then this
+		// is valid too
+		for _, svgChild := range svgChildren {
+			if checkValid(svgChild.Id) {
+				validSVG[svgID] = struct{}{}
```
pradh, line 69: is this redundant, given the addition in line 62?

internal/server/statvar_group.go
pradh: Thanks

pradh (approved):
Thanks for the changes!

internal/server/statvar_group.go
shifucun: For example in X->Y->Z(sv1, sv2) case, this is to mark Y as valid, line62 marks Z as valid.

internal/server/statvar_group.go
shifucun: Done

internal/server/statvar_group.go
shifucun: Done

internal/server/statvar_group.go
pradh: Got it.

pradh (approved):

PR #489, opened 2021-04-02
Send notifiy email if there is test failure

shifucun:
Instead of golden data diff, sometimes there are test failure from data change. Adding this to catch that


pradh (approved):
Good idea.

PR #491, opened 2021-04-07
Update stat var group API to return display and search names

shifucun:


pradh (approved):
Thanks for the change!

PR #496, opened 2021-04-08
Fix get_stat_all test

beets:
Save goldens for both the pop-obs and sv-obs based tests; enable the right comparisons during test.


pradh:
So, for my understanding, does this change fork all golden files for svobs vs. popobs?

pradh (approved):

beets:
Yea, it will temporarily - hopefully not for too long

PR #497, opened 2021-04-14
Handle non-place query for  GetStatVarGroup

shifucun:
GetStatVarGroup API can pass in an optional place dcid. It's likely the dcid is not about a place hence not associated with any stat vars. The API should check the existence of cache read to prevent nil reference.


pradh (approved):
Might be worth documenting the context/bug for this PR in the description.

shifucun:
Thanks! Added background for the fix.

PR #498, opened 2021-04-16
Ranking sources with latest or more data higher

shifucun:
Current ranking only considers the importance of sources. With this change, when two sources are both not in the score list, they are ranked based on the latest available date and the length of the data.


internal/server/ranking.go
```diff
@@ -69,6 +77,31 @@ func (a SeriesByRank) Less(i, j int) bool {
 	if scorei != scorej {
 		return scorei < scorej
 	}
+
+	datesi := []string{}
+	for date := range a[i].Val {
+		datesi = append(datesi, date)
+	}
+	sort.Strings(datesi)
+	latesti := datesi[len(datesi)-1]
```
pradh, line 86: If you're looking anyway, it might be more efficient to compute max while you do that, rather than do another Sort?

internal/server/ranking.go
```diff
@@ -112,6 +143,31 @@ func (a byRank) Less(i, j int) bool {
...
+
+	// Series with latest data is ranked higher
+	if latesti != latestj {
+		return latesti > latestj
+	}
+
+	// Series with more data is ranked higher
+	if len(datesi) != len(datesj) {
+		return len(datesi) > len(datesj)
+	}
```
pradh, line 169: if the maps here and above are of the same, maybe we can use a helper?

test/integration/golden_response/staging/get_landing_page_data/tha.json
```diff
@@ -16920,9 +16920,22 @@
       "Count_HousingUnit_RenterOccupied": {},
       "Count_Person": {
         "data": {
-          "2000": 6355144
+          "1919-04-01": 437294,
```
pradh, line 16923: wikidata source is bad to return as time-series :(

pradh (approved):

internal/server/ranking.go
```diff
@@ -12,9 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Sourcer ranking is based on the following criteria in order:
```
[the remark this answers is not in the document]
shifucun: Done

internal/server/ranking.go
shifucun: Done

internal/server/ranking.go
shifucun: The logic is repeated since one uses Go struct and the other using protobuf. I have a pending task to use protobuf for all the API and make them /v1/..

test/integration/golden_response/staging/get_landing_page_data/tha.json
shifucun: made it non-preferred

pradh (commented):
Thanks for the changes!

PR #499, opened 2021-04-19
API endpoint "/version" to get version information of mixer

shifucun:


beets (approved):
thanks for adding this! very useful

PR #501, opened 2021-04-19
Update mixer deploy scripts

shifucun:
- Move docker build script to reside in the same folder as Dockerfile.
- Make a new script to build and push binaries from local code

### DIFF
--- a/build/push_go_protoc_image.sh
+++ b/build/push_go_protoc_image.sh
@@ -15,6 +15,7 @@
 
 set -e
 
+# Get the current directory path and the root path of the repo
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 ROOT="$(dirname "$DIR")"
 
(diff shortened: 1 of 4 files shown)


build/push_go_protoc_image.sh
```diff
@@ -15,5 +15,10 @@
 
 set -e
 
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
pradh, line 18: A comment on what it does?

pradh (approved):
Thanks!

scripts/push_binary.sh
```diff
@@ -0,0 +1,27 @@
+  #!/bin/bash
```
beets, line 1: nit: spacing

beets (approved):
thanks for this!

build/push_go_protoc_image.sh
shifucun: Done

scripts/push_binary.sh
shifucun: Done

shifucun (commented):
thanks for review!

PR #502, opened 2021-04-20
Add self managed ssl certificate to applicable mixer deployment

shifucun:
For prod, staging and autopush deployment, self managed ssl certificate "multi-domain" is created in the load balancer.

This PR adds this certificate to the Ingress yaml file, so the managed and self created SSL would attach to the Ingress each time it is deployed.


beets (approved):
thanks for the update!

PR #508, opened 2021-04-28
[GetStatCollection] get the latest observation when date is not provided

shifucun:
This is to expand the API and make the date params optional. Clients (like scatter plot, choropleth) should be able to easily consume the API without first figuring out the available dates.

* This is only implemented for svobs mode considering the migration is complete.
* The result might contain a mixed date from mixed sources.


shifucun:
Sending this to unblock Jennifer. Need to add some unit test for the logic in follow up PR.

internal/server/stat_point.go
```diff
@@ -200,47 +200,113 @@ func (s *Server) GetStatCollection(
...
+						}
+					}
 				}
 			}
+			// Lastest date is always needed.
+			sv2dates[sv] = []string{latestDate.date}
+			// The date with most place count is needed too.
+			if mostCount.count > latestDate.count {
+				sv2dates[sv] = append(sv2dates[sv], mostCount.date)
+			}
```
pradh, line 241: Should this block also be inside the if data != nil check?

internal/server/stat_point.go
```diff
@@ -200,47 +200,113 @@ func (s *Server) GetStatCollection(
...
+						if count > mostCount.count || (count == mostCount.count && date > mostCount.date) {
+							mostCount = dateCount{date, count}
+						}
+						if date > latestDate.date || (date == latestDate.date && count >= latestDate.count) {
+							latestDate = dateCount{date, count}
+						}
+					}
 				}
 			}
+			// Lastest date is always needed.
```
pradh, line 236: nit: Latest

internal/server/stat_point.go
```diff
@@ -200,47 +200,113 @@ func (s *Server) GetStatCollection(
...
-			sort.Sort(SeriesByRank(cohorts))
-			result.Data[sv] = cohorts[0]
+
+	// Get a reverse sorted date list like: [2019, 2017, 2016-01]
+	dateSet := map[string]struct{}{}
+	for _, dates := range sv2dates {
+		for _, date := range dates {
+			dateSet[date] = struct{}{}
 		}
 	}
```
pradh, line 255: nit: could compile dataSet in the loop above

internal/server/stat_point.go
```diff
@@ -200,47 +200,113 @@ func (s *Server) GetStatCollection(
...
+		dateList = append(dateList, date)
+	}
+	sort.Sort(sort.Reverse(sort.StringSlice(dateList)))
+
+	// Query the cache from the most recent date. The order is important as the
+	// source with older date is only used when it has better place coverage.
+	// TODO(shifucun): Parallel the BT query.
+	for _, queryDate := range dateList {
+		rowList, keyTokens := buildStatCollectionKeySvObs(
+			parentPlace, childType, queryDate, statVars)
```
pradh, line 267: At this point, do we need to query all statVars? Is it enough to just query those statVars for which this date exists?

internal/server/stat_point.go
```diff
@@ -200,47 +200,113 @@ func (s *Server) GetStatCollection(
...
+				for _, cohort := range cohorts {
+					currSize := len(cohort.Val)
+					if currSize > maxSize {
+						maxSize = currSize
+						usedCohort = cohort
+					}
+				}
+				// Add the cohort to result
+				if result.Data[sv].Val == nil {
+					result.Data[sv] = usedCohort
```
pradh, line 295: Should we return the dates?  For example, either on a per data point basis or perhaps simpler to return a list of dates covered in the results?

internal/server/stat_point.go
```diff
@@ -200,47 +200,113 @@ func (s *Server) GetStatCollection(
...
+				} else {
+					// If the result is populated already, then a latest observation
+					// has been added. For this cohort (with an older date), only need
+					// to add the places that is not present yet.
+
+					// TODO(boxu): In this case, the result contains multiple sources,
```
beets, line 301: this will end up changing the shape of the response in a backwards incompatible way. we should check if there are other callers of this api and version it, or create a new endpoint?

internal/server/stat_point.go
chejennifer: Maybe per data point would be good so that we can clearly show what the data in each datapoint represents.

internal/server/stat_point.go
shifucun: good catch, "continue" earlier with data == nil

test/integration/golden_response/staging/get_stat_collection/US_State.json
```diff
@@ -0,0 +1,220 @@
...
+        "geoId/55": {
+          "date": "2019",
+          "value": 5822434
+        },
+        "geoId/56": {
+          "date": "2019",
+          "value": 578759
+        }
+      },
+      "metadata": [
```
shifucun, line 206: Notice here there are two imports involved.

internal/server/stat_point.go
shifucun: Updated the response to use date->value struct

internal/server/stat_point.go
shifucun: This API has not been used yet. I updated the response to include multiple source metadata now.

test/integration/golden_response/staging/get_stat_collection/US_State.json
```diff
@@ -0,0 +1,220 @@
+{
```
shifucun, line 1: @chejennifer @pradh @beets PTAL the new response format

chejennifer: This response format looks good to me

internal/server/stat_point.go
```diff
@@ -273,41 +287,35 @@ func (s *Server) GetStatCollection(
...
-				}
-				// Add the cohort to result
-				if result.Data[sv].Val == nil {
-					result.Data[sv] = usedCohort
-				} else {
-					// If the result is populated already, then a latest observation
-					// has been added. For this cohort (with an older date), only need
-					// to add the places that is not present yet.
+			cohorts := data.SourceCohorts
+			sort.Sort(CohortByRank(cohorts))
```
beets, line 291: nice!

test/integration/golden_response/staging/get_stat_collection/US_State.json
beets: should we include a pointer to the import per stat? so we'd know which source to attribute to for each datapoint? the use case would be for the tooltip to be specific in its attribution. one could argue that this works for the entire collection, and the other would be the result of /stat/value anyway. thoughts?

beets (commented):
this looks good!

test/integration/golden_response/staging/get_stat_collection/US_State.json
shifucun: Added the import name in each stat value. That can reference the full source information in the response by import name.

beets (approved):
thanks for making the change!

shifucun:
In response, change source info from list to map, with import name as the key. Import name is also stored in each stat, so this helps to access the full source info

chejennifer (approved):

internal/server/stat_point.go
```diff
@@ -273,41 +287,33 @@ func (s *Server) GetStatCollection(
...
-					}
-				}
-				// Add the cohort to result
-				if result.Data[sv].Val == nil {
-					result.Data[sv] = usedCohort
-				} else {
-					// If the result is populated already, then a latest observation
-					// has been added. For this cohort (with an older date), only need
-					// to add the places that is not present yet.
+			cohorts := data.SourceCohorts
```
pradh, line 290: nit: maybe additional sanity len(cohorts) check before doing cohorts[0]?

proto/stat.proto
```diff
@@ -40,6 +40,7 @@ message PointStat {
 message PlacePointStat {
   // Keyed by place DCID.
   map<string, PointStat> stat = 1;
+  map<string, StatMetadata> metadata = 2;
```
pradh, line 43: What is the key?

internal/server/stat_point.go
```diff
@@ -273,41 +287,33 @@ func (s *Server) GetStatCollection(
...
+				result.Data[sv].Stat[place] = &pb.PointStat{
+					Date:  queryDate,
+					Value: val,
+					Metadata: &pb.StatMetadata{
+						ImportName: cohort.ImportName,
+					},
+				}
+			}
+			if _, ok := sv2Meta[sv][cohort.ImportName]; !ok {
+				sv2Meta[sv][cohort.ImportName] = &pb.StatMetadata{
```
pradh, line 305: Hmm... I wonder if the import-name is unique... e.g, a single import-name for the same stat-var could have different measurement-methods or observation-periods (e.g, BLS has adjusted/unadjusted, P1M/P1Y)...

pradh (approved):
Looks much better!

internal/server/stat_point.go
shifucun: Done

proto/stat.proto
shifucun: Added comment, keyed by import name

internal/server/stat_point.go
shifucun: In the current scenario, keyed by import name is fine for the BLS case. Since only one date is picked up (same time and place coverage for BLSSeasonallyUnadjusted, BLSSeasonallyAdjusted, and then only the top ranked cohort is processed. 

Added comments and TODO to reflect the dicussion

PR #509, opened 2021-04-28
PopObs -> SvObs migration clean up

shifucun:
This PR removes all the PopObs model reference in implementation, test and configurations. This effectively removes all the "popobs" and "svobs" keywords throughout the code base.

* Removes the conditional check for svobs mode in all the API.
* Update tests to only test svobs mode.
* Remove all the svobs config in deployment script and yaml files.
* Remove GetPopulation and GetObservation API.

During the migration I realized the GetPlaceObs API (used by old scatter tool) was not migrated, hence completed in this PR. After this is merged, I will update the website code to make it work.


beets (approved):
thanks for the extensive cleanup!

chejennifer (approved):
Thanks for the cleanup!

pradh (approved):
w00000t!

Once again, we should be sure to carefully roll this out to prod next week!

PR #511, opened 2021-04-30
Migration msg for /bulk/place-obs

shifucun:


internal/server/populations.go
```diff
@@ -26,16 +26,41 @@ import (
...
+Please migrate your API to use the new model. Example request:
+
+curl --request POST \
+  --url https://api.datacommons.org/bulk/place-obs \
+  --header 'content-type: application/json' \
+  --data '{
+		"placeType": "State",
+		"statVar": "Count_Person",
+		"date": "2015",
+	}'
```
pradh, line 46: So we don't have API doc for it?

pradh (approved):
Nice.

internal/server/populations.go
shifucun: Nope, this is not a public API

shifucun:
Thanks!

PR #512, opened 2021-05-04
Revive GetStatSetWithinPlace

shifucun:
* Rename GetStatCollection to GetStatSetWithinPlace and also updated the API endpoint from /stat/collection to /stat/set/within-place. This is to align with the [design](https://docs.google.com/document/d/12rdNsaLsMj5LxUWS3dynC8fuYTd0sfxqHio0Hh8Joao/edit#heading=h.9ssx4uezxkkc).

* In the implementation, instead of fetching the ObsCollection cache, now fetches the child places first, then fetch the ObsSeries cache and look for stat value for the date (latest date). This actually runs fairly fast even for "US Cities".

* Consolidate the logic and output format for /stat/set and /stat/set/within-place.


test/integration/golden_response/staging/get_stat_set_within_place/CA_County.json
```diff
@@ -9,13 +9,41 @@
             "import_name": "CensusPEP"
           }
         },
+        "geoId/06003": {
```
shifucun, line 12: Note the added places. The new implementation guarantees full coverage of all the places

beets: nice!

test/integration/golden_response/staging/get_stat_set/2010.json
```diff
@@ -6,24 +6,26 @@
           "date": "2010",
           "value": 36760,
           "metadata": {
-            "importName": "WorldDevelopmentIndicators",
-            "provenanceUrl": "https://datacatalog.worldbank.org/dataset/world-development-indicators/",
-            "observationPeriod": "P1Y",
-            "unit": "InternationalDollar"
+            "importName": "WorldDevelopmentIndicators"
```
beets, line 9: this change to a public api might be problematic. can we maintain backwards compatibility for these?

beets (commented):
nice, clean implementation!

test/integration/golden_response/staging/get_stat_set/2010.json
shifucun: These two APIs are not used since implemented (both internally or externally)  image

Should we use them for a while, gather feedback for interface and then publicly release them (with /v1/ prefix to the REST path as well). I feel this is an alpha version with a high chance of more changes..

shifucun: Also hope to get all the APIs in [design](https://docs.google.com/document/d/12rdNsaLsMj5LxUWS3dynC8fuYTd0sfxqHio0Hh8Joao/edit#heading=h.9ssx4uezxkkc) implemented and used. When we feel confident about the interface, can release all of them as /v1/stat/*** at the same time

beets: that sounds like plan. nice to be able to make the change now while we can.

beets (approved):

test/integration/golden_response/staging/get_stat_set_within_place/CA_County.json
pradh: Nice. So the old implementation had fewer places because of date limitation?

chejennifer (approved):
Looks nice!

internal/server/key.go
```diff
@@ -55,7 +55,7 @@ func buildStatsKey(
 	return rowList, keyToToken
 }
 
-func buildStatCollectionKey(parentPlace, childType, date string, statVars []string) (
+func buildStatSetWithinPlaceKey(parentPlace, childType, date string, statVars []string) (
```
pradh, line 58: Is there any caller with date set?  If not, should we then stop building this ObsCollection cache?

pradh (approved):

test/integration/golden_response/staging/get_stat_set_within_place/CA_County.json
shifucun: For this case, it's a bug from old implementation. The data is correctly picked but the source cohorts was picked by source ranking. Instead it should have picked the source with the most places.

In the file below (get_stat_set_within_place/US_State.json), the missing data is due to CensusPEP not having geoId/11 data in 2019. It should have use both sources.

internal/server/key.go
shifucun: "d/f/" cache is not used now (both w/o date). We can disable the cache build for now (maybe keep the code?).

shifucun:
Thanks all for the review!

PR #513, opened 2021-05-05
Support both GET and POST for most APIs

shifucun:


chejennifer (approved):
Thanks!

beets (commented):
yay! this is nice! we should update our docs too.

shifucun:
> yay! this is nice! we should update our docs too.

That's right, need a bunch of docs update regarding the recent change. I will file a BUG.

PR #514, opened 2021-05-06
Remove redundent cache merge in GetPropertyValue API

shifucun:
Currently readPropertyValues() function only uses the base cache [1], so it's unnecessary to perform cache merging. This currently incurs 2 Bigtable read, hence a much longer time for request that performs O(10K) row read.

[1] https://github.com/datacommonsorg/mixer/blob/542c6aebb26958a831019989a817560f67d48d8e/internal/server/property_value.go

### DIFF
--- a/internal/server/property_value.go
+++ b/internal/server/property_value.go
@@ -105,7 +105,9 @@ func getPropertyValuesHelper(
 	arcOut bool,
 ) (map[string][]*Node, error) {
 	rowList := buildPropertyValuesKey(dcids, prop, arcOut)
-	// Add base cache data
+	// Only use base cache data for property value.
+	// TODO(shifucun): perform a systematic check on current cache data and see
+	// if this is still true.
 	return readPropertyValues(ctx, store, rowList)
 }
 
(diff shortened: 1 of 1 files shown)


internal/server/property_value.go
```diff
@@ -107,36 +106,7 @@ func getPropertyValuesHelper(
...
-			for _, n := range branchNodes {
-				if itemKeys.Contains(n.Dcid + n.Value) {
-					continue
-				}
-				nodeMap[dcid] = append(nodeMap[dcid], n)
-			}
-		}
-	}
-	return nodeMap, nil
+	return readPropertyValues(ctx, store, rowList)
```
pradh, line 109: Add comments for why only base cache?

pradh (approved):
Nice find!

internal/server/property_value.go
shifucun: Done. This is based on past inspection of the cache. Need to do this for current data though. But since current branch cache is targeted on new stats (without addition of schema etc), so this assumption should still holds.

shifucun (commented):
thanks for quick review!

PR #515, opened 2021-05-06
Do not set GKE pod cpu and memory limit

shifucun:
When debugging the mixer 5xx errors for long processing API with large amount data (stats of cities in US), I found it's due to the cpu and memory limit. When looking at the GKE monitoring, the peak cpu and memory are within the limit, however the behavior is not predictable. Client can get various 5xx errors, which may depending how the socket is terminated (mixer processing termination, restart or esp restart, etc.).

Turning off the limit has the following effect based on the [doc1](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/), [doc2](https://kubernetes.io/docs/concepts/policy/limit-range/) : Each Pod/Container can use the cpu/memory up to that of the cluster node, and when exceed, will result in OOM of the node, hence impacting other Pod.

Considering our use pattern that majority of requests are with small payload while some requests need really large memory and cp. Turning off the resource seems more beneficial.

I have tested this by manually deploying to autopush mixer. GetStatSetWithinPlace() of Cities in US for two statvars return within 10 seconds. Before it takes 30+ seconds and most time ends up with 5xx error.

Note this should also help the previous Bio Sparql query issue.

### DIFF
--- a/deploy/base/deployment.yaml
+++ b/deploy/base/deployment.yaml
@@ -48,10 +48,6 @@ spec:
         - name: mixer
           image: gcr.io/datcom-ci/datacommons-mixer:latest
           imagePullPolicy: Always
-          resources:
-            limits:
-              memory: "3G"
-              cpu: "500m"
           args:
             - --mixer_project=$(MIXER_PROJECT)
             - --store_project=$(STORE_PROJECT)
@@ -113,10 +109,6 @@ spec:
                 configMapKeyRef:
                   name: mixer-configmap
                   key: serviceName
-          resources:
-            limits:
-              memory: "0.5G"
-              cpu: "100m"
           readinessProbe:
             httpGet:
               path: /healthz
--- a/deploy/overlays/dev/patch_deployment.yaml
+++ b/deploy/overlays/dev/patch_deployment.yaml
@@ -18,40 +18,37 @@ metadata:
   name: mixer-grpc
 spec:
   selector:
-      matchLabels:
-        app: mixer-grpc
+    matchLabels:
+      app: mixer-grpc
   template:
     spec:
       volumes:
         - name: mixer-grpc-json
           emptyDir: {}
       containers:
-      - name: mixer
-        image: datacommons/mixer
-        imagePullPolicy: Never
-        resources:
-          limits:
-            memory: "1G"
-      - name: esp
-        volumeMounts:
-          - mountPath: /esp
-            name: mixer-grpc-json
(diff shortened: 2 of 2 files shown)


deploy/base/deployment.yaml
```diff
@@ -113,10 +109,6 @@ spec:
                 configMapKeyRef:
                   name: mixer-configmap
                   key: serviceName
-          resources:
-            limits:
-              memory: "0.5G"
-              cpu: "100m"
```
pradh, line 119: How can we monitor the impact of this change?  Meaning, how often does this happen and how many other pods are affected in collateral damage.

pradh (approved):
Maybe I missed it, but in https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/, not finding the part that describes behavior when limit is not set?

chejennifer (approved):

shifucun:
> Maybe I missed it, but in https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/, not finding the part that describes behavior when limit is not set?

Missed to copy some other reads, this one talks about using all resources when there is no limit: https://kubernetes.io/docs/concepts/policy/limit-range/

deploy/base/deployment.yaml
shifucun: Need to setup monitoring on node/pod level mem/cpu usage and pod restart.

beets:
Should we set higher limits, to avoid possible resource contention issues?

>

shifucun:
> Should we set higher limits, to avoid possible resource contention issues?
> […](#)

The number of pods is determined by the total cpu/mem and the limit per pod, so higher limit will result in fewer pods..

shifucun:
If we see resource contention issue when there is no limit set, then it will definitely result in a server error for whatever limit we set. The latter happens a lot as we see now while the former would be very rare, ie, large number of heavy request concurrently, and in this case, all of them will fail anyway..

beets (approved):
gotcha. sounds good, we can give this a go on staging and see how it behaves there with some load tests? and/or by standing up a second pod in prod.

btw, could you hold off on merging this in until after this week's push? thanks!

shifucun:
> gotcha. sounds good, we can give this a go on staging and see how it behaves there with some load tests? and/or by standing up a second pod in prod.
> 
> btw, could you hold off on merging this in until after this week's push? thanks!

Alright, maybe we can merge this so testing it in autopush, but the release can cut before it?

PR #518, opened 2021-05-06
Mixer upgrade and clean up post migration

shifucun:
* Upgrade Golang to 1.16 and update libraries
* Move golden response files out of "staging" since there is no "prod" version now.
* Remove references to "datcom-store-dev"
* Remove cache key for PopObs related cache


beets (approved):
thx for the cleanup!

internal/util/util.go
```diff
@@ -36,26 +36,16 @@ const (
 	BtPlaceStatsVarPrefix = "d/0/"
 	// BtStatVarGroup is the key for statvar group cache.
 	BtStatVarGroup = "d/1"
-	// BtPopObsPrefix for internal pop obs cache.
-	BtPopObsPrefix = "d/2/"
```
shifucun, line 40: @pradh We could also remove these cache from prophet

pradh: d/2/ is needed for pop-obs validation pipeline during import, until we switch import flows

so we're not using this obsid->{observation-about, variable-measured} cache?

shifucun: Not any more, observation properties are fetched using BQ

PR #520, opened 2021-05-11
Add back GKE resource limit to future investigate ESP issue

shifucun:
Seeing pods getting evicted in autopush mixer due to memory issue. I saw several large peaks of requests to the auotpush mixer in the past days, which is a bit mysterious, as it should only take the web-driver tests. 

I investigated the cpu and memory usage a little more with logging locally. With resources limit set, mixer actually was able to complete a large request under 15s, but ESP does not return the response back. 

Check in logging here and deploy to GKE under real load for more investigation.

Compare with past version, also added "requests" field to ensure the cpu and mem allocation.

image

### DIFF
--- a/internal/server/stat_point.go
+++ b/internal/server/stat_point.go
@@ -16,7 +16,9 @@ package server
 
 import (
 	"context"
+	"log"
 	"strings"
+	"time"
 
 	pb "github.com/datacommonsorg/mixer/internal/proto"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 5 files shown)


internal/server/stat_point.go
```diff
@@ -130,6 +135,13 @@ func (s *Server) GetStatSetWithinPlace(
 	childType := in.GetChildType()
 	date := in.GetDate()
 
+	log.Printf(
```
beets, line 138: would be nice to get this type of logging for all requests

beets (approved):
thanks for the investigation

internal/server/stat_point.go
pradh: +1

shifucun: Sg. I will do a PR for all requests

PR #526, opened 2021-05-18
cache data in mixer for stat var hierarchy 

chejennifer:
- on mixer start up, get and cache the raw svg mapping, the mapping of each sv/svg to its parent svg and the search index


internal/server/server.go
```diff
@@ -30,11 +30,19 @@ import (
...
 	"cloud.google.com/go/storage"
 	"github.com/datacommonsorg/mixer/internal/base"
+	pb "github.com/datacommonsorg/mixer/internal/proto"
 	"github.com/datacommonsorg/mixer/internal/store"
 	"github.com/datacommonsorg/mixer/internal/translator"
 	"github.com/datacommonsorg/mixer/internal/util"
 )
 
+// Cache holds cached data for the mixer server.
+type Cache struct {
```
shifucun, line 40: comment on what each cache is about?

internal/server/server.go
```diff
@@ -30,11 +30,19 @@ import (
...
+	ParentSvgMap   map[string]string
+	RawSvg         map[string]*pb.StatVarGroupNode
```
shifucun, line 42: Should use this in https://github.com/datacommonsorg/mixer/blob/master/internal/server/statvar_group.go#L118-L134!

internal/server/statvar_hierarchy_util.go
```diff
@@ -0,0 +1,107 @@
...
+	return svgResp.StatVarGroups, nil
+}
+
+// GetParentSvgMap gets the mapping of svg/sv id to the parent svg for that svg/sv.
+func GetParentSvgMap(rawSvg map[string]*pb.StatVarGroupNode) map[string]string {
+	parentSvgMap := map[string]string{}
+	for svgID, svgData := range rawSvg {
+		for _, childSvg := range svgData.ChildStatVarGroups {
+			if _, ok := parentSvgMap[childSvg.Id]; !ok {
+				parentSvgMap[childSvg.Id] = svgID
```
shifucun, line 64: So now only one parent is stored. What is the implication of the parent selection? It came to mind to store all the parents, but looks like for now, only a random random parent would be used?

internal/server/statvar_hierarchy_util.go
```diff
@@ -0,0 +1,107 @@
...
+
+func updateSearchIndex(tokenString string, index map[string]map[string]RankingInfo, nodeID string) {
+	processedTokenString := strings.ToLower(tokenString)
+	processedTokenString = strings.ReplaceAll(processedTokenString, ",", " ")
+	tokenList := strings.Fields(processedTokenString)
+	approxNumPv := len(strings.Split(nodeID, "_"))
+	if approxNumPv == 1 {
+		// when approxNumPv is 1, most likely a non human curated PV and we want them
+		// ranked lower (less approxNumPv, higher ranking)
+		approxNumPv = 30
```
shifucun, line 84: nit: make this a const

internal/server/statvar_hierarchy_util.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/datacommonsorg/mixer/internal/util"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+// RankingInfo holds the ranking information for each stat var hierarchy search
+// result.
+type RankingInfo struct {
+	ApproxNumPv int
```
shifucun, line 32: Add some comment on the struct fields

internal/server/statvar_hierarchy_util.go
```diff
@@ -0,0 +1,107 @@
...
+		if _, ok := index[token]; !ok {
+			index[token] = map[string]RankingInfo{}
+		}
+		index[token][nodeID] = rankingInfo
+	}
+}
+
+// GetSearchIndex gets the search index for the stat var hierarchy.
+func GetSearchIndex(rawSvg map[string]*pb.StatVarGroupNode) map[string]map[string]RankingInfo {
+	searchIndex := map[string]map[string]RankingInfo{}
```
shifucun, line 97: comment on what the key is for each map

internal/server/statvar_hierarchy_util_test.go
```diff
@@ -0,0 +1,174 @@
...
+						},
+						{
+							Id:          "sv3",
+							SearchName:  "token4,",
+							DisplayName: "sv4",
+						},
+					},
+				},
+			},
+			map[string]map[string]RankingInfo{
```
shifucun, line 117: I wonder given a search token, how is the RankingInfo used?

internal/server/server.go
shifucun: How about naming with "ParentSvg", "SvgInfo" and "SvgSearchIndex" to make them consistent.

shifucun (commented):
looking good! Very clear implementation, just some minor comments

internal/server/statvar_hierarchy_util_test.go
chejennifer: when given a search token, should return the results after having been sorted by rankinginfo .. can mixer return information in a specific order?

internal/server/statvar_hierarchy_util.go
chejennifer: We were always only choosing a single path so I thought we would only need one parent stored? Would it be better to store all the parents?

internal/server/server.go
chejennifer: I was thinking of doing this in a followup pr

shifucun: Yep, but you may try locally see if the cache works (it should)

internal/server/statvar_hierarchy_util.go
shifucun: I think it's better, store a sorted parent and client pick the first one for now.

internal/server/statvar_hierarchy_util_test.go
shifucun: Why not make the struct as map[string][]NewRankingInfo, where NewRankingInfo contains the sv(g) id and RankingInfo?

internal/server/server.go
chejennifer: done

internal/server/statvar_hierarchy_util.go
chejennifer: done

internal/server/statvar_hierarchy_util.go
chejennifer: done

internal/server/statvar_hierarchy_util.go
chejennifer: done

internal/server/statvar_hierarchy_util.go
chejennifer: done

internal/server/statvar_hierarchy_util_test.go
chejennifer: when filtering for tokens, we will need to do an intersection on the results of every token. Map makes it easier to do the intersection

shifucun (approved):
thanks for update, looking good!

PR #529, opened 2021-05-24
Delete branch cache subscriber on mixer server shutdown

shifucun:
Tested locally, will verify after pushing to autopush

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -135,7 +135,7 @@ func main() {
 			if err != nil {
 				log.Fatalf("Failed to delete subscriber: %v", err)
 			}
-			log.Printf("Delete subscriber: %v", sub)
+			log.Printf("Deleted subscriber: %v", sub)
 			os.Exit(1)
 		}()
 	}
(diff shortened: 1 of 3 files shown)


cmd/main.go
```diff
@@ -119,11 +122,22 @@ func main() {
...
 	// Subscribe to cache update
 	if !*bigqueryOnly {
-		err = s.SubscribeBranchCacheUpdate(
+		sub, err := s.SubscribeBranchCacheUpdate(
 			ctx, *storeProject, branchCacheVersionBucket, subscriberPrefix, pubsubTopic)
 		if err != nil {
 			log.Fatalf("Failed to subscribe to branch cache update: %v", err)
 		}
+		c := make(chan os.Signal, 1)
+		signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
```
pradh, line 131: Some comments on what you're doing here?

pradh (approved):
Thanks for the fix!

cmd/main.go
shifucun: Done

PR #531, opened 2021-06-01
Flatten the response for /place/stat-vars/union

shifucun:
The current response has an unnecessary nested layer, flatten it and created a new endpoints with prefix /v1.


beets (approved):
thanks for the fix bo!

@chejennifer fyi

shifucun:
Thanks!

PR #532, opened 2021-06-02
Update place landing page API

shifucun:
This is a breaking change, but as the API is only used by website, updating in place is better than forking a new API. Major changes are:

* Always fetch data for additional stat vars, so client is responsible for making sure missing stat vars in cache are provided in the request. This is only needed for local testing before a new cache generation is completed with new chart config.
* Get rid of "payload" and use protobuf instead of go struct throughout. This leverages some new APIs and helper functions so the code is simpler.


internal/server/landing_page.go
```diff
@@ -247,70 +248,40 @@ func fetchBtData(
...
-					return err
-				}
-				dataChan <- svData{sv: statVar, data: tmp}
-				return nil
-			})
-		}
-		err := errs.Wait()
+	// Fetch addtioinal stats as requested.
+	if len(statVars) > 0 {
+		resp, err := s.GetStatSetSeries(ctx, &pb.GetStatSetSeriesRequest{
```
pradh, line 263: For my understanding, does the website then change to stop setting the extra stat-vars in prod?

https://github.com/datacommonsorg/website/blob/175e9c62f64e62b89f8a1bcc2ed4bfb599c7c6a0/server/routes/api/landing_page.py#L382

internal/server/landing_page.go
```diff
@@ -247,70 +248,40 @@ func fetchBtData(
...
+		for place, seriesMap := range resp.Data {
+			for statVar, series := range seriesMap.Data {
+				if result[place] == nil {
+					result[place] = &pb.StatVarSeries{}
 				}
+				result[place].Data[statVar] = series
+			}
+		}
+	}
+	// Delete the empty entries. This will be moved to cache generation.
```
pradh, line 280: To confirm, this won't be necessary once we drop empty entries in cache?

proto/mixer.proto
```diff
@@ -197,19 +197,39 @@ message GetRelatedLocationsResponse {
...
 // Request to get all stat data in landing page for a place.
 message GetLandingPageDataRequest {
   // The dcid of the place.
   string place = 1;
-  // A list of all stat vars used in the landing page.
+  // A list of additional stat vars need to be fetched in addition to the
+  // data in cache. This is to be used in local development, where new chart is
+  // added to chart config but data is not added to cache (delay from cache
+  // build).
   repeated string stat_vars = 2;
```
pradh, line 218: Call it extra_stat_vars or test_stat_vars ?

Also, should we reserve 2 and use tag 3?

proto/stat.proto
```diff
@@ -98,10 +98,25 @@ message ChartStore {
...
 message PlaceStat {
+  // Key is stat var dcid.
   map<string, ObsTimeSeries> stat_var_data = 3;
 }
 
+// This is effectively the same proto as PlaceStat with different field name.
+// This is used to match the landing page cache data.
+message StatVarObsSeries {
+  // Key is stat var dcid.
+  map<string, ObsTimeSeries> data = 3;
```
pradh, line 112: does the tag num need to be 3? (and below)

proto/stat.proto
```diff
@@ -98,10 +98,25 @@ message ChartStore {
...
+}
+
+message StatVarSeries {
```
pradh, line 115: The protos in this file is very similar looking and confusing :).

For example, why create a new message identical to PlaceStat?  How is this different from SeriesMap?  Some comments would really help.

pradh (approved):

proto/mixer.proto
beets: or new_stat_vars?

internal/server/landing_page.go
beets: along those lines, since this is a website specific api change that isn't backwards compatible, should this change be made in the website repo, along with corresponding flask changes? we might not be able to autopush otherwise.

beets (commented):
thanks for doing this update!

test/integration/get_landing_page_data_test.go
```diff
@@ -86,27 +86,21 @@ func TestGetLandingPageData(t *testing.T) {
...
+		var expected pb.GetLandingPageDataResponse
 		file, _ := ioutil.ReadFile(goldenFile)
-		err = json.Unmarshal(file, &expected)
+		err = protojson.Unmarshal(file, &expected)
 		if err != nil {
 			t.Errorf("Can not Unmarshal golden file %s: %v", c.goldenFile, err)
 			continue
 		}
-		if diff := cmp.Diff(&result, &expected); diff != "" {
+		if diff := cmp.Diff(resp, &expected, protocmp.Transform()); diff != "" {
```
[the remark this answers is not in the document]
shifucun: The diff is viewable now but not optimal, especially for later files.. wdyt?

shifucun (commented):
Thanks for the review!

shifucun:
PTAL. The sharded golden files diff is also hard to look at. I am planning to revert it in a later PR with some sampling strategy.

beets (approved):
sorry for the late review!! thanks for the updates.

proto/mixer.proto
```diff
@@ -197,19 +197,41 @@ message GetRelatedLocationsResponse {
...
+
+  reserved 2;
 }
 
-// Response to get landing page info for places.
+// Response to get landing page info for a place.
 message GetLandingPageDataResponse {
-  string payload = 3;
+  // Keyed by place dcid.
+  map<string, StatVarSeries> data = 1;
```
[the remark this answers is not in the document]
shifucun: Updated

PR #533, opened 2021-06-02
Try shard the golden json file for better diff comparison

shifucun:
When a file is too big, Github refuses to render the diff hence making the golden file diff review impossible.

Try this approach and see if it helps. One possible caveat is when the diff drift to other files, but can try..


test/integration/get_landing_page_data_test.go
```diff
@@ -95,13 +94,17 @@ func TestGetLandingPageData(t *testing.T) {
 
 		goldenFile := path.Join(goldenPath, c.goldenFile)
 		if generateGolden {
-			updateGolden(result, goldenFile)
+			updateGolden(result, goldenFile, true)
```
pradh, line 97: nit: /* shard= */ true

(useful to know arg in the callsite)

test/integration/setup.go
```diff
@@ -166,15 +170,22 @@ func createBranchTable(ctx context.Context) (*bigtable.Table, error) {
 	return server.SetupBigtable(ctx, data)
 }
 
-func updateGolden(v interface{}, fname string) {
+func updateGolden(v interface{}, fname string, shard ...bool) {
 	jsonByte, _ := json.MarshalIndent(v, "", "  ")
-	err := ioutil.WriteFile(fname, jsonByte, 0644)
+	var err error
+	if len(shard) == 1 && shard[0] {
```
pradh, line 176: So, is this the way to do default arg in golang?

test/integration/setup.go
```diff
@@ -189,8 +200,65 @@ func updateProtoGolden(resp protoreflect.ProtoMessage, fname string) {
...
+	err := filepath.WalkDir(path, func(s string, d fs.DirEntry, e error) error {
+		if e != nil {
+			return e
+		}
+		if strings.Contains(s, fname) {
+			result = append(result, s)
+		}
+		return nil
+	})
+	return result, err
```
pradh, line 247: Is result guaranteed to be sorted on name?  Document assumption?

pradh (approved):
Nice improvement, thanks!

test/integration/get_landing_page_data_test.go
shifucun: Done

test/integration/setup.go
shifucun: Yea, it's more of optional args and used as default arg, as there is no native default arg

test/integration/setup.go
shifucun: Yea, the output is sorted, added comment

shifucun (commented):
Thanks for review!

PR #534, opened 2021-06-08
Support sampling golden test json

shifucun:
The integration test golden json files have been growing. For some of the large files, Github refuses to show diff, which makes them useless.

Previous attempt to shard the json files do not help much. As a single line diff can change all the shard files and creates large diff set. This is reverted in this PR.

Here adds a way to sample the json response. Now it supports max sample count and excluding keys at certain level. It gives reasonable control of the output.


internal/util/util.go
```diff
@@ -291,3 +305,73 @@ func MergeDedupe(s1 []string, s2 []string) []string {
...
+		if fd.IsList() {
+			// Sampmle list.
+			oldList := value.List()
+			inc := int(1 / strat.Ratio)
+			var newList protoreflect.List
+			for i := 0; i < oldList.Len(); i += inc {
+				newList.Append(oldList.Get(i))
+			}
+			pr.Set(fd, protoreflect.ValueOfList(newList))
+		} else if fd.IsMap() {
```
beets, line 340: i wonder if we should skip sampling for maps. it'll be hard to keep the sample stable as the keys change.

beets (commented):
this is neat! i wonder if we should apply it only to the timeseries so that the diffs are stable though?

shifucun (commented):
Thanks for the review. Sampling the time series is a really good idea. The sampled json now looks much better now. Also the strategy struct is stretched for the deeply nested fields..

internal/util/util.go
```diff
@@ -87,6 +90,17 @@ type TypePair struct {
 	Parent string
 }
 
+// SamplingStrategy represents the config to use for sampling golden response
+// Json.
+type SamplingStrategy struct {
+	// Sampling ratio for list and map.
+	Ratio float32
+	// Sampling strategy for the child fields.
+	Children map[string]*SamplingStrategy
```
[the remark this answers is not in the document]
shifucun: Yep, and now it is 4 levels deep in landing page cache use case..

internal/util/util.go
shifucun: Also the time series are stored as map, so map sampling is needed.

test/integration/golden_response/get_landing_page_data/asm.sample.json
```diff
@@ -1,5 +1,3099 @@
...
+          "metadata": {
+            "importName": "WorldDevelopmentIndicators",
+            "provenanceUrl": "https://datacatalog.worldbank.org/dataset/world-development-indicators/",
+            "observationPeriod": "P1Y",
+            "scalingFactor": "100.0000000000"
+          }
+        },
+        "Amount_EconomicActivity_GrossDomesticProduction_Nominal_PerCapita": {
+          "val": {
+            "2002": 8700.06770480704,
```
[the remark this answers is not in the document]
shifucun: Updated based on the new strategy

internal/util/util.go
```diff
@@ -93,11 +94,11 @@ type TypePair struct {
 // SamplingStrategy represents the config to use for sampling golden response
 // Json.
 type SamplingStrategy struct {
-	// Sampling ratio for list and map.
-	Ratio float32
+	// Maximum number of samples. -1 means sample all.
```
beets, line 97: can you describe here how the samples are generated based on this value? i think there's still value in picking up new data (e.g. for covid, if we sampled every n). with max samples, we could quickly exhaust it and not see new data, if all samples are kept from the start of the list / map.

internal/util/util.go
```diff
@@ -329,19 +330,33 @@ func Sample(m proto.Message, strategy *SamplingStrategy) proto.Message {
...
+			maxSample := strat.MaxSample
+			if strat.MaxSample == -1 || strat.MaxSample > length {
+				maxSample = length
+			}
+			inc := 1
+			if length > maxSample {
+				inc = int(math.Ceil(float64(length) / float64(maxSample)))
+			}
+			// Get the latest data first
+			for i := 0; i < maxSample; i++ {
```
beets, line 347: hm, the tracked keys won't stay stable this way, so the diffs will be harder to inspect even though the files are smaller (e.g. we won't know if values change, as data gets added which happens for covid).

beets (commented):
thanks for making the changes. the files are nice and much smaller now. i still worry about the stability of the files as data changes, mostly for frequently updated data sources.

internal/util/util.go
shifucun: The sampling is uniform and start from the end. For example, sample from [1,2,3,4,5,6,7,8,9,10] with MaxSample = 4, would result in [1,4,7,10]

internal/util/util.go
shifucun: Yea, if we sample the time series, the stability would indeed be an issue. It's possible the covid addition would cause entire diff of the series.

Or we could go back to the old approach that only samples the place but full data for that place. It's much more stable.

shifucun: We can probably try the current approach and adjust the SampleStrategy to place only if that's better?

beets: Chatted offline, and we'll try place sampling only and see how that works out.
Thanks Bo, and sorry for the back and forth.

shifucun: No worry. Switched back to sample the places only

internal/util/util.go
beets: thanks, could you add to the comment?

beets (approved):
thanks for the revert - let's try this out.

just one small nit comment

internal/util/util.go

pradh (commented):
Very cool change!  Thanks for making it.

PR #538, opened 2021-06-15
Add a new API to fetch stat var group information

shifucun:
Given a stat var group id, this API fetches its child stat var, child stat var groups and parent stat var groups.

This is meant to be used for stat var hierarchy widget that supports incremental data fetch.


internal/server/statvar_group.go
```diff
@@ -138,3 +141,104 @@ func (s *Server) GetStatVarGroup(
...
+			return nil, err
+		}
+		for _, c := range result.ChildStatVars {
+			c.DisplayName = statVarNames[c.Id]
+		}
+	}
+	return result, nil
+}
+
+func fetchStatVarNames(ctx context.Context, s *Server, statVars []string) (
```
pradh, line 210: Should we compute the names and put it in the "name" property of StatVar?

internal/server/statvar_group.go
```diff
@@ -138,3 +141,104 @@ func (s *Server) GetStatVarGroup(
...
+		pv := map[string]string{}
+		cpv := []string{}
+		for _, t := range data.Triples {
+			if t.Predicate == "measuredProperty" {
+				name = t.ObjectName + " of"
+			} else if t.Predicate == "constraintProperties" {
+				cpv = append(cpv, t.ObjectID)
+			} else {
+				pv[t.Predicate] = t.ObjectID
+			}
```
pradh, line 230: There's more special-casing in current logic http://google3/datacommons/prophet/flume_generator/stat_var_group_util.cc?l=320&rcl=377621180. Including that this doesn't ignore dependent PVs.

pradh (approved):

internal/server/statvar_group.go
shifucun: That would be nice!

internal/server/statvar_group.go
shifucun: I will leave it for now and use the name property once that's added in Flume

PR #540, opened 2021-06-16
Add an API to get stat var path in the hierarchhy

shifucun:


proto/mixer.proto
```diff
@@ -526,6 +526,16 @@ message RelatedPlacesInfo {
   reserved 4, 5, 6;
 }
 
+message GetStatVarPathRequest {
+  // DCID of the stat var (group)
```
pradh, line 530: Is it SV or SVG? clarify...

proto/mixer.proto
```diff
@@ -526,6 +526,16 @@ message RelatedPlacesInfo {
...
+  string id = 1;
+}
+
+message GetStatVarPathResponse {
+  // A list of DCIDs from the queried stat var to the root of stat var hierarchy.
```
pradh, line 535: Clarify perhaps that it is a single path, out of possibly multiple, to the root. And clarify the order?

proto/mixer.proto
```diff
@@ -526,6 +526,16 @@ message RelatedPlacesInfo {
   reserved 4, 5, 6;
 }
 
+message GetStatVarPathRequest {
```
pradh, line 529: This is actually a path of SVGs, so should it be GetStatVarGroupPathRequest ?

proto/mixer.proto
pradh, line 535: StatVarGroup DCIDs?

pradh (approved):

proto/mixer.proto
shifucun: Would like to de-emphasize stat var group in API naming and URL as it's more of a way to represent a set of stat vars. From user perspective, they are interested in stat var.

proto/mixer.proto

proto/mixer.proto
shifucun: It could be both, updated

shifucun (commented):
Thanks for review!

PR #541, opened 2021-06-16
Flatten stat var hierarchy index

shifucun:
Create a struct SearchIndex to hold two mappings used for stat var hierarchy search index.

Also separate out token to stat var / stat var group mapping, so Search can easily differentiate between the two types.


internal/server/statvar_hierarchy_util.go
```diff
@@ -86,34 +92,42 @@ func GetParentSvgMap(rawSvg map[string]*pb.StatVarGroupNode) map[string][]string
...
 	processedTokenString = strings.ReplaceAll(processedTokenString, ",", " ")
 	tokenList := strings.Fields(processedTokenString)
 	approxNumPv := len(strings.Split(nodeID, "_"))
 	if approxNumPv == 1 {
 		// when approxNumPv is 1, most likely a non human curated PV
-		approxNumPv = NonHumanCuratedNumPv
+		approxNumPv = nonHumanCuratedNumPv
 	}
-	rankingInfo := RankingInfo{approxNumPv, tokenString}
+	// Ranking info is only dependent on a stat var (group).
```
shifucun, line 106: @chejennifer The refactor is based on this observation. Let me if this makes sense.

chejennifer: yes, makes sense

chejennifer (approved):

PR #543, opened 2021-06-16
Add an API to search for stat var and stat var group

shifucun:


internal/server/statvar_search.go
```diff
@@ -0,0 +1,85 @@
...
+	query := in.GetQuery()
+	result := &pb.SearchStatVarResponse{
+		StatVars:      []string{},
+		StatVarGroups: []string{},
+	}
+	if query == "" {
+		return result, nil
+	}
+	tokens := strings.Split(
+		strings.Replace(strings.ToLower(query), ",", " ", -1), " ")
```
chejennifer, line 38: We don't want to only split by a single space because the token string could be separated by more than a single space or have a space after a comma. eg. "token1  token2" or "token1, token2" which would also become "token1  token2"

I think strings.Fields does what we would want.

Might also want to add tests for these cases of multiple white spaces

shifucun: Updated and added tests.

internal/server/statvar_search.go
```diff
@@ -0,0 +1,85 @@
...
+		if c == len(tokens) {
+			svList = append(svList, sv)
+		}
+	}
+	// Sort stat vars by ID then by number of PV
+	sort.Strings(svList)
+	sort.SliceStable(svList, func(i, j int) bool {
+		ranking := index.ranking
+		return ranking[svList[i]].ApproxNumPv < ranking[svList[j]].ApproxNumPv
+	})
```
pradh, line 70: Wondering about the sort order here.. I'd think we sort by num-pv first, and for those with the same num-pv, sort by ID?

pradh (approved):

chejennifer (approved):

internal/server/statvar_search.go
shifucun: It's actually doing that, updated the comment.

shifucun:
Thanks!

internal/server/statvar_search.go
pradh: Would it more readable (and efficient?) to do a single sort, and in the function do something like:

```
  if (ranking[svList[i]].ApproxNumPv == ranking[svList[j]].ApproxNumPv) {
    return svList[i] < svList[j]
  } else {
    return ranking[svList[i]].ApproxNumPv < ranking[svList[j]].ApproxNumPv
  }
```

PR #544, opened 2021-06-16
Add stat var (group) filtering based on data existence for given places

shifucun:


internal/server/statvar_group.go
```diff
@@ -312,3 +343,36 @@ func computeSpecializedEntity(parentSvg string, childSvg string) string {
...
+	}
+	result := map[string]int{}
+	for _, rowKey := range rowList {
+		placeSv := keyTokens[rowKey]
+		token, _ := keyToTokenFn(rowKey)
+		if _, ok := baseDataMap[token]; ok {
+			result[placeSv.statVar]++
+		} else {
+			// TODO(shifucun): remove the else part once cache is build.
+			result[placeSv.statVar]++
```
shifucun, line 374: This effectively making the filter an no-op. Will remove this after cache is build

internal/server/statvar_group.go
```diff
@@ -187,18 +187,49 @@ func (s *Server) GetStatVarGroupNode(
...
+	if err != nil {
+		return nil, err
+	}
+	// Filter parent stat var groups
+	filteredParentStatVarGroups := []string{}
+	for _, item := range result.ParentStatVarGroups {
+		if c, ok := statExistence[item]; ok && c == len(places) {
+			filteredParentStatVarGroups = append(filteredParentStatVarGroups, item)
+		}
+	}
```
pradh, line 213: Can this be a function perhaps? (maybe have TODO)

internal/server/statvar_search.go
```diff
@@ -37,9 +39,31 @@ func (s *Server) SearchStatVar(
...
+			}
+		}
+		svList = filteredSvList
+		filteredSvgList := []string{}
+		for _, svg := range svgList {
+			if c, ok := statExistence[svg]; ok && c == len(places) {
+				filteredSvgList = append(filteredSvgList, svg)
+			}
+		}
+		svgList = filteredSvgList
```
pradh, line 63: Can this be shared with the other check?

pradh (approved):

internal/server/statvar_group.go
shifucun: Added a to do for a generic way of filtering.

internal/server/statvar_search.go
shifucun: Done

shifucun:
Thanks for review!

PR #545, opened 2021-06-16
Return stat var and stat var group name from search

shifucun:
Name is needed for display purpose.

Also change the ranking by number of pvs then name.


internal/server/statvar_search.go
```diff
@@ -56,31 +57,46 @@ func searchTokens(tokens []string, index *SearchIndex) ([]string, []string) {
...
+				Dcid: sv,
+				Name: index.ranking[sv].RankingName,
+			})
 		}
 	}
 	// Sort stat vars by number of PV; If two stat vars have same number of PV,
-	// then order alphabetically.
-	sort.Strings(svList)
+	// then order by the stat var (group) name.
 	sort.SliceStable(svList, func(i, j int) bool {
```
pradh, line 71: nit: if stable order is not needed, we can just use Strings

internal/server/statvar_search.go
```diff
@@ -56,31 +57,46 @@ func searchTokens(tokens []string, index *SearchIndex) ([]string, []string) {
...
 	})
 
-	svgList := []string{}
+	svgList := []*pb.EntityInfo{}
 	for svg, c := range svgCount {
 		if c == len(tokens) {
-			svgList = append(svgList, svg)
+			svgList = append(svgList, &pb.EntityInfo{
+				Dcid: svg,
+				Name: index.ranking[svg].RankingName,
```
pradh, line 87: So this name comes from StatVarGroups proto. Wonder if we want a consistent approach for getting names.

Options
1.  use the SVG/SV names from StatVarGroups stored in Mixer memory (what's happening here)
2. use the name triple of SV and SVG
3. compute your own SVG/SV (I think you do this elsewhere)

I assume we want to get rid of (3).  The question is whether you want to do (1) or (2). If (2), then can we remove RankingName from the search-index and save memory?

pradh: Having TODO in this PR is fine.

pradh (approved):

internal/server/statvar_search.go
shifucun: yea, this could help for testing

internal/server/statvar_search.go
shifucun: If doing 2, does mixer need to call get_property_values for all the SV/SVG when building the search index?

pradh: StatVarGroups includes 2 names (at least for SV), search name and display name.

In (2), we will still include the search name used for index building, but perhaps need not include display name. But it need not be stored in mixer memory as RankingName?

We can also do (1), in which case we cut additional RPCs (which maybe saves some latency).  And in that case, we probably want a separate name map of SV/SVG => name ?

shifucun: Option 1 with additional name map sounds good.

I realized the name I returned here is the search name not the display name.

shifucun: Updated to use display name now

pradh (approved):

PR #548, opened 2021-06-17
Do not return stat var group (dc/g/Root) root when get path

shifucun:
This is a dummy node for linking the top level stat var groups, but not used by client.

### DIFF
--- a/internal/server/statvar_group.go
+++ b/internal/server/statvar_group.go
@@ -28,6 +28,7 @@ import (
 )
 
 const (
+	svgRoot            = "dc/g/Root"
 	autoGenSvgIDPrefix = "dc/g/"
 	svgDelimiter       = "_"
 )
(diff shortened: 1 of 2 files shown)


internal/server/statvar_group.go
```diff
@@ -295,6 +296,9 @@ func (s *Server) GetStatVarPath(
 	for {
 		if parents, ok := s.cache.ParentSvg[curr]; ok {
 			curr = parents[0]
+			if curr == svgRoot {
```
pradh, line 299: There will be a different PR for the current flow? (i.e., with full hierarchy showing up in the client)

pradh (approved):

internal/server/statvar_group.go
shifucun: I am thinking to do that in mixer. So instead of relying on triples cache, mixer will read from the full hierarchy, and return the queried node.

PR #554, opened 2021-06-18
Rollout data version: 2021_06_17_18_20_41; This includes PV based stat var group

shifucun:
If we think the new hierarchy looks, maybe change flume to produce it by default?


test/integration/get_statvar_path_test.go
```diff
@@ -30,28 +33,21 @@ func TestGetStatVarPath(t *testing.T) {
...
-			"dc/g/Person_Age_ArmedForcesStatus_Employment_EmploymentStatus",
-			[]string{
-				"dc/g/Person_Age_ArmedForcesStatus_Employment_EmploymentStatus",
-				"dc/g/Person_Age_Employment_EmploymentStatus",
-				"dc/g/Person_Age_Employment",
-				"dc/g/Person_Age",
-				"dc/g/Demographics",
-			},
+			"dc/g/Person_DetailedLevelOfSchool-EnrolledInGrade10",
+			"school.json",
```
pradh, line 50: what is the plan here, are we going to promote this to PROD?

pradh (approved):

test/integration/get_statvar_path_test.go
shifucun: Once the website change is also in, we can do some thorough testing. This could be candidate for next week's prod release.

pradh: Probably want a newer PV candidate (built on Monday), so it includes BlockGroups. But the plan SG.

PR #555, opened 2021-06-18
Restrict number of search results for stat var

shifucun:


internal/server/statvar_search.go
```diff
@@ -53,6 +55,12 @@ func (s *Server) SearchStatVar(
 		svList = filter(svList, statExistence, len(places))
 		svgList = filter(svgList, statExistence, len(places))
 	}
+	if len(svList) > maxNumResult {
+		svList = svList[0:maxNumResult]
+	}
+	if len(svgList) > maxNumResult {
+		svgList = svgList[0:maxNumResult]
+	}
```
beets, line 63: as a todo, return the total number of results available? so we can add text to the effect of "x more stat vars"

beets (approved):

internal/server/statvar_search.go
shifucun: Done

pradh (approved):

shifucun:
Thanks for review!

PR #559, opened 2021-06-23
Do not compute specialized entity

shifucun:


pradh (commented):
Sorry, this will be useful when we use triples cache, so maybe keep it?

PR #567, opened 2021-06-29
Add count of stat vars with data in SVG info cache

shifucun:


internal/server/statvar_group.go
```diff
@@ -228,33 +228,37 @@ func (s *Server) GetStatVarGroupNode(
...
 		}
-		filteredParentStatVarGroups := []string{}
-		for _, item := range result.ParentStatVarGroups {
-			if c, ok := statExistence[item]; ok && c == len(places) {
-				filteredParentStatVarGroups = append(filteredParentStatVarGroups, item)
+		// Count for current node.
+		result.NumDescendentStatVars = 0
+		if existence, ok := statVarCount[svg]; ok && len(existence) > 0 {
+			for _, count := range existence {
+				// Use the largest count among all places
```
pradh, line 239: If we are doing intersection, should we pick smallest I wonder?

pradh: Oh, now we are changing it to not be a strict check that all places match (c == len(places) check) ?

pradh (approved):

internal/server/statvar_group.go
shifucun: I was thinking relaxing it to show more. After more thoughts, keeping the intersection makes more sense for all the tools, esp scatter and map where we want to avoid missing data.

shifucun:
Thanks for review!

PR #572, opened 2021-07-01
Return union of all valid stat vars for all places

shifucun:


internal/server/statvar_group.go
```diff
@@ -235,32 +235,30 @@ func (s *Server) GetStatVarGroupNode(
...
 		}
 		// Count for current node.
-		if existence, ok := statVarCount[svg]; ok && len(existence) == len(places) {
+		result.NumDescendentStatVars = 0
+		if existence, ok := statVarCount[svg]; ok && len(existence) > 0 {
 			for _, count := range existence {
-				// Use the smallest count among all places.
-				if count < result.NumDescendentStatVars {
+				// Use the largest count among all places.
+				if count > result.NumDescendentStatVars {
```
pradh, line 242: Is this still an under estimation, if the SVs for the places are non-overlapping?  But probably okay for now...

internal/server/statvar_group.go
```diff
@@ -235,32 +235,30 @@ func (s *Server) GetStatVarGroupNode(
...
 					}
 				}
-			} else {
-				item.NumDescendentStatVars = 0
 			}
 		}
 		// Filter child stat vars
 		for _, item := range result.ChildStatVars {
-			if existence, ok := statVarCount[item.Id]; !ok || len(existence) != len(places) {
+			if existence, ok := statVarCount[item.Id]; ok && len(existence) == 0 {
```
pradh, line 261: a bit confused... if it is !ok, we still want to set HasData = false?

if so !ok || len(existence) == 0 ?

pradh (approved):

internal/server/statvar_group.go
shifucun: Updated

shifucun:
Thanks for review!

PR #574, opened 2021-07-02
Add new tree widget endpoint GetStatVarSummary

chejennifer:
added new endpoint GetStatVarSummary which given a list of stat vars, returns the stat var summaries for those stat vars


proto/mixer.proto
```diff
@@ -567,6 +567,43 @@ message SearchStatVarResponse {
   repeated EntityInfo stat_var_groups = 2;
 }
 
+message StatVarSummary {
+  message Place {
```
shifucun, line 571: nit: there are some indent issues here.

proto/mixer.proto
```diff
@@ -567,6 +567,43 @@ message SearchStatVarResponse {
...
+message GetStatVarSummaryRequest {
+  // A list of stat var dcids
+  repeated string stat_vars = 1;
+}
+
+message GetStatVarSummaryResponse {
+  map<string, StatVarSummary> stat_var_summary = 1;
+}
+
+message FilterStatVarRequest {
```
shifucun, line 595: Maybe we can make the API "CheckPlaceStats"? So the request is the same, but response is the raw data of mapping of place -> stat_var -> boolean. This way, user can use the response with more flexibility.

Also this is very similar to /place/stat-vars, where request have N places, and returns all the stat vars. Here, instead of return all stat vars, we check if given stat vars are valid (which in theory can be done if user just convert /place/stat-vars response to a Set). Maybe we can make this an extension of that, like /place/stat-vars/existence?

shifucun: Add a message comment on what it does?

internal/server/statvar_group.go
```diff
@@ -378,3 +378,58 @@ func countStatVar(
...
+			return &statVarSummary, nil
+		},
+		nil,
+		false, /* readBranch */
+	)
+	if err != nil {
+		return nil, err
+	}
+	result := &pb.GetStatVarSummaryResponse{}
+	result.StatVarSummary = map[string]*pb.StatVarSummary{}
```
shifucun, line 407: nit can do result := &pb.GetStatVarSummaryResponse{StatVarSummary: map[string]*pb.StatVarSummary{}}

shifucun (commented):
Thanks for adding this, very fast progress!

proto/mixer.proto
chejennifer: Since we just want the existence of the stat var for any of the list of places in the request, maybe response of stat_var -> boolean makes more sense?

shifucun: That could be an option. It would be nice to see how clients will use the response. For example, client may want, the union of the stat vars, the intersection of the stat vars, or just the count. Can you try to see what it looks like if the response is a list / map[bool] for each case?

With that, I wonder is it better to key by place or stat var id. If the first level key is place id, could we make this API part of the /place/stat-var API with additional filter (stat var ids).

proto/mixer.proto
chejennifer: done

proto/mixer.proto
chejennifer: removing this endpoint from this PR as it requires some additional thought and design

internal/server/statvar_group.go
chejennifer: done

shifucun (approved):

PR #576, opened 2021-07-08
Update stat var union API

chejennifer:
Update current GetPlaceStatVarUnion api to filter for a given list of stat vars. This unblocks checking if selected stat vars in the tree widget are available when place list is updated, and we will consider redesigning the place/stat-var API after tree widget is completed.


internal/server/place_stat_vars.go
```diff
@@ -135,6 +135,21 @@ func (s *Server) GetPlaceStatVarsUnion(
 func (s *Server) GetPlaceStatVarsUnionV1(
 	ctx context.Context, in *pb.GetPlaceStatVarsUnionRequest) (
 	*pb.GetPlaceStatVarsUnionResponseV1, error) {
+	statVars := in.GetStatVars()
+	dcids := in.GetDcids()
+	if len(statVars) > 0 && len(dcids) > 0 {
```
shifucun, line 140: Can you add a comment for why using separate cache here.

chejennifer: done

pradh (approved):

shifucun (approved):

PR #578, opened 2021-07-09
Fix a translator bug that leads to infinite loop

shifucun:
This is responsible for the timeout of the following command:

`curl --request POST --url https://api.datacommons.org/query  --header 'content-type: application/json' --data '{ "sparql": "SELECT distinct ?d  WHERE { ?ccdt typeOf ChemicalCompoundDiseaseTreatment . ?ccdt compoundID ?c . ?c drugName Prednisone . ?ccdt diseaseID ?dis . ?dis commonName ?d }  LIMIT 10"}'`

and likely the cause for some of the previous bio queries where long Triples tables join are performed

### DIFF
--- a/internal/translator/translate.go
+++ b/internal/translator/translate.go
@@ -771,9 +771,19 @@ func getSQL(
 			}
 			futureTables = append(futureTables, otherCol.Table)
 		}
-		for _, v := range futureTables {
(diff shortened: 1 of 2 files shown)


pradh (approved):

shifucun:
Thanks for review!

PR #581, opened 2021-07-15
[tree widget] update blocklist

chejennifer:


test/integration/golden_response/search_statvar/employed_16.json
```diff
@@ -1,920 +0,0 @@
-{
-  "statVars": [
```
pradh, line 2: I don't understand why this is dropped?

pradh: nm, you renamed it.

pradh (approved):

test/integration/golden_response/search_statvar/employed_16.json
chejennifer: I replaced employed_16.json and employed_16_blocklist.json with accommodation_food_services.json and accommodation_food_services_blocklist.json because employed_16 doesn't have anything that would be blocklisted anymore

PR #584, opened 2021-07-20
Fix a bug when building SQL JOIN conditions

shifucun:
This was introduced in https://github.com/datacommonsorg/mixer/pull/578. Both PRs try to solve the issues when current table has all been used in JOIN conditions, then need to pick a table to extend the join graph.

Also:
* Add some more comments
* Remove a over-done type alias

Need to update the landing page config tests here as well

### DIFF
--- a/internal/translator/translate.go
+++ b/internal/translator/translate.go
@@ -65,8 +65,6 @@ type ProvInfo struct {
 // Graph represents the struct for terms matching.
 type Graph map[interface{}]map[interface{}]struct{}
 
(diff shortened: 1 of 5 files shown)


pradh (approved):
Might be useful to have some description / context of the bug and perhaps a unit test?

shifucun:
> Might be useful to have some description / context of the bug and perhaps a unit test?

Thanks for review! Updated description. I am planning to modularize the functions a bit more and able to write smaller unit test in in follow up PR.

PR #587, opened 2021-08-02
Separate proto into multiple files

shifucun:
The number of protobuf and API has increased dramatically in the past. This PR try to separate based on logical function. Also

* Remove unused protobuf
* Remove old GetPlaceStatVarsUnion API which is not used any more.


proto/landing.proto
```diff
@@ -0,0 +1,59 @@
+// Copyright 2021 Google LLC
```
beets, line 1: naming nit: landing_page.proto to match the route?

proto/misc.proto
```diff
@@ -0,0 +1,73 @@
...
+  int32 max_results = 2;
+}
+
+// Search response from mixer.
+message SearchResponse {
+  // Matching entities broken down by type.
+  repeated SearchResultSection section = 1;
+}
+
+// Etmpty request to get version information.
```
beets, line 59:
```suggestion
// Empty request to get version information.
```

beets (approved):
thanks for this!

were the protos split according to the api route? it would be nice to have a comment at the top of each file noting what should go in there.

shifucun:
> thanks for this!
> 
> were the protos split according to the api route? it would be nice to have a comment at the top of each file noting what should go in there.

Very good idea, added the REST URL as file comment

proto/misc.proto
shifucun: Done

proto/landing.proto
shifucun: I am thinking putting all the landing page in this file, and have naming /landing/place(protein | disease). what do you think?

beets: yea, that sounds good. we can further split it up if it becomes too big..

shifucun: Yep, updated to /internal/place(protein) per discussion offline

shifucun:
Thanks for review and suggestions!

PR #588, opened 2021-08-02
Change landing page to place page

shifucun:
As we will have multiple node pages other than place page, this is to make the naming precise.


beets (approved):
thanks for making things consistent!

PR #589, opened 2021-08-03
Add an API to get bio page data

shifucun:
This now retrieves portion of the data with node traversal. When cache data is ready, it should use the pre-computed sub-graph from Bigtable.


test/integration/golden_response/get_protein_page_data/p53_human.json
```diff
@@ -0,0 +1,1711 @@
+{
```
shifucun, line 1: @Spaceenter, this is a realistic cache data

proto/internal.proto
```diff
@@ -60,4 +68,25 @@ message GetPlacePageDataResponse {
...
+  message LinkedNodes {
+    string property = 1;
+    PropertyDirection direction = 2;
+    repeated GetProteinPageDataResponse nodes = 3;
+  }
+  // A group of neighbour nodes by predicate.
+  repeated LinkedNodes neighbours = 1;
+  // Value of the node, could be dcid or node string value.
+  string value = 2;
 }
```
pradh, line 92: This seems like a general proto, not protein specific. Can we reuse for others?

pradh: So, generally, there is a (type, dcid) and response is a graph?

shifucun: Moved the proto to common.proto. 

Right, for disease, there will be a separate config and a subgraph tailored for the disease page.

shifucun:
Thanks for review!

pradh (approved):

proto/internal.proto
shifucun: Make the API to be more generic with naming changed from protein to bio

PR #590, opened 2021-08-10
Data release 2021_08_09_17_57_11

chejennifer:


test/integration/golden_response/search_statvar/female.json
```diff
@@ -1,5 +1,33 @@
 {
   "statVars": [
+    {
```
beets, line 3: is this the order with which we will show stat var search results? if that's the case, should the new WHO vars be ranked lower than Count_Person_Female (which should be ranked above Count_Death_Female)?

beets (approved):
there are probably some sv search ranking changes to do, but it shouldn't hold up the release, to staging anyway.

pradh (approved):

PR #592, opened 2021-08-18
Use DailyWeather instead of WeeklyWeather table

shifucun:
image


pradh (approved):
Thanks for the change!

pradh:
Any way to keep both?

pradh:
> Any way to keep both?

One has monthly data from 1950, and the other has daily data.

shifucun:
> > Any way to keep both?
> 
> One has monthly data from 1950, and the other has daily data.

We can introduce "MonthlyWeatherObservation" and "DailyWeatherObservation" type to differentiate them? I tried to use the observationPeriod predicate but it does not work right away, it may need some more translation change.

pradh:
> > > Any way to keep both?
> > 
> > 
> > One has monthly data from 1950, and the other has daily data.
> 
> We can introduce "MonthlyWeatherObservation" and "DailyWeatherObservation" type to differentiate them? I tried to use the observationPeriod predicate but it does not work right away, it may need some more translation change.

Sure, since this is all only in the Browser, I think that type change would not be visible anywhere and we can improve on this later. Thanks!

This is so that we can demo that we have long historic data + predictions.

PR #596, opened 2021-08-20
Use new ObsCollection cache for USA->County stat query

shifucun:


pradh (approved):
Thanks for the quick change!

PR #603, opened 2021-08-31
Add a parser for TMCF

shifucun:
This only supports resolved TMCF that does not have entity to entity reference.


internal/parser/tmcf.go
```diff
@@ -0,0 +1,95 @@
...
+			}
+			result[table].ColumnInfo[parts[1]] = &Column{Node: node, Property: head}
+		} else if strings.HasPrefix(body, base.DCS) {
+			// Schema mapping
+			schema := strings.TrimPrefix(body, base.DCS)
+			if _, ok := result[table].NodeSchema[parts[1]]; !ok {
+				result[table].NodeSchema[node] = map[string]string{}
+			}
+			result[table].NodeSchema[node][head] = schema
+		}
```
pradh, line 92: This can be any of: namespace prefix (dcs:, dcid:, schema:), string, number

internal/parser/tmcf.go
```diff
@@ -0,0 +1,95 @@
...
+				return nil, status.Errorf(codes.Internal, "Invalid input for Column:\n%s", line)
+			}
+			if table == "" || node == "" || table != parts[0] {
+				return nil, status.Errorf(codes.Internal, "Invalid input for Column:\n%s", line)
+			}
+			result[table].ColumnInfo[parts[1]] = &Column{Node: node, Property: head}
+		} else if strings.HasPrefix(body, base.DCS) {
+			// Schema mapping
+			schema := strings.TrimPrefix(body, base.DCS)
+			if _, ok := result[table].NodeSchema[parts[1]]; !ok {
```
pradh, line 88: can check that table and node are not empty here.

pradh (approved):

internal/parser/tmcf.go
shifucun: Done

internal/parser/tmcf.go
shifucun: I see. changed to removing these prefix

shifucun:
Thanks for review!

PR #607, opened 2021-09-02
Added a parser for complex value node

shifucun:
This code copies logic from https://github.com/datacommonsorg/import/blob/master/util/src/main/java/org/datacommons/util/ComplexValueParser.java


internal/parser/complex_value.go
```diff
@@ -0,0 +1,135 @@
...
+//  [<unit> <val>]
+//  [<unit> <startval> <endval>]
+//  [<unit> - <endval>]
+//  [<unit> <startval> -]
+//
+// 2. a GeoCoordinates type, coded as one of:
+//  [LatLong <lat_value> <long_value>]
+//  [<lat_value> <long_value> LatLong]
+func ParseComplexValue(val string) string {
+	trimmed := strings.Trim(val, "[]")
```
pradh, line 42: Does caller trim space?

pradh (approved):

internal/parser/complex_value.go
shifucun: Yes, caller (which will be in the csv loader) will trim for each cell)

PR #609, opened 2021-09-03
[tmcf + csv] Load tmcf and csv from GCS; Update server flags

shifucun:
- Now have a flag for each storage: bigquery, base bigtable, branch bigtable and gcs.
- Read tmcf and csv from GCS, parse tmcf into table schema and read in csv rows.
- Next step is to construct time series from each csv row based on table schema.

### DIFF
--- a/internal/store/memdb.go
+++ b/internal/store/memdb.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/csv"
 	"io"
+	"log"
 	"strings"
 
 	"cloud.google.com/go/storage"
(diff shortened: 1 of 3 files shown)


internal/store/memdb.go
```diff
@@ -15,30 +15,110 @@
...
+			if _, err := io.Copy(buf, r); err != nil {
+				return err
+			}
+			schemaMapping, err = parser.ParseTmcf(buf.String())
+			if err != nil {
+				return err
+			}
+			break
+		}
+	}
```
pradh, line 83: If you have multiple TMCFs and CSVs, how do we match them up?

pradh (approved):

internal/store/memdb.go
pradh: One option is we could have have a directory object per import, where each directory object has one TMCF and one or more CSVs.

shifucun: The argument "gcs_object_prefix" is used to hold one import. I make it more clear in the description.

pradh: SG, then

pradh (approved):

PR #610, opened 2021-09-03
[tmcf + csv] Function to populate stat series with csv records

shifucun:


internal/store/memdb.go
```diff
@@ -39,6 +42,106 @@ func (memDb *MemDb) Load(tmcfFile string) error {
...
+		}
+		if v, ok := meta["measurementMethod"]; ok {
+			allNodes[node].meta.MeasurementMethod = v
+		}
+		if v, ok := meta["unit"]; ok {
+			allNodes[node].meta.Unit = v
+		}
+		if v, ok := meta["scalingFactor"]; ok {
+			allNodes[node].meta.ScalingFactor = v
+		}
```
pradh, line 83: These can be constants or can be in the column too? e.g., https://github.com/datacommonsorg/data/blob/master/scripts/un/energy/un_energy.tmcf#L8-L10

internal/store/memdb.go
```diff
@@ -39,6 +42,106 @@ func (memDb *MemDb) Load(tmcfFile string) error {
...
+	allNodes := map[string]*nodeObs{}
+	// Initialize observation entries with the fixed schema
+	for node, meta := range schemaMapping.NodeSchema {
+		if typ, ok := meta["typeOf"]; ok && typ == "StatVarObservation" {
+			allNodes[node] = &nodeObs{
+				statVar: meta["variableMeasured"],
+				meta:    &pb.StatMetadata{},
+			}
+		}
+		if v, ok := meta["measurementMethod"]; ok {
```
pradh, line 75: observationPeriod?

internal/store/memdb.go
```diff
@@ -39,6 +42,106 @@ func (memDb *MemDb) Load(tmcfFile string) error {
...
+			}
+			if col.Property == "value" {
+				allNodes[n].value = cell
+			}
+			if col.Property == "observationDate" {
+				allNodes[n].date = cell
+			}
+			if col.Property == "observationAbout" {
+				allNodes[n].place = cell
+			}
```
pradh, line 112: I wonder if its more flexible to make allNodes[n][<property>] = cell ?

Then basically any property can be a column or constant.

internal/store/memdb.go
```diff
@@ -39,6 +42,106 @@ func (memDb *MemDb) Load(tmcfFile string) error {
...
+		// Derive node property and value for observation.
+		for _, col := range schemaMapping.ColumnInfo[colName] {
+			n := col.Node
+			if _, ok := allNodes[n]; !ok {
+				continue
+			}
+			if col.Property == "value" {
+				allNodes[n].value = cell
+			}
+			if col.Property == "observationDate" {
```
pradh, line 107: nit: observationDate can be a constant too (if the whole CSV is for a single year)

pradh (approved):
Nice!

internal/store/memdb.go
shifucun: Yes, that makes sense. I added a TODO above to handle the missing metadata.

shifucun (commented):
Thanks for review!

PR #611, opened 2021-09-08
Use in-memory DB data in /v1/stat/set/series API (used by timeline tool)

shifucun:


pradh (approved):
Nice!

Could you clarify in the description (or readme?) how this is guarded in prod?

shifucun:
> Nice!
> 
> Could you clarify in the description (or readme?) how this is guarded in prod?

Thanks! Added in developer guide.

PR #612, opened 2021-09-08
GKE deployment configs for private mixer instance

shifucun:
- Use project `datcom-mixer-private`
- Add a "bucket" field in the project config yaml
- Specify "--use_gcs_data", "--gcs_bucket" for private instance


deploy/gke/private.yaml
```diff
@@ -0,0 +1,22 @@
...
+# limitations under the License.
+
+project: datcom-mixer-private
+ip: 34.117.109.110
+domain: mixer.endpoints.datcom-mixer-private.cloud.goog
+api_title: DataCommons API (Private)
+region: us-central1
+nodes: 1
+store:
+bucket: datcom-public
```
pradh, line 22: Do you mean the bucket to be called datcom-public in a private instance?

deploy/overlays/private/kustomization.yaml
```diff
@@ -0,0 +1,59 @@
...
+          maxUnavailable: 75%
+      template:
+        spec:
+          containers:
+            - name: mixer
+              args:
+                - --mixer_project=$(MIXER_PROJECT)
+                - --use_gcs_data=true
+                - --gcs_bucket=datcom-public
+                - --gcs_object_prefix=iea
```
pradh, line 56: So for now, we support a single directory with one import? Out of curiosity where do we get provenance info?

deploy/overlays/private/kustomization.yaml
```diff
@@ -0,0 +1,59 @@
...
+                - --use_bigquery=false
```
pradh, line 57: if use_base_bt=false, does it mean this won't do merge?

pradh (approved):
LG. A few questions for my understanding...

deploy/gke/private.yaml
shifucun: Good point, I have been putting the test files in this bucket. Can create a dedicated bucket later for this instance.

deploy/overlays/private/kustomization.yaml
shifucun: There is no provenance for the private import. I am thinking it being the "bucket + prefix + tmcf" information. When there is an admin page. User can give a provenance url there.

deploy/overlays/private/kustomization.yaml
shifucun: Yes, I want to start with data from in-memory cache only, and test out other APIs

shifucun:
Thanks for review!

deploy/overlays/private/kustomization.yaml
pradh: Another option is to have a simple manifest file that specifies the URL (and maybe even tmcf and csv paths so we need not have one directory restriction)?  For future...

shifucun: That works. This would be easy for user with "simple" case of one dataset. A manifest would be good for users with multiple datasets.

PR #613, opened 2021-09-09
A few fix to private mixer

shifucun:
- Use base BT tables, so website can run off this. Otherwise there are too many APIs not working.
- Use GCP project "datcom-mixer-statvar" as "datcom-mixer-private" has org policy issue for external load balancer
- Check Bigtable client existence in pure private mode
- Remove quote of TMCF schema string


internal/parser/tmcf.go
```diff
@@ -93,6 +93,7 @@ func ParseTmcf(tmcf string) (map[string]*TableSchema, error) {
 			for _, prefix := range []string{"dcs:", "dcid:", "schema:"} {
 				schema = strings.TrimPrefix(schema, prefix)
 			}
+			schema = strings.Trim(schema, "\"")
```
pradh, line 96: comment? Is it to avoid enclosing quotes in string values, or something else?

pradh (approved):

internal/parser/tmcf.go
shifucun: Yes, it's for enclosing quotes like "P1M" in TMCF, added comment

PR #614, opened 2021-09-09
[Private DC] support stat var group node API

shifucun:
All the private stat vars are added under "dc/g/Private" categories.

This change makes the private stat vars explorer-able from the hierarchy widget.

### DIFF
--- a/internal/store/memdb.go
+++ b/internal/store/memdb.go
@@ -20,6 +20,7 @@ import (
 	"io"
 	"log"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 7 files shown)


internal/store/memdb.go
```diff
@@ -52,6 +53,34 @@ func (memDb *MemDb) ReadSeries(statVar, place string) []*pb.Series {
...
+					break
+				}
+			}
+		}
+		if valid {
+			hasDataStatVars = append(hasDataStatVars, statVar)
+		} else {
+			noDataStatVars = append(noDataStatVars, statVar)
+		}
+	}
```
pradh, line 78: nit: maybe this can be precomputed on load?

pradh (approved):
Nice

shifucun:
Thanks!

PR #616, opened 2021-09-11
[Private DC] Various feature update

shifucun:
- Implement /stat/set/within-place API. Also fixed an old bug, that the within-place cache is not used when "date" is provided.
- Fix a bug to check the memory database existence.
- Updated TestOption to be a struct.
- Implement /stat-var/path API.


shifucun:
@pradh Can you take a look before Monday, thanks!

internal/server/stat_point.go
```diff
@@ -155,80 +155,121 @@ func (s *Server) GetStatSetWithinPlace(
...
+						Metadata: &pb.StatMetadata{ImportName: cohort.ImportName},
 					}
+					result.Data[statVar].Metadata[cohort.ImportName] = metaData
 				}
 			}
 		}
-		if gotResult {
-			return result, nil
+	}
+	// Check if need to read from memory database.
```
pradh, line 219: nit: this was already a long function doing two things, and we're adding a third, might be good to modularize it into functions and comment like 1. read cache x, 2. read memdb, 3. read cache y, ...

internal/store/memdb.go
```diff
@@ -53,6 +63,56 @@ func (memDb *MemDb) ReadSeries(statVar, place string) []*pb.Series {
 	return []*pb.Series{}
 }
 
+// ReadPointValue reads one observation point.
+func (memDb *MemDb) ReadPointValue(statVar, place, date string) (
```
pradh, line 67: nit: document function contract based on "date" value?

pradh (approved):

internal/server/stat_point.go
shifucun: Okay, I will do this in a follow up PR.

shifucun:
thanks for review!

PR #617, opened 2021-09-13
[Private DC] Support manifest.json in the import

shifucun:
manifest.json contains the minimal required fields for an import: import_name, provenance_url and data_download_url.

This will be used to populate data metadata field, as well as in the data dashboard (to be implemented)

### DIFF
--- a/proto/common.proto
+++ b/proto/common.proto
@@ -47,5 +47,6 @@ message Manifest {
   // The url of the data, ususally the source url link
   string provenance_url = 2;
   // Data download link. For private import, this should be the GCS folder url.
+  // Example: https://pantheon.corp.google.com/storage/browser/datcom-public/test
   string data_download_url = 3;
 }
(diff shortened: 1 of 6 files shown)


proto/common.proto
```diff
@@ -38,4 +38,14 @@ message GraphNode {
...
   string value = 2;
+}
+
+// Holds the import manifest information.
+message Manifest {
+  // Name of the import
+  string import_name = 1;
+  // The url of the data, ususally the source url link
+  string provenance_url = 2;
+  // Data download link. For private import, this should be the GCS folder url.
```
pradh, line 49: Indicate example path? gs://... or something else

pradh (approved):

proto/common.proto
shifucun: Done

shifucun:
Thanks!

PR #619, opened 2021-09-13
[Private DC] Support API for /place/stat-vars, /v1/place/stat-vars/union

shifucun:
This is used by scatter tool.

Also fixed stat var path bug for private stat vars

### DIFF
--- a/internal/server/place_stat_vars.go
+++ b/internal/server/place_stat_vars.go
@@ -81,6 +81,12 @@ func (s *Server) GetPlaceStatVars(
 			resp.Places[dcid].StatVars = util.MergeDedupe(
 				resp.Places[dcid].StatVars, baseDataMap[dcid].([]string))
 		}
+		// Also merge from memdb
+		if !s.store.MemDb.IsEmpty() {
+			hasDataStatVars, _ := s.store.MemDb.GetStatVars([]string{dcid})
+			resp.Places[dcid].StatVars = util.MergeDedupe(
+				resp.Places[dcid].StatVars, hasDataStatVars)
+		}
 	}
 	return &resp, nil
 }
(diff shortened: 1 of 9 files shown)


internal/server/place_stat_vars.go
```diff
@@ -99,53 +105,72 @@ func keysToSlice(m map[string]bool) []string {
...
+				if _, ok := filterStatVarSet[sv]; ok {
+					set[sv] = true
+				}
+			}
 		}
 	}
-	return &pb.GetPlaceStatVarsUnionResponse{
-		StatVars: keysToSlice(set),
-	}, nil
+	result.StatVars = util.MergeDedupe(result.StatVars, keysToSlice(set))
```
pradh, line 174: Can this MergeDedupe be inside the memDb check?

internal/server/place_stat_vars.go
```diff
@@ -99,53 +105,72 @@ func keysToSlice(m map[string]bool) []string {
...
+) (*pb.GetPlaceStatVarsUnionResponse, error) {
+	// Check places
+	places := in.GetDcids()
+	if len(places) == 0 {
 		return nil, status.Error(
 			codes.InvalidArgument, "Missing required arguments: dcids")
 	}
+	// filtered stat vars
+	filterStatVars := in.GetStatVars()
+	filterStatVarSet := map[string]struct{}{}
```
pradh, line 118: Is this fixing a bug from before?

maybe add comment like "// Stat vars may have dups, so compute set."

pradh (approved):

internal/server/place_stat_vars.go
shifucun: Good point!

internal/server/place_stat_vars.go
shifucun: This is more of checking the stat vars efficiently with a map. Added a comment.

shifucun (commented):
Thanks for review!

PR #620, opened 2021-09-17
update statvarsummary

n-h-diaz:


proto/stat_var.proto
```diff
@@ -54,8 +54,52 @@ message StatVarSummary {
     // List of example places that have stats for this stat var
     repeated Place top_places = 2;
   }
+
+  message SeriesSummary {
```
shifucun, line 58: Can you add a TODO here for considering consolidate this with https://github.com/datacommonsorg/mixer/blob/master/proto/stat.proto#L36

shifucun (approved):

proto/stat_var.proto
n-h-diaz: done

n-h-diaz:
thanks for the review!

PR #621, opened 2021-09-17
update go proto files

n-h-diaz:


shifucun (approved):
Thanks for the update!

PR #624, opened 2021-09-22
Update replicas for staging and autopush mixer

shifucun:
- Autopush mixer is used by local dev and webdriver tests, should have more instances
- Staging mixer is only used by adhoc check during deployment, which does not need many replicas


deploy/gke/staging.yaml
```diff
@@ -19,5 +19,5 @@ region: us-central1
 ip: 34.107.161.252
 domain: staging.api.datacommons.org
 api_title: DataCommons API (Staging)
-nodes: 3
+nodes: 1
```
beets, line 22: should we keep this to at least 2, so that it's closer to prod setup (and not a lone instance?)

beets (approved):

PR #626, opened 2021-09-28
Update place stat var ranking test

shifucun:
This is to make the test passing after website change: https://github.com/datacommonsorg/website/pull/1141


beets (approved):
thanks for fixing the tests!

PR #628, opened 2021-10-04
Add logging to GetStats() API

shifucun:
We have seen several 200+ seconds latency on this API. Logging the params for better debugging.

### DIFF
--- a/internal/server/stat_series.go
+++ b/internal/server/stat_series.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"log"
 	"sort"
+	"time"
 
 	"cloud.google.com/go/bigtable"
 	pb "github.com/datacommonsorg/mixer/internal/proto"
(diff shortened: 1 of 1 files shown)


internal/server/stat_series.go
```diff
@@ -120,6 +121,12 @@ func (s *Server) GetStats(ctx context.Context, in *pb.GetStatsRequest) (
 	placeDcids := in.GetPlace()
 	statsVarDcid := in.GetStatsVar()
 
+	log.Printf(
+		"GetStats(): placeDcids: %s, statsVarDcid: %v",
+		placeDcids,
+		statsVarDcid,
+	)
```
pradh, line 128: How will this show up in the log? Is there any way to line it up with the request time (aka, how to tell a request with these args took N secs)?

If the requests generally succeeded, an option is to log the stats with latency when we return.

pradh (commented):
Thanks for this... one question

internal/server/stat_series.go
shifucun: Logged the processing time as well (it would be less than latency though) when it's > 30s.

internal/server/stat_series.go
```diff
@@ -166,6 +162,16 @@ func (s *Server) GetStats(ctx context.Context, in *pb.GetStatsRequest) (
...
+
+	latency := time.Since(ts).Seconds()
+	if latency > 30 {
+		log.Printf(
+			"GetStats() took %f seconds with: placeDcids: %s, statsVarDcid: %v",
+			latency,
+			placeDcids,
+			statsVarDcid,
+		)
+	}
```
pradh, line 174: Is it worth putting this as a defer function so it will run no matter wherever else we return from ?

pradh (approved):
Thanks

internal/server/stat_series.go
shifucun: That's a very good idea!

internal/server/stat_series.go
pradh: Nice.  If this works ok, we can probably add this to all APIs.

pradh (approved):

PR #629, opened 2021-10-06
Add preference for UN Energy data

beets:
Related to https://github.com/datacommonsorg/data/issues/524
A quick fix to prefer UN Energy data sources where available, especially since it's complete at the country level


shifucun:
Found the state only has eia data but no UN data, the country, state comparison chart is still off..

pradh:
What's an example? CA seems reasonable:
 
image

(nit: nearby for states seems not super meaningful?)

PR #650, opened 2021-10-23
Add Indian Census population to rank it over OECD for Indian State

pradh:
The goal is to avoid prefer Indian Census over OECD for Indian States/Districts/etc.

Screen Shot 2021-10-23 at 3 26 44 PM


pradh:
Thanks for the review!

PR #659, opened 2021-11-03
fix ranking for india

n-h-diaz:


pradh (approved):
Maybe wait for Bo to ack too

PR #666, opened 2021-11-19
Add new API stat/set/within-place/all; Update metadata hash mechanism

shifucun:
The new API returns all the source series data that is available.

This PR also updates the metadata hash. Previous, the hash is the import name, which works when picking one source for each stat var but not anymore when returning all sources. Here implements a simple hash store that use the counter of the metadata as the hash key.


proto/stat.proto
```diff
@@ -46,16 +46,25 @@ message PointStat {
...
 message PlacePointStat {
   // Keyed by place DCID.
   map<string, PointStat> stat = 1;
-  // Keyed by import name.
+  // Keyed by metadata hash.
   map<string, StatMetadata> metadata = 2;
 }
 
+message PlacePointStatAll {
+  // Keyed by place DCID.
```
pradh, line 64: nit: obsolete comment?

proto/stat.proto
```diff
@@ -46,16 +46,25 @@ message PointStat {
   // date in IOS-8601 format
   string date = 1;
   double value = 2;
+  // Full metadata of the stat
   StatMetadata metadata = 3;
+  // A hash string of the meta data, this is to be used together with a map from
+  // the hash to the full metatdata.
```
pradh, line 52: should metadata and meta_hash be in a oneof?  Note that you can move existing fields to under a oneof ("you can safely move a single field into a new oneof " ... https://developers.google.com/protocol-buffers/docs/proto)

proto/mixer.proto
```diff
@@ -179,6 +179,18 @@ service Mixer {
     };
   }
 
+  // Get the stat value for children places of certain place type at a given
```
pradh, line 182: elaborate on "stat value"?  i.e., stats from *all* series, from latest dates (or all dates?)...

internal/util/util.go
```diff
@@ -423,3 +428,27 @@ func Sample(m proto.Message, strategy *SamplingStrategy) proto.Message {
 	})
 	return m
 }
+
+// HashStore is a simple a store to produce hash for protobuf message
```
pradh, line 432: nit: s/a store/store/

internal/util/util.go
```diff
@@ -423,3 +428,27 @@ func Sample(m proto.Message, strategy *SamplingStrategy) proto.Message {
...
+
+// GetHash retrieves a hash string for a given protobuf message.
+// Note this should be restrict to a request scope.
+func (hs *HashStore) GetHash(m proto.Message) string {
+	if hs.store == nil {
+		hs.store = map[string]int{}
+	}
+	b, err := proto.Marshal(m)
+	if err != nil {
+		return "0"
```
pradh, line 445: Should we return an error here instead of a fake 0?

internal/util/util.go
```diff
@@ -423,3 +428,27 @@ func Sample(m proto.Message, strategy *SamplingStrategy) proto.Message {
...
+	store map[string]int
+}
+
+// GetHash retrieves a hash string for a given protobuf message.
+// Note this should be restrict to a request scope.
+func (hs *HashStore) GetHash(m proto.Message) string {
+	if hs.store == nil {
+		hs.store = map[string]int{}
+	}
+	b, err := proto.Marshal(m)
```
pradh, line 443: Should we worry about the stability of using serialized generic proto as a hash key, esp in the presence of map fields? given https://stackoverflow.com/a/60512308

If this is only needed for metadata for now, I wonder about making it specific to that.

internal/util/util.go
```diff
@@ -423,3 +428,27 @@ func Sample(m proto.Message, strategy *SamplingStrategy) proto.Message {
...
+	if err != nil {
+		return "0"
+	}
+	hash, ok := hs.store[string(b)]
+	if ok {
+		return fmt.Sprint(hash)
+	}
+	count := len(hs.store) + 1
```
pradh, line 451: why +1 ?

proto/stat.proto
```diff
@@ -46,16 +46,25 @@ message PointStat {
...
+  string meta_hash = 4;
 }
 
 message PlacePointStat {
   // Keyed by place DCID.
   map<string, PointStat> stat = 1;
-  // Keyed by import name.
+  // Keyed by metadata hash.
```
pradh, line 59: I wonder why not make it an index into metadata_list below? (fewer bytes stored overall)

internal/server/stat_point.go
```diff
@@ -90,26 +98,145 @@ func getStatSet(
...
+						}
+					}
+					tmpResult[statVar][metaHash].Stat[place] = ps
+				}
+			}
+		}
+	}
+
+	// Convert tmp result to result
+	// tmpResult := map[string]map[string]*pb.PlacePointStat{}
```
pradh, line 205: drop?

internal/util/util.go
```diff
@@ -423,3 +428,27 @@ func Sample(m proto.Message, strategy *SamplingStrategy) proto.Message {
...
 	return m
 }
+
+// HashStore is a simple a store to produce hash for protobuf message
+type HashStore struct {
+	store map[string]int
+}
+
+// GetHash retrieves a hash string for a given protobuf message.
+// Note this should be restrict to a request scope.
```
chejennifer, line 438: Does this mean across different requests, for the same proto message, we could get different hashes?

internal/server/stat_point.go
```diff
@@ -247,12 +379,148 @@ func (s *Server) GetStatSetWithinPlace(
...
+	}
+	// Check if need to read from memory database.
+	statVarInMemDb := false
+	for _, statVar := range statVars {
+		if s.store.MemDb.HasStatVar(statVar) {
+			statVarInMemDb = true
+			break
+		}
+	}
+	// Need to fetch child places if to read data from memory database or
```
chejennifer, line 488: nit: "...if to read data..." -> "...if need to read data ..."

internal/server/stat_point.go
shifucun: Done

internal/server/stat_point.go
shifucun: Done

internal/util/util.go

internal/util/util.go
shifucun: Changed to use hash function that generates deterministic number

internal/util/util.go
shifucun: Changed this to hash the concatenated StatMetadata field strings now.

internal/util/util.go
shifucun: Use hash number now

internal/util/util.go
shifucun: Use hash number now

proto/mixer.proto

proto/stat.proto
shifucun: Tried oneof but that makes the field nested a bit harder to access. Having the two fields here are okay too, since they can coexist potentially.

proto/stat.proto
shifucun: Good point. Moved the map as response level information.

proto/stat.proto
shifucun: Updated

chejennifer (approved):

proto/stat.proto
pradh: I see. Maybe some comment on which api sets which field between metadata and meta_hash might be helpful.

proto/stat.proto
```diff
@@ -48,20 +48,18 @@ message PointStat {
...
 message PlacePointStat {
   // Keyed by place DCID.
   map<string, PointStat> stat = 1;
-  // Keyed by metadata hash.
-  map<string, StatMetadata> metadata = 2;
 }
 
 message PlacePointStatAll {
-  // Keyed by place DCID.
+  // A list of PlacePointStat, with each one from one source.
```
pradh, line 62: Do you mean each PlacePointStat corresponds to one series?  Then, can we not have meta_hash in PlacePointStat instead of PointStat ?

pradh: Or perhaps you need it there for GetStatSetResponse ?

pradh (approved):
Thanks for the changes!

proto/stat.proto
shifucun: Right it's needed for GetStatSetResponse. I added meta_hash field for PlacePointStat as well, so PlacePointStatAll only has one meta_hash for each source now.

proto/stat.proto
shifucun: Done

shifucun:
Thank you all for the review!

proto/stat.proto
```diff
@@ -46,16 +46,20 @@ message PointStat {
   // date in IOS-8601 format
   string date = 1;
   double value = 2;
-  // Full metadata of the stat
+  // Full metadata of the stat. This is set when the proto is used statalone,
```
pradh, line 49: nit: standalone

proto/stat.proto
```diff
@@ -46,16 +46,20 @@ message PointStat {
...
-  // the hash to the full metatdata.
+  // the hash to the full metatdata. This is set in /stat/set/within-place/*
+  // APIs.
   uint32 meta_hash = 4;
 }
 
 message PlacePointStat {
   // Keyed by place DCID.
   map<string, PointStat> stat = 1;
+  // Hash of the matdata, which applies to all the PointStat in `stat` field.
```
pradh, line 61: nit: metadata

state which API this applies to?

proto/stat.proto
```diff
@@ -46,16 +46,20 @@ message PointStat {
   // date in IOS-8601 format
   string date = 1;
   double value = 2;
-  // Full metadata of the stat
+  // Full metadata of the stat. This is set when the proto is used statalone,
+  // ex, in place page cache.
   StatMetadata metadata = 3;
   // Hash of the metadata, this is to be used together with a map from
-  // the hash to the full metatdata.
+  // the hash to the full metatdata. This is set in /stat/set/within-place/*
```
pradh, line 53: nit: It doesn't apply to all right? maybe worth being precise on what APIs set this.

pradh (commented):
Just some comment improvement comments for a future PR.

PR #668, opened 2021-11-22
Data release 2021_11_19_23_36_10

shifucun:


test/integration/golden_response/get_place_page_data/asm.sample.json
```diff
@@ -6927,12 +6935,15 @@
...
-            "2019": 5
+            "2019": 4,
+            "2020": 11,
+            "2021": 8
           },
           "metadata": {
             "importName": "CycloneNOAA",
-            "provenanceUrl": "https://www.ncdc.noaa.gov/",
-            "measurementMethod": "DataCommonsAggregate"
+            "provenanceUrl": "https://www.ncdc.noaa.gov/ibtracs/index.php?name=ib-v4-access",
```
shifucun, line 6944: This is expected from the cyclone data refresh

test/integration/golden_response/get_place_page_data/county.sample.json
```diff
@@ -15034,6 +15034,27 @@
             "observationPeriod": "P1D"
           }
         },
+        "Annual_Emissions_CarbonDioxide_NonBiogenic": {
```
shifucun, line 15037: These are from the landing page config update

pradh (approved):
This looks good.

To confirm, there are golden files related to temperature right? And there isn't any diffs with them?

I originally thought the default model for US might change, but then the mixer ranking specifies the CCSM4 model - https://github.com/datacommonsorg/mixer/blob/13f1c4a3b21e14e02e1e75cf0788f537fa6b88d6/internal/server/ranking.go#L62-L65.  But also there are more import-names, and so we need to revisit that..

shifucun:
> This looks good.
> 
> To confirm, there are golden files related to temperature right? And there isn't any diffs with them?
> 
> I originally thought the default model for US might change, but then the mixer ranking specifies the CCSM4 model -
> 
> https://github.com/datacommonsorg/mixer/blob/13f1c4a3b21e14e02e1e75cf0788f537fa6b88d6/internal/server/ranking.go#L62-L65
> 
> . But also there are more import-names, and so we need to revisit that..

Update ranking sounds good. Also the map tool now show dropdown selection of all the sources.

PR #677, opened 2021-12-10
Fix a nil pointer deref in mixer

shifucun:
This is causing mixer crash and server restart in production: https://pantheon.corp.google.com/errors;service=gke_instances;time=P1D?project=datcom-website-prod

We should deploy this fix to prod ASAP


beets (approved):

beets:
do we know which query is causing this crash? we can test post push

shifucun:
> do we know which query is causing this crash? we can test post push

Not sure on that. It's from place page but I can not reproduce. We need logging for all the requests..

PR #678, opened 2021-12-10
Add an API to fetch place metadata

shifucun:


internal/server/places.go
```diff
@@ -241,3 +241,67 @@ func (s *Server) GetLocationsRankings(ctx context.Context,
...
+		}
+		raw := baseDataMap[place].(*pb.PlaceMetadataCache)
+		processed := pb.PlaceMetadata{}
+		metaMap := map[string]*pb.PlaceMetadataCache_PlaceInfo{}
+		for _, info := range raw.Places {
+			metaMap[info.Dcid] = info
+		}
+		processed.Self = &pb.PlaceMetadata_PlaceInfo{
+			Dcid: place,
+			Name: metaMap[place].Name,
```
beets, line 288: could there be some places without names (iirc there were some in india). will this result in nil or ''?

beets (approved):

internal/server/places.go
shifucun: It will be empty string. Added a test case for this.

PR #679, opened 2021-12-14
Use PlaceMetadata cache for parent places in place page

shifucun:


internal/util/util.go
```diff
@@ -387,3 +388,9 @@ func Sample(m proto.Message, strategy *SamplingStrategy) proto.Message {
 	})
 	return m
 }
+
+// TimeTrack is used to track function execution time.
+func TimeTrack(start time.Time, name string) {
```
pradh, line 393: nice

pradh (approved):

PR #685, opened 2021-12-29
SV hierarchy search updates

chejennifer:
- use trie data structure for sv hierarchy search index
- add synonyms to search index
- filter out sv and svg that are under miscellaneous/uncategorized when creating the search index


internal/server/resource/resource.go
```diff
@@ -65,42 +71,49 @@ type RankingInfo struct {
...
+		for _, c := range token {
+			if currNode.ChildrenNodes == nil {
+				currNode.ChildrenNodes = map[rune]*TrieNode{}
+			}
+			if _, ok := currNode.ChildrenNodes[c]; !ok {
+				currNode.ChildrenNodes[c] = &TrieNode{}
+			}
+			currNode = currNode.ChildrenNodes[c]
+		}
 		if isSvg {
```
shifucun, line 107: this should block be inside the for loop above? Now only the last currNode has SvgIds

internal/server/resource/synonyms.json
```diff
@@ -0,0 +1,11 @@
+{
```
shifucun, line 1: Should synonyms be bidirectional? so we also has "woman" to "female". 

Or should each synonyms be a set and we built the mapping from it?

internal/server/statvar_search.go
```diff
@@ -104,12 +104,33 @@ func searchTokens(
 	tokens []string, index *resource.SearchIndex) ([]*pb.EntityInfo, []*pb.EntityInfo) {
 	svCount := map[string]int{}
 	svgCount := map[string]int{}
+
+	// Get all matching sv and svg from the trie for each token
 	for _, token := range tokens {
-		for sv := range index.TokenSVMap[token] {
-			svCount[sv]++
+		currNode := index.RootTrieNode
```
shifucun, line 110: Can you add comments for each block of logic? Here is logic intensive code :)

shifucun: Also add some example case in the comment.

internal/server/statvar_search_test.go
```diff
@@ -24,25 +24,73 @@ import (
 )
 
 func TestSearchTokens(t *testing.T) {
+	token1 := resource.TrieNode{
```
shifucun, line 27: node1 ? and below

internal/server/statvar_hierarchy_util.go
```diff
@@ -91,31 +96,53 @@ func getIgnoredSVGHelper(
...
 // BuildStatVarSearchIndex builds the search index for the stat var hierarchy.
 func BuildStatVarSearchIndex(
 	rawSvg map[string]*pb.StatVarGroupNode,
 	blocklist bool) *resource.SearchIndex {
 	// map of token to map of sv/svg id to ranking information.
 	searchIndex := &resource.SearchIndex{
-		TokenSVMap:  map[string]map[string]struct{}{},
-		TokenSVGMap: map[string]map[string]struct{}{},
-		Ranking:     map[string]*resource.RankingInfo{},
+		RootTrieNode: &resource.TrieNode{},
```
shifucun, line 118: Can you add a timer on this function? Can use https://github.com/datacommonsorg/mixer/blob/master/internal/util/util.go#L393

shifucun (commented):
This is very cool! Look forward to seeing this in action!

internal/server/resource/resource.go
chejennifer: My current implementation is for only the last currNode to have the SvgIds. I thought about adding the SvgId to every node along the way but I think that will use more space?

chejennifer: My current implementation is for only the last currNode to have the SvgIds. I thought about adding the SvgId to every node along the way but I think that will use more space?

internal/server/resource/synonyms.json
chejennifer: good idea, updated

internal/server/statvar_search.go
chejennifer: done

internal/server/statvar_search_test.go
chejennifer: done

internal/server/statvar_hierarchy_util.go
chejennifer: done

internal/server/resource/resource.go
shifucun: For example when user search for "wom" (for "women"), would it have linked SvgIds?

internal/server/statvar_search.go
shifucun: nice!

internal/server/resource/resource.go
chejennifer: yes, because when user searches for "wom", the function will get to the "m: node and then traverse the rest of the subtrie that is rooted at the "m" node and get all the SvgIds and SvIds

internal/server/statvar_hierarchy_util.go
chejennifer: done

internal/server/resource/resource.go
shifucun: I see, that makes sense.

shifucun (approved):

PR #688, opened 2022-01-06
update stat/set/within-place APIs to handle different date granularities

chejennifer:
- In scatter tool, we will want to get the data for a given date, but the stat vars might have data with different time granularities. So in these APIs, when given a date, we want to return values for other date granularities that are also valid for the given date (eg. given 2018, 2018-01 and 2018-03-04 should both be valid).


chejennifer:
upon discussion, the use case for this change is too specific so we won't be making this change

PR #689, opened 2022-01-06
Create packages to group APIs and /internal/server/

shifucun:
This makes the API structure much more clear and unveils some odd dependencies. The placement of some functions and type definitions are still not optimal, leaving to later PR.


chejennifer (approved):
awesome, thanks for the cleanup!

PR #690, opened 2022-01-06
Move golden test files into the corresponding API implementation folder.

shifucun:
Changes in API can be located in a folder instead of several places. This also makes the testing and golden file update locally instead of always triggering all golden testing.

- Changed cloud build machine type to N1_HIGHCPU_32 so it can run locally.
- Remove the option to update golden files in Docker mode as it is never used


build/ci/cloudbuild.goldendiff.yaml
```diff
@@ -40,4 +40,4 @@ options:
   volumes:
     - name: tmp
       path: /tmp
-  machineType: "E2_HIGHCPU_32"
+  machineType: "N1_HIGHCPU_32"
```
beets, line 43: Could you document these build changes in the PR description?

test/e2e/place_statvar_ranking_test.go
```diff
@@ -198,8 +198,8 @@ func TestChartConfigRankings(t *testing.T) {
 			})
 
 			goldenFile := path.Join(goldenPath, c.goldenFile)
-			if generateGolden {
-				updateGolden(missingRankings, goldenPath, c.goldenFile)
+			if GenerateGolden {
```
beets, line 201: Were these other style updates automated? Which tool did you use to generate these? Would be nice to standardize on these so we can keep this in sync.

scripts/update_golden.sh
```diff
@@ -46,13 +46,4 @@ else
     ARG=""
 fi
 
-if [[ $DOCKER == "true" ]]; then
-  DOCKER_BUILDKIT=1 docker build --tag datacommons/mixer-golden-update  -f build/Dockerfile --target golden-update .
```
beets, line 50: could you also document this removal in the PR description?

test/e2e/setup.go
```diff
@@ -40,16 +40,11 @@ import (
...
+	UseMemdb bool
 }
 
-var generateGolden bool
-
-func init() {
-	flag.BoolVar(
-		&generateGolden, "generate_golden", false, "generate golden files")
-}
+var GenerateGolden = os.Getenv("GENERATE_GOLDEN") == "true"
```
beets, line 47: Does this need to be unset on exit?

beets (approved):
thanks for the cleanup!

build/ci/cloudbuild.goldendiff.yaml
shifucun: Done

scripts/update_golden.sh
shifucun: Done. The docker mode is never used, so removed to skip updating it.

test/e2e/place_statvar_ranking_test.go
shifucun: Updated using vscode IDE.

test/e2e/setup.go
shifucun: I think setting the env in the code has no effect to the actual environment. The env is set in update_golden.sh and scoped in that file, so would have no effect outside.

shifucun (commented):
Thanks for the review!

PR #691, opened 2022-01-07
Fix source ranking for CensusACS5YearSurvey_AggCountry

shifucun:
Looks like the measurement method is changed from "dcAggregate/CensusACS5yrSurvey" to "CensusACS5yrSurvey". Not sure if this intended? @pradh 

Also fixed a script issue in golden updater.

### DIFF
--- a/internal/server/placepage/golden/get_place_page_data/asm.sample.json
+++ b/internal/server/placepage/golden/get_place_page_data/asm.sample.json
@@ -7445,20 +7445,21 @@
         },
         "Count_Household": {
           "val": {
-            "2011": 115991452,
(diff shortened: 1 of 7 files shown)


pradh2 (approved):

pradh2:
> Looks like the measurement method is changed from "dcAggregate/CensusACS5yrSurvey" to "CensusACS5yrSurvey". Not sure if this intended? @pradh

I think the mixer spec should not have the dcAggregate/ prefix because prophet would disaggregate it and set a bool [here](http://google3/knowledge/answers/stats/proto/chart_store.proto;l=36;rcl=416206620)

Issue #683, opened 2021-12-17
Deprecate d/f/... cache, replace it with the 3 new split cache.


Spaceenter:
Notably, replace d/f/... with:

(a)
// ObsTimeSeries
//   <placeId^statVarId>, ChartStore containing ObsTimeSeries. 
kCacheObsTimeSeries[] = "3";

(b)
// ObsCollectionDateFrequency
//   <parentId^childType^statVarId^>, ChartStore with ObsCollection containing
//     the frequency of each date across places.
kCacheObsCollectionDateFrequency[] = "8";

(c)
// ObsCollection
//   <ancestorId^childType^statVarId^obsDate>, ChartStore with ObsCollection
//     of data values for child places.
//     If ShouldObsCollectionIncludeIndirectAncestors() == true:
//       The ancestors include:
//       * Direct parents linked via containedInPlace.
//       * Indirect ancestors defined by ShouldProduceAncestorObsCollection().
//     If ShouldObsCollectionIncludeIndirectAncestors() == false:
//       The ancestors include:
//       * Direct parents linked via containedInPlace.
//   <ancestorId^childType^statVarId^LATEST>, ChartStore with ObsCollection of
//     data values for child places for the latest date.
//     The ancestors include:
//     * Indirect ancestors defined by ShouldProduceAncestorObsCollection().
//     NOTE: Direct parents could be missing here.
kCacheObsCollection[] = "e";

Spaceenter:
The new cache is already in BT. 
After this change, we need to delete the d/f/... BT cache from Prophet. 
@shifucun @pradh

PR #693, opened 2022-01-13
Split d/f/ cache type to three difference cache types.

Spaceenter:
This address the issue: https://github.com/datacommonsorg/mixer/issues/683


shifucun:
Thanks! This makes it much more clear now!

shifucun (approved):

PR #694, opened 2022-01-14
Fix to use CorhortByRank instead of SeriesByRank when ranking corhorts

shifucun:
This fix is still a no-op since the ranking is override later by latest date. But good to make it correct.

### DIFF
--- a/internal/server/place/place_stat_date.go
+++ b/internal/server/place/place_stat_date.go
@@ -69,7 +69,7 @@ func GetPlaceStatDateWithinPlace(
 	for sv, data := range cacheData {
 		if data != nil && len(data.SourceCohorts) > 0 {
 			cohorts := data.SourceCohorts
-			sort.Sort(ranking.SeriesByRank(cohorts))
+			sort.Sort(ranking.CohortByRank(cohorts))
 			dates := []string{}
 			for date := range cohorts[0].Val {
 				dates = append(dates, date)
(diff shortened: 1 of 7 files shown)


beets (commented):
qq: will this also apply to a standalone requests (/stat/value)? e.g. requesting population of country/USA --> 2019 data from Census PEP, vs 2020 Decennial? my understanding was we would prefer latest if possible, but if we need this for per capita computation, then we might prefer a different mmethod which will more closely align with the numerator sv. for standalone population results, i think we would want to show that we have the latest 2020 census data.

maybe the other update we need to do is to bump up the ranking of Decennial for now until we have the option to align dates.

shifucun:
> qq: will this also apply to a standalone requests (/stat/value)? e.g. requesting population of country/USA --> 2019 data from Census PEP, vs 2020 Decennial? my understanding was we would prefer latest if possible, but if we need this for per capita computation, then we might prefer a different mmethod which will more closely align with the numerator sv. for standalone population results, i think we would want to show that we have the latest 2020 census data.
> 
> maybe the other update we need to do is to bump up the ranking of Decennial for now until we have the option to align dates.

If ranking Decennial higher, then time series API would only return 3 points. The original rational to prefer ranking (instead of latest date) is that all data are from same source when possible. But I feel this needs some more thoughts from different scenario, and maybe worth a one pager to discuss. I will open an issue for this. For now just keep the change that fixes the  ranking bug.

shifucun:
> 
Updated title.

pradh (approved):

shifucun:
thanks!

PR #697, opened 2022-01-19
Add config for production recon service

shifucun:


deploy/gke/recon-prod.yaml
```diff
@@ -0,0 +1,24 @@
+# Copyright 2022 Google LLC
```
pradh, line 1: do we need recon-autopush?

pradh (approved):
Nice!

deploy/gke/recon-prod.yaml
shifucun: Yep, we do need that. Have some issue on deploying it. Will send a PR when it's ready.

shifucun:
Thanks!

PR #699, opened 2022-01-20
Clean up some stat series conversion, filter logic

shifucun:
The ObsSeries struct is used for the old GetStats() API and has a hack for reusing fields. Split it into two structs for clean logic. This is to prepare for merging time series cache data.

Also updated prod recon config based on most recent deployment.


deploy/gke/recon-prod.yaml
```diff
@@ -15,10 +15,10 @@
 # Production Recon Service GKE params
 
 project: datcom-recon-prod
-ip: 34.98.98.180
+ip: 34.117.28.86
 domain: recon.datacommons.org
 api_title: Data Commons Recon (Prod)
 api: datacommons.Recon
-region: us-central1
+region: us-west1
```
shifucun, line 22: There is still something issue with this deployment, and this config is for testing. Going to switch it back to us-central1 when the issue is solved.

beets (approved):
thanks for bringing more sanity to our stat protos :) 

just to confirm that these changes are internal and there's no change to api responses?

also, i don't understand the change in the recon test. is that more related to the config change? in general, it'd be best if these config updates were done in separate pr's for easy rollback, debugging, ...

shifucun:
> thanks for bringing more sanity to our stat protos :)
> 
> just to confirm that these changes are internal and there's no change to api responses?
> 
> also, i don't understand the change in the recon test. is that more related to the config change? in general, it'd be best if these config updates were done in separate pr's for easy rollback, debugging, ...

Yea, this does not change the response, as reflected in golden tests.

Looks like the recon tests result is not stable.. need to spend some time or ask Wei to look into that. For the config change, it's still WIP, this is to match the current deployment..

shifucun:
Thanks for the review!

internal/server/recon/golden/resolve_entities/result.json
```diff
@@ -1,43 +1,43 @@
 {
   "resolvedEntities": [
     {
-      "sourceId": "newId/SunnyvaleId",
+      "sourceId": "newId/VietnamId",
```
shifucun, line 4: @Spaceenter can you check if the response is stable or if we need to do some sorting here?

pradh: Looks it is sorted - https://github.com/datacommonsorg/mixer/blob/master/internal/server/recon/resolve.go#L209-L212 .. I think the new order is right?

PR #700, opened 2022-01-21
Support multiple base BT in mixer

shifucun:
Now only supports merging base tables for stat series/collection cache. All other cache only uses one base table.

This way, we don't need a flag to indicate the import group case, as it will work for just one group (the current base cache). And can gradually add merging capability for all the cache.


internal/store/bigtable/stat_reader.go
```diff
@@ -55,17 +55,27 @@ func ReadStats(
...
 	}
 	for _, rowKey := range rowList {
 		token, _ := keyToTokenFn(rowKey)
 		psv := keyTokens[rowKey]
-		if data, ok := branchDataMap[token]; ok {
-			result[psv.Place][psv.StatVar] = data.(*model.ObsTimeSeries)
-		} else if data, ok := baseDataMap[token]; ok {
-			result[psv.Place][psv.StatVar] = data.(*model.ObsTimeSeries)
+		// Different base data has different source series, concatenate them together.
+		ss := result[psv.Place][psv.StatVar].SourceSeries
```
pradh, line 65: So this assumes there must be no overlap?

pradh (approved):

internal/store/bigtable/stat_reader.go
shifucun: Yep, for the base import groups. I will move the merging logic into separate functions, and after we introduce the branch import group, will need to add the capability of picking from preferred group.

PR #701, opened 2022-01-21
[tree widget] hide orphan sv and svg

chejennifer:
- When creating search index and parent map, ignore orphan sv and svg (ones not attached to the root) so that they don't show up in search results


internal/server/statvar/statvar_hierarchy_util.go
```diff
@@ -61,21 +61,38 @@ func GetRawSvg(ctx context.Context, baseTable *cbt.Table) (
 	return svgResp.StatVarGroups, nil
 }
 
-// BuildParentSvgMap gets the mapping of svg/sv id to the parent svg for that svg/sv.
+// BuildParentSvgMap gets the mapping of svg/sv id to the parent svg for that
+// svg/sv. Only gets the parent svg that have a path to the root node.
 func BuildParentSvgMap(rawSvg map[string]*pb.StatVarGroupNode) map[string][]string {
 	parentSvgMap := map[string][]string{}
```
pradh, line 67: nit: ancestorSvgMap ?

pradh (approved):
Nice!

internal/server/statvar/statvar_hierarchy_util.go
chejennifer: This map only holds the direct parent svgs of each sv or svg, so I'm not sure that ancestorSvgMap will make more sense?

pradh: Ah, I see... sorry i misread it.

PR #702, opened 2022-01-21
Data Release: 2022_01_21_01_03_06

pradh:
There is some shift in the hate crime Stat Vars I think because the definition changed in cl/422282043, but vertical spec wasn't updated yet.


scripts/prepare_release.sh
```diff
@@ -49,7 +49,7 @@ function commit() {
   git commit -a -m "Data Release: $VERSION"
 }
 
-update_version
+# update_version
```
beets, line 52: revert?

beets (approved):

scripts/prepare_release.sh
pradh: update_version was broken when autopush.datacommons.org needed login :(.  Commenting it makes rest of the flow work.

pradh (commented):
Thanks!

PR #705, opened 2022-01-25
Add import group config and setup golden testing to use it

shifucun:
This is to allow golden test to take either <base, branch> cache or a set of base tables. A test can then produce two sets of golden files. When we are confident about the merging. We will compare the two set of golden files and switch to use the import group.

This PR updates /stat/set/series API golden test to incorporate result from import group as an example.


deploy/storage/bigtable.json
```diff
@@ -0,0 +1,7 @@
+{
```
pradh, line 1: nit: Should we call this bigtable_import_groups.json or something to distinguish it from the current one?

internal/server/stat/golden/get_stat_set_series/IG_nyt_covid_cases.json
```diff
@@ -0,0 +1,2228 @@
...
+            "2022-01-19": 245673,
+            "2022-01-20": 253711,
+            "2022-01-21": 257994,
+            "2022-01-22": 257994
+          },
+          "metadata": {
+            "importName": "NYT_COVID19",
+            "provenanceUrl": "https://www.nytimes.com/article/coronavirus-county-data-us.html",
+            "measurementMethod": "NYT_COVID19_GitHub"
+          }
```
pradh, line 2223: So this one works!

pradh (approved):
Nice!

internal/server/stat/golden/get_stat_set_series/IG_nyt_covid_cases.json
shifucun: Yep! Need to add "infrequent" and see for more.

PR #707, opened 2022-01-25
[Import Group] Add test for /stat/value and /stat/set

shifucun:
Updates the /stat/value response behavior so it returns empty value, instead of throwing error, when no data is found. This is to be consistent with other APIs.

Improved StatVarObservation filter field naming.


proto/stat.proto
```diff
@@ -213,6 +213,9 @@ message GetStatsRequest {
   // (Optional) The observation period of the observation. If not specified,
   // stats series with any observation period could be returned.
   string observation_period = 6;
+
```
pradh, line 216: This is user-visible change? Does this need any doc update?

pradh (approved):

proto/stat.proto
shifucun: Yes, this is additional field. I plan to go over the docs after this round of API change (remove payload)

shifucun:
Thanks!

PR #708, opened 2022-01-26
Add infrequent import group and update golden files

shifucun:
Merging looks good. The golden files between current and import group mode all match.


pradh (approved):
Very nice!

PR #710, opened 2022-01-26
Add branch cache to version and fetch storage data from server metadata

shifucun:
Instead of fetching version from environment, save them in server metadata and use from that.

Turns out branch cache was disabled in a previous PR. Fixed that here too.


proto/misc.proto
```diff
@@ -72,8 +72,12 @@ message GetVersionResponse {
   string store = 1;
   // Big Query dataset
   string big_query = 2;
-  // Bigtable instance
-  string big_table = 3;
+  // Base Bigtable instance
```
pradh, line 75: needs an aligned website change?

cmd/main.go
```diff
@@ -53,7 +52,7 @@ var (
 	useBaseBt     = flag.Bool("use_base_bt", true, "Use base bigtable cache")
 	baseTableName = flag.String("base_table", "", "Base cache Bigtable table.")
 	// Branch Bigtable Cache
-	useBranchBt = flag.Bool("use_branch_bt", false, "Use branch bigtable cache")
+	useBranchBt = flag.Bool("use_branch_bt", true, "Use branch bigtable cache")
```
pradh, line 55: how do we catch such regressions?

pradh (approved):

proto/misc.proto
shifucun: Yes, website will use this API for versioning too.

cmd/main.go
shifucun: We should have a prober to fetch the few stat vars from branch cache and make sure the latest date is no more than N days.

PR #714, opened 2022-01-28
[Import Group] Merge for property value cache

shifucun:
Nodes are read from the most preferred group first. When the same node is seen in less preferred group, only add it if it has not been added. 

Note out arcs only get from preferred import group. In arcs are merged across import groups.

### DIFF
--- a/internal/server/node/golden/get_property_values_test.go
+++ b/internal/server/node/golden/get_property_values_test.go
@@ -29,103 +29,112 @@ import (
 )
 
 func TestGetPropertyValues(t *testing.T) {
-	ctx := context.Background()
-	client, _, err := e2e.Setup()
-	if err != nil {
-		t.Fatalf("Failed to set up mixer and client")
-	}
-	_, filename, _, _ := runtime.Caller(0)
-	goldenPath := path.Join(
-		path.Dir(filename), "get_property_values")
-
 	t.Parallel()
-	for _, c := range []struct {
-		goldenFile string
-		dcids      []string
-		property   string
-		direction  string
-		valueType  string
-		limit      int32
-	}{
-		{
-			"name.json",
-			[]string{"State", "geoId/05", "Count_Person", "dc/p/cmtdk79lnk2pd"},
-			"name",
-			"out",
-			"",
-			0,
-		},
-		{
-			"contained_in_place.json",
-			[]string{"geoId/06085", "geoId/0647766"},
-			"containedInPlace",
-			"",
-			"City",
-			0,
-		},
-		{
-			"contained_in_place_all.json",
-			[]string{"geoId/06085", "geoId/0647766"},
-			"containedInPlace",
-			"out",
-			"",
-			0,
-		},
-		{
-			"location.json",
-			[]string{"geoId/05", "geoId/06"},
-			"location",
-			"",
-			"Election",
-			0,
-		},
-		{
-			"limit.json",
-			[]string{"country/USA"},
-			"name",
-			"out",
-			"",
-			1,
-		},
+	ctx := context.Background()
+
+	for _, opt := range []*e2e.TestOption{
+		{},
+		{UseImportGroup: true},
 	} {
-		req := &pb.GetPropertyValuesRequest{
-			Dcids:     c.dcids,
-			Property:  c.property,
-			Direction: c.direction,
-			ValueType: c.valueType,
-		}
-		if c.limit > 0 {
-			req.Limit = c.limit
-		}
-		resp, err := client.GetPropertyValues(ctx, req)
+		client, _, err := e2e.Setup(opt)
 		if err != nil {
-			t.Errorf("could not GetPropertyValues: %s", err)
-			continue
+			t.Fatalf("Failed to set up mixer and client")
 		}
-		goldenFile := path.Join(goldenPath, c.goldenFile)
+		_, filename, _, _ := runtime.Caller(0)
+		goldenPath := path.Join(
+			path.Dir(filename), "get_property_values")
 
-		var result map[string]map[string][]*model.Node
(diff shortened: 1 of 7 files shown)


internal/server/node/property_value.go
```diff
@@ -108,12 +113,57 @@ func GetPropertyValuesHelper(
 	arcOut bool,
 ) (map[string][]*model.Node, error) {
 	rowList := bigtable.BuildPropertyValuesKey(dcids, prop, arcOut)
-	// Current branch cache is targeted on new stats (without addition of schema etc),
-	// so only use base cache data for property value.
-	//
-	// TODO(shifucun): perform a systematic check on current cache data and see
-	// if this is still true.
-	return readPropertyValues(ctx, store, rowList)
+	baseDataList, _, err := bigtable.Read(
```
pradh, line 116: nit: Can still use the helper partially?  Smaller funcs more readable, etc :)

internal/server/node/property_value.go
```diff
@@ -108,12 +113,57 @@ func GetPropertyValuesHelper(
...
+		},
+		nil,
+		true, /* readBranch */
+	)
+	if err != nil {
+		return nil, err
+	}
+	result := map[string][]*model.Node{}
+	visited := map[string]map[string]struct{}{}
+	// Loop the import groups. They are ordered by the preferences. So only add
```
pradh, line 136: nit: Loop over

pradh, line 136: sorry, could you point at where they are ordered?

pradh, line 136: Perhaps clarify more like ... "for every entity we get values for a property from exactly one cache.  The specific cache we get from is based on a preference list..."

internal/server/node/golden/get_property_values_test.go
```diff
@@ -29,103 +29,112 @@ import (
...
-		if e2e.GenerateGolden {
-			e2e.UpdateGolden(result, goldenPath, c.goldenFile)
-			continue
-		}
+			var result map[string]map[string][]*model.Node
+			err = json.Unmarshal([]byte(resp.GetPayload()), &result)
+			if err != nil {
+				t.Errorf("Can not Unmarshal payload")
+				continue
+			}
```
Spaceenter, line 120: if err = json.Unmarshal([]byte(resp.GetPayload()), &result); err != nil {
    ....
}

internal/server/node/property_value.go
```diff
@@ -108,12 +113,57 @@ func GetPropertyValuesHelper(
...
+				nodes := data.([]*model.Node)
+				for _, n := range nodes {
+					// Check if a duplicate node has been added to the result.
+					// Duplication is based on either the DCID or the value.
+					if n.Dcid != "" {
+						if _, ok := visited[dcid][n.Dcid]; !ok {
+							result[dcid] = append(result[dcid], n)
+							visited[dcid][n.Dcid] = struct{}{}
+						}
+					} else if n.Value != "" {
```
pradh, line 156: To confirm, are we combining values?  I thought we would pick all values of a property from the preferred cache..

internal/server/node/property_value.go
pradh: Ok I realize that understanding may not be right here...

internal/server/node/property_value.go
```diff
@@ -108,12 +113,57 @@ func GetPropertyValuesHelper(
...
+	baseDataList, _, err := bigtable.Read(
+		ctx,
+		store.BtGroup,
+		rowList,
+		func(dcid string, jsonRaw []byte) (interface{}, error) {
+			var propVals model.PropValueCache
+			err := json.Unmarshal(jsonRaw, &propVals)
+			if err != nil {
+				return nil, err
+			}
```
Spaceenter, line 125: nit: 
```
if err := json.Unmarshal(jsonRaw, &propVals); err != nil {
  ...
}
```

pradh:
How do the diffs look relative to prod?

internal/server/node/property_value.go
shifucun: This function just had 2 lines before.. and the helper is only used here. So might just merge?

internal/server/node/property_value.go
shifucun: Done

internal/server/node/property_value.go
```diff
@@ -108,12 +113,57 @@ func GetPropertyValuesHelper(
...
+	for _, baseData := range baseDataList {
+		for dcid, data := range baseData {
+			if _, ok := result[dcid]; !ok {
+				result[dcid] = []*model.Node{}
+			}
+			if _, ok := visited[dcid]; !ok {
+				visited[dcid] = map[string]struct{}{}
+			}
+			if data != nil {
+				nodes := data.([]*model.Node)
```
Spaceenter, line 147: Golang type assertions should be checked, in case the type conversion failed. Something like:

```
nodes, ok := data.([]*model.Node)
if !ok {
  return some error messages
}
```

Same comments for other Golang conversions, but those could be fixed a separate PR.

internal/server/node/property_value.go
shifucun: Order is in bigtable_import_groups.json now. When import groups are finalized, should need ordering when creating the table clients.

internal/server/node/property_value.go
shifucun: Yes, the loop starts from the preferred cache. New node is only added when it is not in visited.

Spaceenter (approved):

shifucun:
> How do the diffs look relative to prod?

All golden files match exactly.

pradh (approved):
Thanks for the change!

pradh:
> > How do the diffs look relative to prod?
> 
> All golden files match exactly.

nice!

PR #715, opened 2022-01-28
[Import Group] Merge place page cache

shifucun:
All golden response matches except for the recent days of covid data


internal/server/stat/stat_util.go
```diff
@@ -248,3 +249,35 @@ func getMetadataHash(m *pb.StatMetadata) uint32 {
...
+		"%s_%s_%s_%s_%s",
+		s.ImportName,
+		s.MeasurementMethod,
+		s.ObservationPeriod,
+		s.ScalingFactor,
+		s.Unit)
+}
+
+// MergeSourceSeries merges lists of SourceSeries. For same source series, the one
+// with the latest data will be used.
```
pradh, line 266: MergeSourceSeries gives the sense that a single list gets merged... perhaps worth noting that never happens.

Wonder if the function should be named like CollectDistinctSourceSeries(), or similar...

internal/server/stat/stat_util.go
```diff
@@ -248,3 +249,35 @@ func getMetadataHash(m *pb.StatMetadata) uint32 {
...
+func MergeSourceSeries(seriesList ...[]*pb.SourceSeries) []*pb.SourceSeries {
+	result := []*pb.SourceSeries{}
+	resultMap := map[string]*pb.SourceSeries{}
+	for _, series := range seriesList {
+		for _, s := range series {
+			key := getSourceSeriesKey(s)
+			if _, ok := resultMap[key]; ok && len(s.Val) < len(resultMap[key].Val) {
```
pradh, line 273: nit: The comment above says latest, but we are checking longest?   Clarify in the comment above that for a given series, longest series often has the latest data too...

pradh (approved):

internal/server/stat/stat_util.go
shifucun: Done

PR #717, opened 2022-01-31
[Import Group] Stat Var related API clean up

shifucun:
In general, there is no need to merge for stat var related API, while the preferred cache table should be used. This PR

- Clean up logic throughout stat var APIs by using bigtable.Read() function
- Remove the condition to build stat var hierarchy using triples. That's only used during hierarch dev stage.
- Split functions into separate folders for readability.
- Restrict some function parameter scope from "store" to "bigtable group".


internal/server/statvar/statvar_group.go
```diff
@@ -169,61 +173,18 @@ func GetStatVarGroupNode(
 	}
 
 	result := &pb.StatVarGroupNode{}
-
-	if in.GetReadFromTriples() {
```
pradh, line 173: is GetReadFromTriples being deprecated?

pradh (approved):
Nice cleanup!  (though might be easier to review when function moves across files are in their own logic-less PR)

internal/server/statvar/statvar_group.go
shifucun: Yep, this is not used anywhere now.

shifucun:
Thanks for review!

PR #718, opened 2022-02-01
Skip proto branch cache update

shifucun:


internal/server/server.go
```diff
@@ -137,6 +137,10 @@ func (s *Server) SubscribeBranchCacheUpdate(
 		func(ctx context.Context, msg *pubsub.Message) error {
 			branchTableName := string(msg.Data)
 			log.Printf("branch cache subscriber message received with table name: %s\n", branchTableName)
+			if v, ok := msg.Attributes["serialization"]; ok && v == "proto" {
```
pradh, line 140: Just to understand, what will the behavior be if we got this wrong?

pradh (approved):

internal/server/server.go
shifucun: This branch cache table will be used

shifucun: Will not crash, but failed with reading empty data..

PR #719, opened 2022-02-01
[stat var hierarchy] group search results

chejennifer:
- group stat var search results by stat var group: If a stat var matches the tokens and has a parent stat var group that also matches the tokens, nest the stat var result under the stat var group result.


proto/stat_var.proto
```diff
@@ -181,10 +181,19 @@ message SearchStatVarRequest {
   bool enable_blocklist = 3;
 }
 message SearchStatVarResponse {
```
shifucun, line 183: Nit: can move SearchResultSVG one level up, so it's easier to reference in the code

internal/server/statvar/statvar_search.go
```diff
@@ -105,6 +124,43 @@ func filter(
...
+		}
+	}
+	resultSv := []*pb.EntityInfo{}
+	// Iterate through the list of stat vars. If a stat var has a parent svg that
+	// is a result, add the stat var to that svg result. If not, add the stat var
+	// to the result list of stat vars
+	for _, sv := range svList {
+		isGrouped := false
+		if parents, ok := parentMap[sv.Dcid]; ok {
+			for _, parent := range parents {
```
shifucun, line 144: I wonder should sort the parents here so stat var is always added to the top parent.

shifucun (approved):
Very nice change! The response is much better read now.

proto/stat_var.proto
chejennifer: done

internal/server/statvar/statvar_search.go
chejennifer: I think that's a good idea, done

beets (approved):
this is nice!

chejennifer:
thanks!

PR #720, opened 2022-02-01
[Import Group] Update test for bio page cache; Cleanup bigtable related code

shifucun:


internal/server/biopage/bio_page.go
```diff
@@ -55,8 +55,10 @@ func GetBioPageData(
 	if err != nil {
 		return nil, err
 	}
-	if _, ok := data[0][dcid]; !ok {
-		return nil, nil
+	for _, baseData := range baseDataList {
```
pradh, line 58: Perhaps add a comment that there is no merging going on here, and order is based on the preferred base-cache list?

pradh (approved):

internal/server/biopage/bio_page.go
shifucun: Done

PR #721, opened 2022-02-03
Add golden test for get_stat_set_series_within_place for import group.

Spaceenter:
IG_county_city_population.json is exactly the same as county_city_population.json, compared with vimdiff.


shifucun (approved):
Thanks for the change!

PR #722, opened 2022-02-03
update golden tests for import group for get_stat{_all, _series, s}

Spaceenter:
Checked with "vimdiff IG_xxx.json xxx.json".

The only diff found is us_covid, which has a few more days' data, as expected.


shifucun (approved):
nice! Thanks!

PR #725, opened 2022-02-04
clang-format proto files

shifucun:


pradh (approved):
Nice cleanup!

For future, might be nice to check clang-format on presubmit.

shifucun:
> Nice cleanup!
> 
> For future, might be nice to check clang-format on presubmit.

Yea, need a bunch of cloudbuild change. Can do after import group launch.

PR #727, opened 2022-02-04
Use server cache for /stat-var/group/all API

shifucun:
The old implementation read the data from BT for each request. As the raw data is already loaded once in server start, can directly use it.

Here modify the filterSVG function to return a new copy instead of modifying in place.


pradh (approved):
Nice!  Thanks for the quick change!  Eager to see how this does...

PR #728, opened 2022-02-04
Update import group tables to the latest build

shifucun:


pradh (approved):
Nice, this fixes the dcAggregate/

Issue #661, opened 2021-11-04
Investigating GKE memory and cpu resource limit setup


shifucun:
Currently we set container cpu and memory, but have been bitten a few times for requests that takes up cpu and memory. It turns out large request like /stat/set/within-place of <Europe, IPCCPlace_50> uses peak CPU of 3000m and get throttled. Following are some read about not setting limit:

https://erickhun.com/posts/kubernetes-faster-services-no-cpu-limits/
https://medium.com/omio-engineering/cpu-limits-and-aggressive-throttling-in-kubernetes-c5b20bd8a718

We should try to remove the resource limit and see if it works.

shifucun:
@pradh @beets please take a look at the articles and see if they make sense

PR #729, opened 2022-02-05
Update resource constraint: more memory for esp; remove cpu limit

shifucun:
A few articles have suggested to remove cpu limit (https://github.com/datacommonsorg/mixer/issues/661). Basically even the cpu is far below limit, it can still get throttled and result in poor performance.

We have seen issues with "/stat-var/group/all" endpoints not returning anything, as the size grows. I tested with `curl http://localhost:8081/stat-var/group/all --output svg.json`. The observation is as follow:

1) The data is read from server cache, so mixer has nothing to read or process.
2) When esp has memory limit of 1G as we have now, it won't return the result, giving error "server didn't response".
3) When esp has memory limit of 2G, it works fine, with the response json size of 226M. As esp has to convert protobuf to JSON and send it out, so 1G of memory could be too tight.

### DIFF
--- a/deploy/base/deployment.yaml
+++ b/deploy/base/deployment.yaml
@@ -51,10 +51,8 @@ spec:
           resources:
             limits:
               memory: "6G"
-              cpu: "800m"
             requests:
               memory: "6G"
-              cpu: "800m"
           args:
             - --mixer_project=$(MIXER_PROJECT)
             - --store_project=$(STORE_PROJECT)
@@ -122,11 +120,9 @@ spec:
                   key: serviceName
           resources:
             limits:
-              memory: "1G"
-              cpu: "200m"
+              memory: "2G"
             requests:
-              memory: "1G"
-              cpu: "200m"
+              memory: "2G"
           readinessProbe:
             httpGet:
               path: /healthz
--- a/deploy/overlays/local/patch_deployment.yaml
+++ b/deploy/overlays/local/patch_deployment.yaml
@@ -29,24 +29,10 @@ spec:
         - name: mixer
           image: datacommons/mixer
           imagePullPolicy: Never
-          resources:
-            limits:
-              memory: "5G"
-              cpu: "1000m"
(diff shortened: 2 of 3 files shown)


pradh:
> A few articles have suggested to remove cpu limit (#661). Basically even the cpu is far below limit, it can still get throttled and result in poor performance.
> 
> We have seen issues with "/stat-var/group/all" endpoints not returning anything, as the size grows. I tested with `curl http://localhost:8081/stat-var/group/all --output svg.json`. The observation is as follow:
> 
> 1. The data is read from server cache, so mixer has nothing to read or process.
> 2. When esp has memory limit of 1G as we have now, it won't return the result, giving error "server didn't response".
> 3. When esp has memory limit of 2G, it works fine, with the response json size of 226M. As esp has to convert protobuf to JSON and send it out, so 1G of memory could be too tight.

Nice!  Any tips on how this looks like in the ESP logs, or something to identify it in the future?

pradh (approved):

shifucun:
> > A few articles have suggested to remove cpu limit (#661). Basically even the cpu is far below limit, it can still get throttled and result in poor performance.
> > We have seen issues with "/stat-var/group/all" endpoints not returning anything, as the size grows. I tested with `curl http://localhost:8081/stat-var/group/all --output svg.json`. The observation is as follow:
> > 
> > 1. The data is read from server cache, so mixer has nothing to read or process.
> > 2. When esp has memory limit of 1G as we have now, it won't return the result, giving error "server didn't response".
> > 3. When esp has memory limit of 2G, it works fine, with the response json size of 226M. As esp has to convert protobuf to JSON and send it out, so 1G of memory could be too tight.
> 
> Nice! Any tips on how this looks like in the ESP logs, or something to identify it in the future?

I don't see anything from the ESP log. Also not sure in this case if the esp can log anything. Maybe need to learn how to find corresponding GKE events or logs at pod level..

PR #730, opened 2022-02-06
[Import Group] Merge triples cache

shifucun:
Due to the nested proto structure, it's not possible to use a protobuf for response and make it backward compatible. A lot of the change has been on data converting in this PR.

The diff from current and import group mode matches reasonably. This changes added some fields that are missing in the current cache, like "subjectTypes" "subjectName" etc.

Applied sorting to both old and new golden files so they can be compared


internal/server/handler.go
```diff
@@ -219,12 +220,13 @@ func (s *Server) GetPropertyLabels(
 		return nil, err
 	}
 	jsonRaw, err := protojson.Marshal(resp)
-	// To make the API response backward compatible. This is to add the outer
-	// level `{\"data\":`
-	jsonRaw = jsonRaw[8 : len(jsonRaw)-1]
 	if err != nil {
 		return nil, err
 	}
+	// To make the API response backward compatible. This is to add the outer
```
Spaceenter, line 227: s/add/remove ?

internal/server/node/convert.go
```diff
@@ -0,0 +1,138 @@
...
+	"sort"
+
+	pb "github.com/datacommonsorg/mixer/internal/proto"
+	"github.com/datacommonsorg/mixer/internal/server/model"
+)
+
+// Convert pb.GetTriplesResponse to legacy golang struct to make the API response
+// backward compatible.
+func ToLegacyResult(in *pb.GetTriplesResponse) map[string][]*model.Triple {
+	out := map[string][]*model.Triple{}
```
Spaceenter: nit: "out" might confuse the readers that this means "outEntity", so maybe change the variable name to "result"?

internal/server/node/golden/get_triples_test.go
```diff
@@ -31,107 +31,115 @@ import (
...
+			}
+			resp, err := client.GetTriples(ctx, req)
+			if err != nil {
+				t.Errorf("could not GetTriples: %s", err)
+				continue
+			}
+			var result map[string][]*model.Triple
+			err = json.Unmarshal([]byte(resp.GetPayload()), &result)
+			if err != nil {
+				t.Errorf("Can not Unmarshal payload")
```
Spaceenter, line 110: t.Errorf("Can not Unmarshal payload: %s", err)

internal/server/node/golden/get_triples_test.go
```diff
@@ -31,107 +31,115 @@ import (
...
+			if c.limit > 0 {
+				req.Limit = c.limit
+			}
+			resp, err := client.GetTriples(ctx, req)
+			if err != nil {
+				t.Errorf("could not GetTriples: %s", err)
+				continue
+			}
+			var result map[string][]*model.Triple
+			err = json.Unmarshal([]byte(resp.GetPayload()), &result)
```
Spaceenter, line 108: if err := json.Unmarshal(...); err != nil {

Spaceenter (approved):

internal/server/handler.go
```diff
@@ -219,12 +221,13 @@ func (s *Server) GetPropertyLabels(
...
 	jsonRaw, err := protojson.Marshal(resp)
-	// To make the API response backward compatible. This is to add the outer
-	// level `{\"data\":`
-	jsonRaw = jsonRaw[8 : len(jsonRaw)-1]
 	if err != nil {
 		return nil, err
 	}
+	// To make the API response backward compatible. This is to remove the outer
+	// level `{\"data\":`
+	jsonRaw = jsonRaw[8 : len(jsonRaw)-1]
```
pradh, line 229: nit: just wondering, do we not need to skip the trailing "}" ?

pradh (approved):

internal/server/handler.go
shifucun: Oh, it's doing that. Added in the comments

internal/server/node/convert.go
shifucun: Done

shifucun:
Thanks for review!

PR #731, opened 2022-02-07
Put branch cache table into table group for consistent logic handling

shifucun:
Current server saves an explicit branch table object and the bigtable.Read() function returns data for branch table explicitly. As now we have merging for all the cache with import group changes, branch cache table is no special at all. 

This PR put the branch cache table (even in non import-group mode) into the table group and let the general table merge logic to handle it. In fact, in import group mode, the branch cache is jus a regular group in all the tables.


test/e2e/setup.go
```diff
@@ -105,29 +106,29 @@ func setupInternal(
...
+		if err != nil {
+			return nil, nil, err
+		}
+		tables = append(tables, bigtable.NewTable(name, baseTable))
+		branchTable, err := createBranchTable(ctx)
 		if err != nil {
 			return nil, nil, err
 		}
-		baseTables = append(baseTables, baseTable)
+		tables = append(tables, bigtable.NewTable(branchTableName, branchTable))
```
pradh, line 131: Could you maybe add a few lines to the description that clarify why this change is safe in current prod?

internal/store/bigtable/group.go
```diff
@@ -34,64 +34,104 @@ var groupRank = map[string]int{
...
+	tables := []*Table{}
+	for _, t := range g.tables {
+		if t.name != g.branchTableName {
+			tables = append(tables, t)
+		}
+	}
+	tables = append(tables, branchTable)
+	g.branchTableName = branchTable.name
+	g.tables = tables
+	if g.isProto {
```
pradh, line 116: else, is the assertion that branch cache always come after base?  Wonder if it's worth asserting?

pradh (approved):

internal/store/bigtable/group.go
shifucun: Done

test/e2e/setup.go
shifucun: Done

PR #733, opened 2022-02-10
[Import Group] Update merging for stat var related APIs

shifucun:
There is some data issue with StatVarSummary cache. Some City places are treated as AA or Place. Will comeback to update those when Prophet issue fixed.


Spaceenter (approved):

proto/stat_var.proto
```diff
@@ -126,11 +127,12 @@ message StatVarGroupNode {
     // Name suitable for search.
     string search_name = 2;
     // Names suitable for search.
-    repeated string search_names = 5;
+    repeated string search_names = 4;
     // Name suitable for display in tree.
     string display_name = 3;
+    //------------------ Below are fields not in original cache.
```
pradh, line 133: ... and thus we start with a large tag number

proto/stat_var.proto
```diff
@@ -34,20 +34,20 @@ import "entity.proto";
...
+  int64 num_descendent_stat_vars = 1;
 }
 
 message StatVarSummary {
   message Place {
     string dcid = 1;
-    string name = 2;
-    int64 person_count = 3;
+    string name = 3;
+    int64 person_count = 2;
```
pradh, line 44: This is dropped now?

pradh: https://source.corp.google.com/piper///depot/google3/datacommons/proto/prophet/cache_data.proto;dlc=426232895;bpv=0;bpt=0;rcl=427571961

internal/server/statvar/statvar_summary.go
```diff
@@ -56,8 +58,33 @@ func GetStatVarSummary(
...
+			for pt := range svs.PlaceTypeSummary {
+				summary, ok := result.StatVarSummary[dcid].PlaceTypeSummary[pt]
+				if ok && svs.PlaceTypeSummary[pt].NumPlaces < summary.NumPlaces {
+					continue
+				}
+				result.StatVarSummary[dcid].PlaceTypeSummary[pt] = svs.PlaceTypeSummary[pt]
+			}
+			//
+			for source := range svs.ProvenanceSummary {
+				result.StatVarSummary[dcid].ProvenanceSummary[source] = svs.ProvenanceSummary[source]
```
pradh, line 85: if there is an overlapping source, maybe we could prefer the one from the better cache?

pradh (commented):
Sorry for the delay!

internal/server/statvar/count.go
```diff
@@ -61,17 +62,29 @@ func Count(
...
 	for _, rowKey := range rowList {
 		placeSv := keyTokens[rowKey]
 		token, _ := keyToTokenFn(rowKey)
-		if data, ok := baseDataList[0][token]; ok {
-			c := data.(*pb.PlaceStatVarExistence)
-			result[placeSv.StatVar][placeSv.Place] = c.NumDescendentStatVars
+		for _, baseData := range baseDataList {
+			if data, ok := baseData[token]; ok {
+				c := data.(*pb.PlaceStatVarExistence)
+				if _, ok := result[placeSv.StatVar][placeSv.Place]; !ok {
```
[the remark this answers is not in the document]
shifucun: result is initialized in line 65-67, so this should be okay.

internal/server/statvar/count.go
```diff
@@ -61,17 +62,29 @@ func Count(
...
+					// This is for the case where c.NumDescendentStatVars = 0.
+					// In this case, placeSv.StatVar is a stat var (not a stat var group)
+					// so the number of descendent is 0.
+					// If not performance this check but only the one below, the the
```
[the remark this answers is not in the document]
shifucun: Done

internal/server/statvar/statvar_summary.go
shifucun: Done

proto/stat_var.proto
shifucun: I will remove it when switch to the new cache, otherwise the Unmarshal would fail now.

proto/stat_var.proto
shifucun: Done

pradh (approved):
Thanks for the changes!

PR #735, opened 2022-02-11
Merge for GetPlaceMetadata Cache

shifucun:
Since place data is in every import group. This picks the most preferred group and only use non-preferred group if it has newer place (in theory this should not happen).

### DIFF
--- a/internal/server/place/places.go
+++ b/internal/server/place/places.go
@@ -331,40 +331,45 @@ func GetPlaceMetadata(
 	}
 	result := map[string]*pb.PlaceMetadata{}
 	for _, place := range places {
-		if baseDataList[0][place] == nil {
+		if _, ok := result[place]; ok {
 			continue
 		}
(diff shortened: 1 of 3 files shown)


Spaceenter (approved):

internal/server/place/places.go
```diff
@@ -331,40 +331,45 @@ func GetPlaceMetadata(
...
 				continue
 			}
-			processed.Parents = append(processed.Parents, &pb.PlaceMetadata_PlaceInfo{
-				Dcid: curr,
-				Name: metaMap[curr].Name,
-				Type: metaMap[curr].Type,
-			})
-			visited[curr] = struct{}{}
-			parents = append(parents, metaMap[curr].Parents...)
+			raw := baseData[place].(*pb.PlaceMetadataCache)
```
Spaceenter, line 341: nit:

raw, ok := baseData[place].(*pb.PlaceMetadataCache)
if !ok {
...
}

shifucun: Done

PR #738, opened 2022-02-11
[stat var hierarchy] return the matching strings in stat var search

chejennifer:
- When returning the results for /stat-var/search, for each result, return the strings that match the searched for tokens. This information will be used to highlight the parts of the results that match the search tokens.


proto/stat_var.proto
```diff
@@ -173,13 +171,20 @@ message GetStatVarPathResponse {
...
-  // DCID of the stat var group
+
+message SearchResultEntry {
+  // DCID of the result
   string dcid = 1;
-  // Name of the stat var group
+  // Name of the result
   string name = 2;
+  // The strings in the name that match the search tokens
+  repeated string matches = 3;
```
beets, line 181: instead of embedding this with each result, can we achieve the same (with a much smaller response) by returning a set of matching words across all results at the SearchStatVarResponse level? For most searches, this set would stay fairly small, linear with the size of the search string (not very long) rather than the search results (usually very long).

In both cases, we would still need to do string matching to highlight the matched terms so the client would be handling the information the same way. wdyt?

beets (commented):
sorry for the late review!

proto/stat_var.proto
shifucun: Agree with this.

Also the search index now is getting bigger too with each token have multiple linked text. I have seen issues with the search index blowing up the mixer memory. Do you think there is also a way to make the record here smaller?

chejennifer: Actually, yes, that makes sense. My original thinking was that because of synonyms, not all results will have the same set of matching words. But now that I think about it, since the client side is doing string matching, the client side can figure out which of the words in the set actually match each result.

proto/stat_var.proto
```diff
@@ -208,9 +204,11 @@ message SearchStatVarRequest {
 message SearchStatVarResponse {
   // A ranked list of stat vars that aren't included in any of the returned stat
   // var groups.
-  repeated SearchResultEntry stat_vars = 1;
+  repeated EntityInfo stat_vars = 1;
   // A list of stat var group results ranked by relevance.
   repeated SearchResultSVG stat_var_groups = 2;
+  // A list of strings in the names of the results that match the search tokens
```
beets, line 210: nit: clarify that these are unique strings?

internal/server/statvar/statvar_search.go
```diff
@@ -234,15 +238,18 @@ func searchTokens(
 		}
 	}
 
+	matchingStrings := map[string]struct{}{}
```
beets, line 241: since this isn't entirely straightforward, worth a comment that this is a set, and that struct{}{} is an empty struct.

you could also use `var exists = struct{}{}` to help readability

https://www.davidkaya.com/sets-in-golang/

beets (approved):
thanks for the update!

proto/stat_var.proto
chejennifer: done

internal/server/statvar/statvar_search.go
chejennifer: done

PR #739, opened 2022-02-15
Use preferred group for recon APIs

shifucun:
Recon APIs will not merge across import groups but would pick the data from a preferred group if it exists.

Also branch cache is not used.


internal/server/recon/coordinate.go
```diff
@@ -87,92 +89,72 @@ func ResolveCoordinates(
...
-			return &info, nil
-		},
-
-		func(rowKey string) (string, error) {
-			l := strings.TrimPrefix(rowKey, bigtable.BtOutPropValPrefix)
-			return strings.TrimSuffix(l, fmt.Sprintf("^%s", geoJSONPredicate)), nil
-		},
-		false,
-	)
+	geoJSONData, err := node.GetPropertyValuesHelper(ctx, store, questionablePlaceList, geoJSONPredicate, true)
```
shifucun, line 110: @Spaceenter here changed to use GetPropertyValuesHelper() function

Spaceenter (approved):
Please leave TODOs in id.go and resolve.go to read from all import groups, and merge the lists of IDs (instead of reading only from the preferred import group).

shifucun:
> Please leave TODOs in id.go and resolve.go to read from all import groups, and merge the lists of IDs (instead of reading only from the preferred import group).

Done

PR #740, opened 2022-02-15
Clean up places in API logic

shifucun:
- Remove a duplicate GetChildPlaces function
- Move the response formatting to handler.go, so GetPlacesIn can be reused in other modules.


internal/server/handler.go
```diff
@@ -123,7 +124,29 @@ func (s *Server) GetStatSetSeriesWithinPlace(
 // GetPlacesIn implements API for Mixer.GetPlacesIn.
 func (s *Server) GetPlacesIn(ctx context.Context, in *pb.GetPlacesInRequest,
 ) (*pb.GetPlacesInResponse, error) {
-	return place.GetPlacesIn(ctx, in, s.store)
+	// Current places in response is node idea.
+	// In next version of API, should simply return placesInData in the final response.
```
Spaceenter, line 128: This seems to be a TODO, maybe add "TODO" explicitly here?

Spaceenter (approved):

internal/server/handler.go
shifucun: Done

PR #743, opened 2022-02-17
[stat var hierarchy] various updates

chejennifer:
- add more synonyms, including phrases
- handle search when there are synonyms that are phrases
- reduce size of search index by making each TrieNode hold a set of matching strings instead of having each sv/svg in each TrieNode be mapped to their matching string for the token ending at that TrieNode
- fix flaky TestGroupStatVars test


internal/server/resource/resource.go
```diff
@@ -55,20 +55,13 @@ type SearchIndex struct {
...
-	// }
-	SvgIds map[string]string
-	SvIds  map[string]string
+	// SvgIds and SvIds are sets where Ids are keys and each key is mapped to an
+	// empty struct
+	SvgIds map[string]struct{}
+	SvIds  map[string]struct{}
+	// Matches is a set of strings that match the token ending at the current
+	// trienode
+	Matches map[string]struct{}
```
shifucun, line 64: this is nice!

internal/server/statvar/statvar_search.go
```diff
@@ -187,14 +175,27 @@ func groupStatVars(
...
 }
 
 func searchTokens(
 	tokens []string, index *resource.SearchIndex,
 ) ([]*pb.EntityInfo, []*pb.EntityInfo, []string) {
-	svMatches := map[string][]string{}
-	svgMatches := map[string][]string{}
+	// svMatches and svgMatches are maps of sv/svg id to a map of tokens that
+	// match each sv/svg to a set of strings that match the token
+	svMatches := map[string]map[string]map[string]struct{}{}
```
shifucun, line 197: maybe have a new type def for at lease 1/2 of the map here, so it is easy to see what's the map is. That also simplify the init statements below.

shifucun (commented):
Very nice change!

internal/server/statvar/statvar_search.go
beets: +1, it's also worth to do a typedef for empty = struct{}{} for readability

beets (approved):
this is a nice change!

internal/server/statvar/statvar_search.go
chejennifer: added new type def for part of the map.

but not quite understanding where the empty = struct{}{} will be used? Because I'm not setting anything to struct{}{}

shifucun (approved):

internal/server/statvar/statvar_search.go
beets: ah ok, i was misreading that statement. i thought it was an assignment

PR #744, opened 2022-02-18
Add import group golden tests for get_stat_set_within_place{_all}

Spaceenter:
Compared with:

vimdiff IG_xxx.json xxx.json

And no diffs are found for all newly generated json files.


shifucun (approved):
this is great!

PR #750, opened 2022-02-23
Various fixes for import group migration

shifucun:
- Fix get_triples bug when object is a string not entity. Should not check types when it's not entity.
- Update import group tables to incorporate fix for population side table in branch cache
- Update memdb test


internal/server/statvar/golden/get_statvar_group/IG_earth.json
```diff
@@ -2778,10 +2778,6 @@
         {
           "id": "dc/g/Consumption",
           "specializedEntity": "Consumption"
-        },
```
shifucun, line 2781: Is this a real data change?

pradh (approved):

PR #753, opened 2022-02-25
Add import group option to the test for  /place/stat/date/within-place

beets:


internal/server/place/golden/get_place_stat_date_within_place_test.go
```diff
@@ -27,57 +27,65 @@ import (
...
+	for _, opt := range []*e2e.TestOption{
+		{UseCache: true, UseMemdb: true},
+		{UseCache: true, UseMemdb: true, UseImportGroup: true},
 	} {
-		resp, err := client.GetPlaceStatDateWithinPlace(ctx, &pb.GetPlaceStatDateWithinPlaceRequest{
-			AncestorPlace: c.ancestorPlace,
-			PlaceType:     c.placeType,
-			StatVars:      c.statVars,
-		})
+		client, _, err := e2e.Setup()
```
shifucun, line 36: need to pass in opt here "e2e.Setup(opt)"

shifucun (approved):
This is missed! Thanks for catching up and the fix! Please rerun with the "opt" passed in.

internal/server/place/golden/get_place_stat_date_within_place_test.go
beets: thanks for the catch! fixed. oddly, no change in the results ...

shifucun: That's good, so there are no diff between current and IG!

PR #742, opened 2022-02-16
Update tables and a bunch of fixes

shifucun:
This updates all the base/branch and import group tables. This incorporates latest proto field change from int64->double.

Other notable fixes here:

* Use real branch cache in golden tests.
* In regular base + branch table mode, should also prefer branch table over base table.
* Fix a few merging bug when data is not present in preferred group.

PR #754, opened 2022-02-25
Fix partially cleaned up PlacesIn APIs

shifucun:
In https://github.com/datacommonsorg/mixer/pull/742, the placesIn API are moved into a separate module but the old impl were not removed (and they are stale code without merging).

This PR removes the stale code in places.go and remove GetChildPlaces helper function. Now all the callers would call placein.GetPlacesIn() function with non-payload response.


beets (approved):
nice, thanks for the fix.

PR #755, opened 2022-02-25
PlaceType Allowlist for Place Page ParentPlaces

chejennifer:
- as discussed in [this buganizer](https://b.corp.google.com/issues/220148645#comment5), filter parentPlaces returned by place page api for places that are of a type in the allowlist


beets (approved):
thanks for the quick fix!

PR #757, opened 2022-02-25
Investigate place stat var dcid no using human readable name

shifucun:


internal/server/statvar/golden/get_place_stat_vars/IG_idn.json
```diff
@@ -0,0 +1,3816 @@
...
+        "WHO/vita_Quintile3",
+        "WHO/vita_Quintile4",
+        "WHO/vita_Quintile5",
+        "WHO/vita_SecondaryOrHigher",
+        "dc/0ny9m7k0lzmn5",
+        "dc/7fmnve6yyc508",
+        "dc/n9rqwq8d3lcx3",
+        "dc/p49ghylpwz93f",
+        "dc/qdbhwwr88yhl7",
+        "dc/tbbjnf9bk96yh",
```
shifucun, line 3418: @pradh This is the same as Covid19MobilityTrend_Workplace. Is there a preference on the human curated stat var?

pradh: Wait this is not expected... if there is a human curated SV, that should be preferred... let me see

PR #758, opened 2022-02-25
Add a framework for latency test for import groups.

Spaceenter:
If this is OK, I could modify all the other golden tests.

In this PR, each API generates a separate file with one line "<apiName>,<duration-useImportGroup=false>,<duration-useImportGroup=true>".

Later on, we can easily combine all the files into a single CSV file.


test/e2e/ig_latency_GetPlacesIn.csv
```diff
@@ -0,0 +1 @@
+GetPlacesIn,2.244155,3.795856
```
pradh, line 1: Can we add header?

Spaceenter: This is intended to be a temp file. Later on, I plan to concatenate all the ~30 CSV files into a single CSV file. Header will be added at that time.

Spaceenter: Or, maybe we can just leave it as it is -- each API has its own latency CSV file? If so, yea could add a header.

docs/developer_guide.md
```diff
@@ -171,6 +171,13 @@ go run cmd/main.go \
 ./scripts/update_golden.sh
 ```
 
+### Run import group latency tests
+
+Run the following command in the root directory:
+```bash
```
pradh, line 177: So the normal tests will also run import-group?

test/e2e/ig_latency.go
```diff
@@ -0,0 +1,45 @@
...
+	"path"
+	"runtime"
+	"time"
+)
+
+func TestWithImportGroupLatency(
+	apiName string, opt *TestOption, testSuite func(*TestOption)) error {
+	durations := map[bool]float64{} // Map: useImportGroup -> duration in seconds.
+
+	for _, useImportGroup := range []bool{true, false} {
```
pradh, line 29: For reasonable performance evals, to avoid noise, it's generally a good idea to run a few times and report the individual numbers or mean/variance ?

shifucun, line 29: I wonder if the order here matters? If change it {false, true}, could the time change? Not sure if there is any Bigtable read cache.

test/e2e/ig_latency_GetPlacesIn.csv
pradh: A wrapper script to collect all the CSVs in one place SG.  After import-group rollout, this can act as a general perf regression suite, where we can report % diffs against checked-in version ...

internal/server/stat/golden/get_stat_all_test.go
```diff
@@ -80,4 +77,9 @@ func TestGetStatAll(t *testing.T) {
 			}
 		}
 	}
+
+	if err := e2e.TestWithImportGroupLatency(
+		"GetStatAll", &e2e.TestOption{}, testSuite); err != nil {
+		t.Errorf("TestWithImportGroupLatency() = %s", err)
+	}
```
pradh, line 84: neat!

pradh (approved):

test/e2e/ig_latency.go
pradh: I just commented, we should typically run all perf tests a few times...

Spaceenter: "true" and "false" are reading two separate BT tables? So there shouldn't be cache? (cache only happens if you read the same row in the same table?)

test/e2e/ig_latency_GetStatAll.csv
```diff
@@ -0,0 +1 @@
+GetStatAll,1.581786,3.319007
```
shifucun, line 1: This is huge difference...

internal/server/stat/golden/get_stat_all_test.go
```diff
@@ -33,10 +33,7 @@ func TestGetStatAll(t *testing.T) {
 	_, filename, _, _ := runtime.Caller(0)
 	goldenPath := path.Join(path.Dir(filename), "get_stat_all")
 
-	for _, opt := range []*e2e.TestOption{
-		{},
-		{UseImportGroup: true},
-	} {
+	testSuite := func(opt *e2e.TestOption) {
```
shifucun, line 36: when running this as performance test, should we stop before all the golden read and comparison. So the latency is purely for the API calls

docs/developer_guide.md
Spaceenter: Split - now we have normal test, and latency tests.

pradh: Sorry I was confused that latencyTest argument was always false and not passed as flag to the binary?

pradh (approved):
Thanks for the change!

docs/developer_guide.md
Spaceenter: For example, in get_places_in_test.go, line 109 has latencyTest=false, which is the normal golden test. Then on line 110 (inside e2e.TestWithImportGroupLatency), latencyTest=true, which is the latency test. 

So we have both latencyTest = true & false there.

pradh: Got it... sorry, bad eyes!

test/e2e/ig_latency.go
```diff
@@ -0,0 +1,81 @@
...
+	opt *TestOption,
+	testSuite func(*TestOption, bool)) error {
+	if LatencyTest {
+		return latencyTest(apiName, opt, testSuite)
+	}
+	goldenTest(opt, testSuite)
+	return nil
+}
+
+func goldenTest(opt *TestOption, testSuite func(*TestOption, bool)) {
```
shifucun, line 38: since goldenTest is here, maybe change filename from ig_latency.go to something general indicating this holds various test funcs

internal/server/stat/golden/get_stat_all_test.go
```diff
@@ -80,4 +82,9 @@ func TestGetStatAll(t *testing.T) {
 			}
 		}
 	}
+
+	if err := e2e.TestWithImportGroupLatency(
```
shifucun, line 86: update here to use "latencyTest" and "goldenTest"?

shifucun (approved):

test/e2e/ig_latency.go
Spaceenter: Good point. Renamed to test_driver.go

internal/server/stat/golden/get_stat_all_test.go
Spaceenter: latencyTest() and goldenTest() are called in e2e.TestWithImportGroupLatency().

but I see your confusion -- renamed TestWithImportGroupLatency to TestDriver().

Spaceenter:
Also saved all the latency result CSVs to a separate folder for being clean.

PR #760, opened 2022-02-27
Add latency tests for all methods. Add latency change rate to result …

Spaceenter:
…CSV.


internal/server/placein/golden/get_places_in_test.go
```diff
@@ -71,17 +71,17 @@ func TestGetPlacesIn(t *testing.T) {
...
-			}
-
 			var result []map[string]string
 			err = json.Unmarshal([]byte(resp.GetPayload()), &result)
 			if err != nil {
 				t.Errorf("Can not Unmarshal payload")
 				continue
 			}
 
+			if latencyTest {
```
shifucun, line 81: why move this block?

shifucun (approved):
Thanks for all the update! Looks like import group is consistently slower. Need to figure out the root cause..

PR #762, opened 2022-03-01
Sort and stabilize triples API output; Fix duplicate bug when entity has multiple types

shifucun:


internal/server/node/triple.go
```diff
@@ -272,19 +272,40 @@ func applyLimit(dcid string, t *pb.Triples, limit int) {
...
 				neighborTypes = t.ObjectTypes
 			}
-
-			for _, nt := range neighborTypes {
-				key := isOut + t.Predicate + nt
-				if _, ok := existTriple[key]; !ok {
-					existTriple[key] = []*pb.Triple{}
-				}
-				existTriple[key] = append(existTriple[key], t)
+			var nt string
```
pradh, line 275: should this choice of nt be deterministic?

pradh (approved):

internal/server/node/triple.go
shifucun: the neighbor type is deterministic now, either "" for value, or the first type of the corresponding entity.

PR #765, opened 2022-03-03
Exclude e2e.Setup() from latency test

Spaceenter:


pradh (approved):

shifucun (approved):
Thanks for the update and the result looks great! Curious now there are some very big latency drop now.

PR #767, opened 2022-03-03
Update latency test result with some fixes

shifucun:
- Allow running individual test, thus need to keep csv for each test
- Keep 3 decimal for result
- Update latency test boundary for GetPropertyLabels
- Minor format change

Overall the two mode has similar results. In the future should find a way to run this in cloudbuild with fixed cpu, memory.


test/e2e/run_ig_latency.sh
```diff
@@ -15,11 +15,31 @@
...
+    ARG=""
+fi
+
 export LATENCY_TEST=true
-go test -parallel 1 ./...
+go clean -testcache
+go test -parallel 1 ./... $ARG
 
 cd test/e2e/ig_latency
+> ig_latency
```
Spaceenter, line 41: What does this line do?

Spaceenter (approved):

test/e2e/run_ig_latency.sh
shifucun: to clear the content in ig_latency

PR #769, opened 2022-03-04
Remove duplicate observation period when computing metadata hash

shifucun:


internal/server/stat/stat_util.go
```diff
@@ -242,7 +242,6 @@ func getMetadataHash(m *pb.StatMetadata) uint32 {
 		m.ImportName,
 		m.MeasurementMethod,
 		m.ObservationPeriod,
```
shifucun, line 244: This is the actual fix

pradh (approved):

Spaceenter (approved):

PR #771, opened 2022-03-04
Update frequent Import Group version

pradh:
This is to include China Admin Areas.

Looks like the unemployment data refreshed as well.


deploy/storage/bigtable_import_groups.json
```diff
@@ -1,6 +1,6 @@
 {
   "tables": [
-    "frequent_2022_02_25_15_15_46",
+    "frequent_2022_03_03_23_08_30",
```
pradh, line 3: This is the main change.

ajaits (approved):

PR #773, opened 2022-03-05
Autopush to pick up version from import groups

shifucun:
It's easier to use a text file instead of JSON, given that there are no other information for the import groups.


scripts/deploy_gke.sh
```diff
@@ -58,8 +58,19 @@ if [[ $ENV == "autopush" ]]; then
   # Update bigtable and bigquery version
   gsutil cp gs://datcom-control/latest_base_cache_version.txt deploy/storage/bigtable.version
   gsutil cp gs://datcom-control/latest_base_bigquery_version.txt deploy/storage/bigquery.version
+  # Import group
+  # TODO: Loop through gs://datcom-control/autopush/datcom-control/autopush
+  > deploy/storage/bigtable_import_groups.version
+  for group in frequent infrequent ipcc
```
pradh, line 64: why not:

```
for src in $(gsutil ls gs://datcom-control/autopush/*_latest_base_cache_version.txt); do
  echo "Copying $src"
  gsutil cat $src >> deploy/storage/bigtable_import_groups.version
done
```

pradh: we could also require the files have a newline at the end

test/e2e/setup.go
```diff
@@ -96,12 +96,8 @@ func setupInternal(
...
 	schemaPath := path.Join(path.Dir(filename), mcfPath)
 
-	baseTableJSON, _ := ioutil.ReadFile(path.Join(path.Dir(filename), btJSON))
-	tableConfig := &bigtable.TableConfig{}
-	err := json.Unmarshal(baseTableJSON, &tableConfig)
-	if err != nil {
-		log.Fatalf("failed to read base table config: %v", err)
-	}
+	btGroupString, _ := ioutil.ReadFile(path.Join(path.Dir(filename), btGroup))
+	tableNames := strings.Split(string(btGroupString), "\n")
```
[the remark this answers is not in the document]
shifucun: Using bufio.ReadLine to read the lines now. also use builtin TrimSpace to trim each name

scripts/deploy_gke.sh
shifucun: This is much better bash! Also since echo produces a newline, we should be fine here even. the original txt has no newline.

shifucun:
Thanks!

test/e2e/setup.go
```diff
@@ -124,8 +121,11 @@ func setupInternal(
 	tables = append(tables, bigtable.NewTable(branchTableName, branchTable))
 
 	if useImportGroup {
-		for _, t := range tableConfig.Tables {
+		for _, t := range tableNames {
 			name := strings.TrimSpace(t)
```
pradh, line 125: nit: maybe we can safely trust the tableNames returned by util.ParseBigtableGroup() if it also skips empties? Especially since we are doing it in two places...

pradh (approved):
Thanks!

cmd/main.go
```diff
@@ -54,7 +54,7 @@ var (
 	useBaseBt         = flag.Bool("use_base_bt", true, "Use base bigtable cache")
 	baseTableName     = flag.String("base_table", "", "Base cache Bigtable table.")
 	useImportGroup    = flag.Bool("use_import_group", false, "Use multiple base tables from import group; Only applicable when --use_base_bt=true")
-	importGroupTables = flag.String("import_group_tables", "", "A JSON string of import group tables; Only applicable when --use_import_group")
+	importGroupTables = flag.String("import_group_tables", "", "Newline separated list of import group tables; Only applicable when --use_import_group")
```
Spaceenter, line 57: I think most programs use "comma separated list" instead of "newline separated list", because newline might be harder to type into a command line?

internal/util/util.go
```diff
@@ -418,3 +420,18 @@ func KeysToSlice(m map[string]bool) []string {
 	sort.Strings(s)
 	return s
 }
+
+// ParseBigtableGroup retrieves Bigtable names for import group from
+// newline separated list.
+func ParseBigtableGroup(s string) []string {
```
Spaceenter, line 426: If using "comma separated list", this can be:

importGroups := strings.Split(s, ",")

I think that's another point to better use comma instead of newline to separate. Because newline can mean "\n", "\r", "\n\r", etc. But comma just means ",".

pradh: That was my early concern, but seems like readLine might handle different newlines?  I do like the comma-seperated suggestion too.

shifucun: The reason I use newline separated format is for ease of manual editing. Imagine the version file contains 20 tables, it would be hard to update selected ones, and that's the only way for us to update the file.

test/e2e/setup.go
shifucun: Removed the extra check

PR #783, opened 2022-03-11
Add(update) the place stat date cache to use the raw data which has richer information

shifucun:
This will be used by map slider and source picker widget to fetch the source series(cohorts) information.

Should deprecate the original API when this (v1) is released.


internal/server/stat/golden/get_stat_date_within_place/CA_County.json
```diff
@@ -0,0 +1,156 @@
+{
+  "data": {
```
shifucun, line 2: @n-h-diaz @chejennifer This is how the response looks like.

internal/server/stat/golden/get_stat_date_within_place/CA_County.json
```diff
@@ -0,0 +1,163 @@
...
+          }
+        },
+        {
+          "date": {
+            "2011": 54,
+            "2012": 1,
+            "2014": 2
+          },
+          "metadata": {
+            "import_name": "HumanCuratedStats",
```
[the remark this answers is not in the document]
shifucun: We changed the name explicitly and decided to show this as "HumanCuratedStats". @pradh On the other hand, do we still need to surf this externally or could just filter this in Flume?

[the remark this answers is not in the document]
pradh: I thought we went through it changed the name?  With import-groups we can drop this from all the named import-groups, so it will only show up in laelaps cache.

proto/stat.proto
```diff
@@ -389,14 +389,23 @@ message GetStatSetAllResponse {
   map<uint32, StatMetadata> metadata = 2;
 }
 
-// Requests to get observation for all place.
-message GetPlaceObsRequest {
-  // The type of the place.
-  string place_type = 1;
+message StatDate {
+  // Map from date to the number of places.
+  map<string, double> date = 1;
```
pradh, line 394: date_popularity / date_frequency?

pradh: (if this is user visible)

internal/server/stat/stat_date.go
```diff
@@ -0,0 +1,89 @@
...
+			}
+		}
+	}
+
+	for sv := range result.Data {
+		if result.Data[sv] == nil {
+			result.Data[sv] = &pb.StatDateList{}
+			// Fetch from memdb.
+			if store.MemDb.HasStatVar(sv) {
+				result.Data[sv] = store.MemDb.ReadStatDate(sv)
```
pradh, line 84: I see.. so in terms of layering KG data trumps memdb data?  But the assumption is memdb will typically have non-overlapping (private) SVs?

pradh (approved):
Sorry for the late review

internal/server/stat/stat_date.go
```diff
@@ -0,0 +1,89 @@
...
+	if len(statVars) == 0 {
+		return nil, status.Errorf(codes.InvalidArgument,
+			"Missing required argument: stat_vars")
+	}
+
+	// Initialize result.
+	result := &pb.GetStatDateWithinPlaceResponse{
+		Data: make(map[string]*pb.StatDateList),
+	}
+	// Initialize with nil to help check if data is in mem-cache. The nil field
```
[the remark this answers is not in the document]
shifucun: Here indeed can be simplified by just checking `StatDate` field in the struct. Also this uncovers a bug which I fixed too.

internal/server/stat/golden/get_stat_date_within_place/CA_County.json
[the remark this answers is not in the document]
shifucun: Yes, HumanCuratedStats is the changed name. And dropping it from non-laelaps groups sounds good.

proto/stat.proto
shifucun: How about date_place_count?

internal/server/stat/stat_date.go
shifucun: yea, now memdb assumes only having new stat vars.

internal/server/stat/golden/get_stat_date_within_place/CA_County.json
chejennifer: this looks good to me!

chejennifer (approved):

internal/server/stat/golden/get_stat_date_within_place/CA_County.json
n-h-diaz: got it thanks! will switch to using dates from this!

n-h-diaz (approved):

proto/stat.proto
pradh: SG

beets (approved):
thanks for the updates!

shifucun:
Thanks!

PR #785, opened 2022-03-15
Clear bigtable_import_groups.version before updating

shifucun:


pradh (commented):
Thanks for the cleanup!

PR #787, opened 2022-03-23
Add an API to find stat var matches given property values

shifucun:
This will be used by the NLP interface to get ranked stat vars. Current version has a very basic ranking based on match count. We will look into better ranking based on must/optional match, property weights etc.


proto/stat_var.proto
```diff
@@ -232,4 +232,22 @@ message GetStatVarSummaryRequest {
...
   map<string, StatVarSummary> stat_var_summary = 1;
+}
+
+message GetStatVarMatchRequest {
+  // The property value pair. Ex
+  // {gender: Male}, {statType: measuredValue}
+  // To improve this, can explicitly specify "statType", "measuredProperty",
+  // and constraint properties. Then the match could have must-match terms and
+  // optional match terms that have weights. That would greatly reduce the
+  // number of returned match stat vars.
```
shifucun, line 243: @nopper Let me know if this makes sense

internal/server/statvar/statvar_match.go
```diff
@@ -0,0 +1,60 @@
...
+) (*pb.GetStatVarMatchResponse, error) {
+	propertyValue := in.GetPropertyValue()
+	statVarCount := map[string]int32{}
+	// TODO: consider parallel this if performance is an issue.
+	for property, value := range propertyValue {
+		resp, err := node.GetPropertyValuesHelper(ctx, store, []string{value}, property, false)
+		if err != nil {
+			return nil, err
+		}
+		for _, node := range resp[value] {
```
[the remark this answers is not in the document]
shifucun: Added filter here

internal/server/statvar/statvar_match.go
```diff
@@ -0,0 +1,60 @@
...
+			statVarCount[node.Dcid]++
+		}
+	}
+	result := &pb.GetStatVarMatchResponse{}
+	statVars := []string{}
+	for statVar := range statVarCount {
+		statVars = append(statVars, statVar)
+	}
+	sort.Strings(statVars)
```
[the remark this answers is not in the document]
shifucun: Merged this to the sort below.

proto/stat_var.proto
```diff
@@ -232,4 +232,22 @@ message GetStatVarSummaryRequest {
...
+  map<string, string> property_value = 1;
```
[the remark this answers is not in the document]
shifucun: What is multiple values?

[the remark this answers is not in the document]
shifucun: @nopper Can you confirm if multiple values could exist for one property?

This is a base version of the API which we hope to iterate. As in the comment above, more information could be added in the request, so I am trying to make it simple for now and it would need more changes anyway.

proto/stat_var.proto
```diff
@@ -232,4 +232,22 @@ message GetStatVarSummaryRequest {
...
+}
+
+message GetStatVarMatchResponse {
+  message MatchInfo {
+    string stat_var = 1;
+    int32 match_count = 2;
+  }
+  repeated MatchInfo match_info = 1;
```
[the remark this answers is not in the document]
shifucun: Done

pradh (approved):
Thanks for the changes.

internal/server/statvar/golden/get_statvar_match_test.go
```diff
@@ -0,0 +1,88 @@
+// Copyright 2021 Google LLC
```
nopper, line 1: Should this be 2022?

proto/stat_var.proto
```diff
@@ -232,4 +232,22 @@ message GetStatVarSummaryRequest {
...
+  // and constraint properties. Then the match could have must-match terms and
+  // optional match terms that have weights. That would greatly reduce the
+  // number of returned match stat vars.
+  map<string, string> property_value = 1;
+}
+
+message GetStatVarMatchResponse {
+  message MatchInfo {
+    string stat_var = 1;
+    int32 match_count = 2;
```
nopper, line 252: How this is computed? add a comment maybe?

proto/stat_var.proto
nopper: Just to clarify, at the moment `statType` and `measuredProperty` will be ignored? Only constraint properties work?

proto/stat_var.proto
```diff
@@ -232,4 +232,22 @@ message GetStatVarSummaryRequest {
...
+
+message GetStatVarMatchRequest {
+  // The property value pair. Ex
+  // {gender: Male}, {statType: measuredValue}
+  // To improve this, can explicitly specify "statType", "measuredProperty",
+  // and constraint properties. Then the match could have must-match terms and
+  // optional match terms that have weights. That would greatly reduce the
+  // number of returned match stat vars.
+  map<string, string> property_value = 1;
+}
```
nopper, line 245: I would probably add a `k` parameter to limit the number of returned matches. I don't think we are interested in downloading the full schema at every request.

internal/server/statvar/golden/get_statvar_match_test.go
```diff
@@ -0,0 +1,88 @@
...
+			propertyValues map[string]string
+			goldenFile     string
+		}{
+			{
+				map[string]string{
+					"gender":           "Female",
+					"measuredProperty": "count",
+					"nativity":         "USC_ForeignBorn",
+				},
+				"female_usc_foreignborn.json",
```
nopper, line 47: These are pretty large files for testing. Can we trim it down to managable size, say at most 50 statvars?

nopper (commented):
Thanks looks pretty good. Left some comments here and there. PTAL

proto/stat_var.proto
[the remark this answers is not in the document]
nopper: The current prototype only emits a single value per property. So it should be fine like this at the moment.

proto/stat_var.proto
shifucun: I will submit this for now. Once we have a better idea of partial match, I will send a follow up PR.

PR #791, opened 2022-03-28
Only keep the post endpoint for stat var match API

shifucun:
Apparently "get" does not work when request have map or nested message


nopper:
LGTM.

proto/mixer.proto
```diff
@@ -336,8 +336,7 @@ service Mixer {
   rpc GetStatVarMatch(GetStatVarMatchRequest)
       returns (GetStatVarMatchResponse) {
     option (google.api.http) = {
-      get : "/stat-var/match"
-      additional_bindings : {post : "/stat-var/match" body : "*"}
+      post : "/stat-var/match" body : "*"
```
nopper, line 339: LGTM

nopper (approved):
LGTM.

PR #796, opened 2022-03-29
Implement Variables() and BulkVariables() methods.

Spaceenter:


internal/server/v1/variables/variables.go
```diff
@@ -0,0 +1,121 @@
...
+
+// Variables implements API for Mixer.Variables.
+func Variables(
+	ctx context.Context,
+	in *pb.VariablesRequest,
+	store *store.Store,
+) (*pb.VariablesResponse, error) {
+	entity := in.GetEntity()
+	if entity == "" {
+		return nil, status.Error(codes.InvalidArgument, "Missing required arguments: entity")
```
shifucun, line 52: In theory this check is not necessary since the ESP will check that the path is not valid if "entity" is missing. This is more of URI error than argument error.

internal/server/v1/variables/variables.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	resp := &pb.VariablesResponse{Entity: entity}
+	statVars, ok := entityToStatVars[entity]
+	if !ok {
+		return resp, nil
+	}
+	resp.Variables = append(resp.Variables, statVars.StatVars...)
```
shifucun, line 68: This should be assignment rather than append?

internal/server/v1/variables/variables.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+		sort.Strings(resp.Data[0].Variables)
+	} else {
+		sort.Strings(entities)
+		for _, entity := range entities {
+			if statVars, ok := entityToStatVars[entity]; ok {
+				resp.Data = append(resp.Data, &pb.VariablesResponse{
+					Entity:    entity,
+					Variables: statVars.StatVars,
+				})
```
shifucun, line 115: for !ok case here, should have an entry with only "Entity" field too?

proto/mixer.proto
```diff
@@ -357,4 +359,19 @@ service Mixer {
       additional_bindings : {post : "/v1/bulk/properties/{direction}" body : "*"}
     };
   }
+
+  rpc Variables(datacommons.v1.VariablesRequest)
+      returns (datacommons.v1.VariablesResponse) {
+    option (google.api.http) = {
+      get : "/v1/variables/{entity}"
```
shifucun, line 366: {entity=**}, otherwise dcid with "/" won't work

shifucun (approved):
Very nice! Just some minor comments.

internal/server/v1/variables/variables.go
Spaceenter: Removed.

internal/server/v1/variables/variables.go
Spaceenter: done

internal/server/v1/variables/variables.go
Spaceenter: done

proto/mixer.proto
Spaceenter: done

PR #797, opened 2022-03-29
Use SearchNames instead of SearchName for stat var search

shifucun:
"SearchNames" contain more curated names and is a superset of "SearchName" field.

Once this is in prod. We can deprecate "SearchName"


pradh (approved):
Thanks!

chejennifer (approved):
thanks for the change!

PR #798, opened 2022-03-30
[V1 API] Implement /v1/observations/point/*

shifucun:


proto/mixer.proto
```diff
@@ -374,4 +375,11 @@ service Mixer {
       additional_bindings : {post : "/v1/bulk/variables" body : "*"}
     };
   }
+
+    rpc ObservationsPoint(datacommons.v1.ObservationsPointRequest)
+      returns (PointStat) {
```
Spaceenter, line 380: I think it's better to do:

returns (datacommons.v1.ObservationsPointResponse)

?

This is to make request and response having the same format for all APIs.

internal/server/v0/statpoint/get_stat_value.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	result := &pb.GetStatValueResponse{}
+	obsTimeSeries = btData[place][statVar]
+	if obsTimeSeries == nil {
+		return result, nil
+	}
+	obsTimeSeries.SourceSeries = stat.FilterSeries(obsTimeSeries.SourceSeries, filterProp)
+	value, err := stat.GetValueFromBestSource(obsTimeSeries, date)
+	if err != nil {
+		log.Println(err)
```
Spaceenter, line 66: remove this?

Spaceenter (approved):

proto/mixer.proto
shifucun: I started with that but realized these two protos are exactly the same, and there is no alias in protobuf. If we create another one, then need a no-op copy of all the data.

This is actually a common issue which even now exist, for example we have:

rpc GetStatVarGroup(GetStatVarGroupRequest) returns (StatVarGroups)

Basically if we directly return the data from cache, then the response will be the cache proto..

Maybe we leave it for now before we come to a conclusion for this.

PR #799, opened 2022-03-30
Implement PlaceInfo() and BulkPlaceInfo().

Spaceenter:


internal/server/v1/places/places.go
```diff
@@ -0,0 +1,81 @@
+// Copyright 2022 Google LLC
```
shifucun, line 1: Should we call this file "info.go", since this is already under "places".

Also I wonder should we use singular form for folder name like "place", since this is a general concept instead of REST resource

internal/server/v1/variables/variables.go
```diff
@@ -12,20 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Copyright 2022 Google LLC
```
shifucun, line 15: nice..

proto/v1/places.proto
```diff
@@ -0,0 +1,36 @@
...
+
+import "place.proto";
+
+message PlaceInfoRequest {
+  string entity = 1;
+}
+  
+message PlaceInfoResponse {
+  string entity = 1;
+  datacommons.PlaceMetadata data = 2;
```
shifucun, line 27: data -> info? In the response, there is top level "data" and the data here. I feel it's helpful to have more concrete field name here.

internal/server/v1/places/places.go
Spaceenter: Done.

proto/v1/places.proto
Spaceenter: Done.

shifucun (approved):

PR #805, opened 2022-03-31
Remove schema mapping for weather data

shifucun:
The data is added as regular stat var observations.


deploy/mapping/kustomization.yaml
```diff
@@ -19,6 +19,4 @@ configMapGenerator:
   - name: schema-mapping
     files:
       - base.mcf
-      - dailyweather.mcf
```
pradh, line 22: Same question as in https://github.com/datacommonsorg/website/pull/1500#issuecomment-1084994900 - should we wait until dailyweather lands in obs?

pradh (approved):
Thanks!

PR #806, opened 2022-03-31
Use real BQ table for translator tool

shifucun:


pradh (approved):

internal/server/translator/translate.go
```diff
@@ -38,7 +38,7 @@ func Translate(
 	}
 
 	out := pb.TranslateResponse{}
-	mappings, err := mcf.ParseMapping(in.GetSchemaMapping(), "bq")
+	mappings, err := mcf.ParseMapping(in.GetSchemaMapping(), "datcom-store.dc_kg_latest")
```
pradh, line 41: Can add a TODO to switch to the BQ analytics table once we switch over.  Then the public will have access to it...

shifucun: Done

PR #807, opened 2022-04-01
Implement PlacePage().

Spaceenter:


proto/v1/page.proto
```diff
@@ -26,3 +27,19 @@ message ProteinPageResponse {
...
+  // added to chart config but data is not added to cache (delay from cache
+  // build).
+  repeated string new_stat_vars = 2;
+  // Seed value for random selection. Used by test to get deterministic result.
+  int64 seed = 3;
+}
+
+message PlacePageResponse {
+  string entity = 1;
+  datacommons.GetPlacePageDataResponse data = 2;
```
shifucun, line 44: Since we only have singular version of this API and it's internal, can just use GetPlacePageDataResponse here (so no need to have the wrapper with entity)?

Spaceenter: done

shifucun (approved):

PR #809, opened 2022-04-01
Add newline to translator SQL

shifucun:


beets (approved):
wow, thanks for adding all the newlines!

PR #813, opened 2022-04-06
Data Release 2022-04-06

shifucun:
Not much change since last version update and they all look reasonable.


shifucun:
thanks!

deploy/storage/bigtable_import_groups.version
```diff
@@ -1,4 +1,4 @@
-frequent_2022_04_04_01_15_38
+frequent_2022_04_06_08_29_34
```
pradh, line 1: Can we keep this as is?

PR #815, opened 2022-04-06
Less aggressive rolling update

shifucun:
Too many pods concurrently rolling updates can cause peak resources usages, especially now there is intense computation to build stat var search index. This causes frequent restart and slower completion.

### DIFF
--- a/deploy/base/deployment.yaml
+++ b/deploy/base/deployment.yaml
@@ -25,9 +25,10 @@ spec:
   strategy:
     type: RollingUpdate
     rollingUpdate:
-      # maximum number of Pods that can be created over the desired number of Pods
+      # Maximum number of pods that can be created over the desired number of pods.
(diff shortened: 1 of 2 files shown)


deploy/base/deployment.yaml
```diff
@@ -28,7 +28,7 @@ spec:
       # maximum number of Pods that can be created over the desired number of Pods
       maxSurge: 1
       # maximum number of Pods that can be unavailable during the update process
-      maxUnavailable: 25%
+      maxUnavailable: 5
```
pradh, line 31: Maybe note 5 is what % of pods?

pradh (approved):

deploy/base/deployment.yaml
shifucun: Done.

PR #821, opened 2022-04-09
[V1 API] Implement /v1/bulk/observations/series

shifucun:


proto/mixer.proto
```diff
@@ -422,8 +422,8 @@ service Mixer {
   rpc BulkObservationsPoint(datacommons.v1.BulkObservationsPointRequest)
       returns (datacommons.v1.BulkObservationsPointResponse) {
     option (google.api.http) = {
-      get : "/v1/bulk/observations"
-      additional_bindings : {post : "/v1/bulk/observations" body : "*"}
+      get : "/v1/bulk/observations/point"
+      additional_bindings : {post : "/v1/bulk/observations/points" body : "*"}
```
Spaceenter, line 426: "get" uses "point", but post uses "points" -- shall we be consistent between "get" and "post"?

Spaceenter (approved):

proto/mixer.proto
shifucun: Good catch! Changed

PR #824, opened 2022-04-12
[stat var hierarchy] only cache blocklisted search index

chejennifer:
- Originally, autopush used the full search index and staging and prod used the blocklisted search index. Search index takes up a lot of memory and there is not much value in keeping 2 copies of the search index.


internal/server/statvar/golden/search_statvar_test.go
```diff
@@ -36,102 +36,88 @@ func TestSearchStatVar(t *testing.T) {
...
 				"crime.json",
 			},
 			{
 				"female",
 				[]string{"country/USA"},
 				false,
-				false,
 				"female.json",
 			},
 			{
```
shifucun, line 62: can remove this test and the test file

shifucun (approved):
Thanks for the quick update!

internal/server/statvar/golden/search_statvar_test.go
chejennifer: wait, why would I remove this test file? this file tests for the search results retrieved given a query ..

PR #826, opened 2022-04-14
Facet ranking logic update with 'inferior facet' concept

shifucun:
- Facet with score below a threshold (1000) is considered as low quality, interior facet and only used when no better facet is present.
- Use this for observation point related APIs.
- Fixed weather related facet ranking.


internal/server/ranking/ranking.go
```diff
@@ -139,9 +139,9 @@ func (a CohortByRank) Swap(i, j int) {
 
 func (a CohortByRank) Less(i, j int) bool {
 	oi := a[i]
-	scorei := getScorePb(oi)
+	scorei := GetScorePb(oi)
 	oj := a[j]
-	scorej := getScorePb(oj)
+	scorej := GetScorePb(oj)
 	// Higher score value means lower rank.
```
pradh, line 145: Should we drop cohort "place coverage" ranking to be aligned with BQ results?

internal/server/stat/stat_point.go
```diff
@@ -256,26 +256,23 @@ func GetStatSetWithinPlace(
...
+				// Otherwise, response date is the latest date from the cache.
+				respDate := date
+				if respDate == "" {
+					respDate = cohort.PlaceToLatestDate[place]
+				}
+				pointStat, exist := result.Data[statVar].Stat[place]
+				// No data set yet.
+				shouldSetValue := !exist
+				// New date is from good facet with later date
+				shouldResetValue := exist && respDate > pointStat.Date && !IsInferiorFacetPb(cohort)
```
pradh, line 272: I'm not fully sure I follow this, but are we doing this only when the user has not provided the date?

Can we add more description to that line?

internal/server/v1/observations/bulk_point.go
```diff
@@ -88,10 +86,16 @@ func BulkPoint(
...
 							entityObservations.PointsByFacet, ps)
 					}
+					result.Facets[facet] = metadata
+					if !allFacets {
+						break
+					}
 				} else {
-					// Date is not given, find the latest value
-					latestDate := ""
+					if !allFacets && idx > 0 && stat.IsInferiorFacetPb(series) {
```
pradh, line 94: Comment?

internal/server/v1/observations/golden/bulk_point/preferred_latest.json
```diff
@@ -260,11 +319,21 @@
       "measurement_method": "BLSSeasonallyAdjusted",
       "observation_period": "P1M"
     },
+    "3981252704": {
+      "import_name": "WorldDevelopmentIndicators",
+      "provenance_url": "https://datacatalog.worldbank.org/dataset/world-development-indicators/",
+      "observation_period": "P1Y"
+    },
```
pradh, line 326: Are any of these diffs due to the change in the logic?

pradh (commented):
Nice!

internal/server/ranking/ranking.go
shifucun: Done

internal/server/stat/stat_point.go
shifucun: Added more comments

internal/server/v1/observations/bulk_point.go
shifucun: Done

internal/server/v1/observations/golden/bulk_point/preferred_latest.json
shifucun: Yes,  these are from the logic change and are as expected.

internal/server/stat/stat_point.go
pradh: Thanks.

Re-reading this, I think the answer to my question is that when caller sets "date" then this whole logic of shouldResetValue will never be true, because the data returned from cache is only for that date, right?

pradh (approved):

internal/server/stat/stat_point.go
shifucun: Yep!

PR #828, opened 2022-04-20
Use SearchName in bleveIndex as svDefinition seems to be empty with new data.

nopper:


shifucun (commented):
SearchName is going to be removed, so please use "SearchNames" field. On the other hand, from quality point of view, is the search names better than "definition"?

shifucun:
The "definition" issue is fixed now, so it should be available in the latest autopush instance.

nopper:
SG. Then let's close the PR.

PR #830, opened 2022-04-22
Restructure memdb to support custom stat var name and hierarchy in private import

shifucun:
This is a major refactor of private mixer instance to enrich the support for private import.

- Support user provided stat var group naming and hierarchy 
- Consolidate manifest and stat var group config as a checked in config file, instead of a file in GCS. This allows better management and tracking
- Merge customer provided stat var hierarchy into the main hierarchy, so stat var search, display are baked in the API
- Remove cloud pubsub of data folder update. This complicates the system with not much benefit so far.
image


internal/store/memdb/memdb.go
```diff
@@ -300,6 +343,30 @@ func (memDb *MemDb) LoadFromGcs(ctx context.Context, bucket, prefix string) erro
 		}
 	}
 	log.Printf("Number of csv rows added: %d", count)
+	// Populate placeSvExistence field
```
pradh, line 346: nit: This function is large enough... maybe this can be a helper with a useful name (like buildMemSVExistenceCache()?)

pradh (commented):
Nice!

shifucun (commented):
Thanks for review!

deploy/overlays/private/memdb.json
```diff
@@ -0,0 +1,105 @@
+{
+  "importName": "Feeding America",
+  "provenanceUrl": "https://www.feedingamerica.org/",
+  "dataDownloadUrl": "https://www.feedingamerica.org/",
+  "rootSvg": "g/Feeding_America",
```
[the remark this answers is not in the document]
pradh: This is fine for now.

(1) ... I think private DC people will probably need to deal with TMCF/MCF anyway?  But I see the point in avoiding code to build the proto-like thing :).  Perhaps we could perhaps simplify by not requiring separate "displayName" vs. "searchNames" etc?

(3) ... I just asked because this was very Feeding America specific, and only supported one import :)

internal/server/v1/observations/golden/bulk_point_linked_test.go
```diff
@@ -41,31 +41,31 @@ func TestBulkObservationsPointLinked(t *testing.T) {
...
-				"CA_County.json",
-			},
-			{
-				"County",
-				"country/USA",
-				[]string{"Count_Person", "Median_Age_Person", "NumberOfMonths_WetBulbTemperature_35COrMore_RCP45_MinRelativeHumidity"},
-				"",
-				"USA_County.json",
-			},
+			// {
```
pradh, line 44: Want to delete instead of comment?

pradh (approved):
Thanks for the changes.

deploy/overlays/private/memdb.json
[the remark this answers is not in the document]
shifucun: (1) We could definitely simply the config requirement, so it minimally just asks a list of DCID and names (even names could be omitted then DCID is used to display).

internal/server/v1/observations/golden/bulk_point_linked_test.go
shifucun: I need to talk to Francesco about this flakiness. We do want to uncomment this when he has a proper fix.

PR #837, opened 2022-04-28
Add protobuf for pagination and functions to encode/decode pagination proto

shifucun:
Non-bulk API token example: `H4sIAAAAAAAA/+JSEWKSEBFi42AUYJTgEmLjYILSzAJMEkZCbBwsID4AAAD//wEAAP//htWoVyYAAAA=`

Bulk API (2 entities) token example:
`H4sIAAAAAAAA/+LS4+JIT833TNE3MBVikhARYuNgFGCU4BJi42CC0swCTBJGQmwcLCA+lxFcvZkQqwCrxAlGsBYOiRQhdg4mAXaJaYxgPfwSNmA9LBIaAAAAAP//AQAA//82b3t4ZAAAAA==`

The length of the token seems okay.


proto/v1/pagination.proto
```diff
@@ -0,0 +1,37 @@
...
+package datacommons.v1;
+
+
+message ImportGroupCursor {
+  // The index of the import group.
+  int32 index = 1;
+  // The page index of the cursor, starts from 0.
+  int32 page = 2;
+  // The position of the next read item in the current page, starts from 0.
+  int32 pos = 3;
```
pradh, line 26: name nit:  s/index/import_group/  s/pos/item/

that way, the field names are names of "things" that are indexed.

proto/v1/pagination.proto
```diff
@@ -0,0 +1,37 @@
...
+}
+
+message ImportGroupCursorGroup {
+  // Entity DCID or other information that identifies the ImportGroupCursorGroup.
+  string key = 1;
```
pradh, line 31: what are the different possibilities here?  is it always "entity"? if not, the name SG.

pradh (approved):
Nice.

proto/v1/pagination.proto
shifucun: Nice change.

proto/v1/pagination.proto
shifucun: It could be entity dcids or could be multiple properties (when building triples API from propVal APIs, concurrently asking for multiple properties)

PR #838, opened 2022-04-29
Adapt Bigtable Reader to take distinct keys for each import group table

shifucun:


internal/store/bigtable/reader.go
```diff
@@ -76,72 +76,62 @@ func readRowFn(
...
-//		go struct based on the cache content.
-// getToken: A function to get back the indexed token (like place dcid) from
-//		bigtable row key.
-//
+// Note all Bigtable read use the same set of rowList.
 func Read(
 	ctx context.Context,
 	btGroup *Group,
-	rowSet cbt.RowSet,
+	rowSet cbt.RowList,
```
Spaceenter, line 83: Is it worthwhile to change the variable name: rowSet->rowList?

internal/store/bigtable/reader.go
```diff
@@ -76,72 +76,62 @@ func readRowFn(
...
-		return nil, status.Errorf(codes.Internal, "Unsupported RowSet type: %v", v)
+// ReadWithGroupRowSet reads BigTable rows from multiple Bigtable in parallel.
+// Reading is chunked as the size limit for RowSet is 500KB.
+//
+// Note the read could have different RowList for each import group Bigtable as
+// needed by the pagination APIs.
+func ReadWithGroupRowSet(
+	ctx context.Context,
+	btGroup *Group,
+	rowSetMap map[int]cbt.RowList,
```
Spaceenter, line 102: Is it worthwhile to change the variable name: rowSetMap->rowListMap?

internal/store/bigtable/reader.go
```diff
@@ -76,72 +76,62 @@ func readRowFn(
...
-		if len(tables) > 0 {
-			for j := 0; j < len(tables); j++ {
-				j := j
-				if tables[j] != nil {
-					errs.Go(readRowFn(errCtx, tables[j], rowSetPart, getToken, action, chans[j]))
-				}
+	// Read from each import group tables. Note each table could have different
+	// rowSet in pagination APIs.
+	for i := 0; i < len(tables); i++ {
+		i := i
```
Spaceenter, line 123: This line is not quite useful?

Spaceenter (changes requested):

Spaceenter (approved):

internal/store/bigtable/reader.go
shifucun: The errs.Go routine takes a func that has i in the imput, I think here needs to reassign the value i to avoid it being captured there

internal/store/bigtable/reader.go
shifucun: Done

internal/store/bigtable/reader.go
shifucun: Done

shifucun (commented):
Thanks for review!

PR #841, opened 2022-05-04
[V1 API] Implement bulk series linked API and made a bunch of bug fixes

shifucun:


proto/mixer.proto
```diff
@@ -450,6 +450,14 @@ service Mixer {
     };
   }
 
+  rpc BulkObservationsSeriesLinked(datacommons.v1.BulkObservationsSeriesLinkedRequest)
+      returns (datacommons.v1.BulkObservationsSeriesResponse) {
```
Spaceenter, line 454: If this is an API endpoing visible to external users, I slightly feel it's better to create its own response proto message instead of reusing one, even though their contents might be the same:

BulkObservationsSeriesLinkedResponse

internal/server/v1/observations/bulk_point_linked.go
```diff
@@ -47,7 +47,7 @@ func BulkPointLinked(
 		return nil, status.Errorf(codes.InvalidArgument,
 			"Missing required argument: linked_entity")
 	}
-	if linkedProperty == "" {
+	if linkedProperty != "containedInPlace" {
```
Spaceenter, line 50: So we only support 'containedInPlace' as linkedProperty for now... Shall we document this somewhere?

Or, shall we indicate this in the error message below? Something like:

"Invalid argument: linked_property. Now the following values are supported: containedInPlace, XXX, YYY, etc."

Spaceenter (approved):

internal/server/v1/observations/bulk_point_linked.go
shifucun: Done

proto/mixer.proto
shifucun: 1. The user visible JSON structure won't be different with current or a newly created proto if they are the same.
2. If we use a new proto for naming clearness, there would need a dummy transformation from old proto to the new proto since the "linked" API calls the bulk API. In that sense, it's easier for those APIs share the same response proto.

PR #843, opened 2022-05-07
[V1 API] paged property values (out) API

shifucun:
- Refactor the processor struct into "state" struct which allows composition for "in" and "out" cases.
- Consolidate to use one handler for both "in" and "out" cases.
- More abstraction of "state" functions for re-use.
- Change naming patter from "in_property_values" to "property_values_in" to match API URI


internal/server/v1/propertyvalues/common.go
```diff
@@ -0,0 +1,39 @@
...
+
+func buildEmptyCursorGroup(n int) *pb.CursorGroup {
+	result := &pb.CursorGroup{Cursors: []*pb.Cursor{}}
+	for i := 0; i < n; i++ {
+		result.Cursors = append(result.Cursors, &pb.Cursor{ImportGroup: int32(i)})
+	}
+	return result
+}
+
+var action = func(jsonRaw []byte) (interface{}, error) {
```
pradh, line 35: nit: action seems too general a name -  perhaps unmarshalAction?

proto/v1/property_values.proto
```diff
@@ -37,11 +37,13 @@ message InPropertyValuesRequest {
   // [Optional]
   // The pagination token for getting the next set of entries. This is empty
   // for the first request and needs to be set in the subsequent request.
-  // This is the value returned from a prior call to InPropertyValuesResponse
+  // This is the value returned from a prior call to PropertyValuesResponse
   string next_token = 4;
+  // Direction can only be "in" and "out"
+  string direction = 5;
```
pradh, line 43: Should we use enum (preferable), or boolean?

internal/server/v1/propertyvalues/property_values.go
```diff
@@ -97,132 +78,123 @@ func InPropertyValues(
...
+			if !hasNext {
+				break
+			}
 		}
-	}
-	respToken := ""
-	for _, d := range proc.rawEntities {
-		// If any import group has more data, then should compute the token
-		if d != nil {
+		// If used import group has more data, then should compute the token
```
pradh, line 98: Add some clarifying comment why this is only one import group, unlike "in"?

internal/server/v1/propertyvalues/state.go
```diff
@@ -0,0 +1,184 @@
...
+	state
+	// Min heap for entity merge sort
+	heap *entityHeap
+}
+
+type outState struct {
+	state
+	// The used import group (only one import group is used for out property values).
+	usedImportGroup int
+	//
```
pradh, line 51: nit: delete line

internal/server/v1/propertyvalues/state.go
```diff
@@ -0,0 +1,184 @@
...
+	"container/heap"
+	"context"
+	"log"
+
+	cbt "cloud.google.com/go/bigtable"
+	pb "github.com/datacommonsorg/mixer/internal/proto"
+	"github.com/datacommonsorg/mixer/internal/store/bigtable"
+)
+
+type state struct {
```
pradh, line 27: struct level comment?

internal/server/v1/propertyvalues/state.go
```diff
@@ -0,0 +1,184 @@
...
+
+type inState struct {
+	state
+	// Min heap for entity merge sort
+	heap *entityHeap
+}
+
+type outState struct {
+	state
+	// The used import group (only one import group is used for out property values).
```
pradh, line 49: I see, so this is how it is today?

internal/server/v1/propertyvalues/state.go
```diff
@@ -0,0 +1,184 @@
...
+		nil,
+	)
+	if err != nil {
+		return err
+	}
+	// Store the raw cache data.
+	for idx, btData := range btDataList {
+		if data, ok := btData[entity]; ok {
+			pe := data.(*pb.PagedEntities)
+			s.rawEntities = append(s.rawEntities, pe.Entities)
```
pradh, line 94: Should we append triples across multiple import-groups for out too? (assuming each distinct idx is a separate import-group)

internal/server/v1/propertyvalues/state.go
```diff
@@ -0,0 +1,184 @@
...
+		ctx,
+		btGroup,
+		rowListMap,
+		action,
+		nil,
+	)
+	if err != nil {
+		return err
+	}
+	if data, ok := btDataList[importGroup][s.entity]; ok {
```
pradh, line 125: assert !arcOut || importGroup == s.usedImportGroup ?

internal/server/v1/propertyvalues/state.go
shifucun: Done

internal/server/v1/propertyvalues/state.go
shifucun: Yep

internal/server/v1/propertyvalues/state.go
shifucun: This is the base case where only raw entities are saved. The rawEntities looks like
[
  [ig1_1, ig1_2, ..],
  [],
  [ig3_1, ig3_2]
]
There are no merging here.

internal/server/v1/propertyvalues/state.go
shifucun: Here "s" is the base struct that does not have usedImportGroup field. The caller func nextOut() in property_values.go could guarantee this condition.

proto/v1/property_values.proto
shifucun: This field is parsed by the ESP from the REST url where we could only get a string.

pradh (approved):

PR #844, opened 2022-05-10
[V1 API] Implement /v1/variable/ancestors/{entity} API

shifucun:
This is equivalent to the current /stat-var/path API


internal/server/v1/variable/ancestors.go
```diff
@@ -0,0 +1,56 @@
...
+	in *pb.VariableAncestorsRequest,
+	store *store.Store,
+	cache *resource.Cache,
+) (*pb.VariableAncestorsResponse, error) {
+	entity := in.GetEntity()
+	if entity == "" {
+		return nil, status.Errorf(
+			codes.InvalidArgument, "Missing required argument: entity")
+	}
+	ancestors := []string{entity}
```
Spaceenter, line 40: "ancestors" includes itself?

Spaceenter (approved):

internal/server/v1/variable/ancestors.go
shifucun: Nice catch. The current api has self in the response, which is indeed redundant. Removed it in v1.

PR #845, opened 2022-05-10
[V1 API] Implement /v1/variable/groups

shifucun:


Spaceenter (approved):

proto/v1/variable.proto
```diff
@@ -30,3 +32,14 @@ message VariableAncestorsResponse {
...
+
+message VariableGroupsRequest {
+  // (optional) When specified, only return the variable group that the entities
+  // have data for.
+  repeated string entities = 1;
+}
+
+message VariableGroupsResponse {
+  // Key is StatVar Group ID.
+  map<string, StatVarGroupNode> data = 1;
```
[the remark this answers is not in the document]
shifucun: Yea, in general I think list is better. For this case, the response access pattern involves point access. Also this is the raw svg data used by several other related API, changing the interface would require quite some changes. For now it prob be good to keep as is.

proto/v1/variable.proto
```diff
@@ -30,3 +32,14 @@ message VariableAncestorsResponse {
   // alphabetical sorted order is selected.
   repeated string ancestors = 1;
 }
+
+message VariableGroupsRequest {
+  // (optional) When specified, only return the variable group that the entities
```
pradh, line 37: nit: variable groups

pradh (approved):

proto/v1/variable.proto
shifucun: Done

PR #847, opened 2022-05-17
Refactor bigtable reader key token handling

shifucun:
Bigtable reader function needs to handle bulk read of keys like:

d/1/geoId1/prop1
d/1/geoId1/prop2
d/2/geoid2/prop1

The response rows need to be grouped (in this case, by place id and by prop).

This is now handled by passing a func to process the key into tokens and key the result by these tokens. This approach has a few issues:

1) multiple dimension key components can not be handled concisely. A separate key to token mapping needs to be used.
2) A rowlist builder function is needed for each cache. All these functions are very similar, in the form of prefix^token1^token2..., with slightly different variant.

This PR simplifies this approach by returning the key components together with the actual data. This completely removes the callback function as well as all the rowlist building functions. 

The bigtable reader function now takes the prefix and a slice of string slices. Each string slice contains all possible keys of one component. The reader will makes cross product of all the components.

The caller now process each row, extract the tokens for further processing. This results in simpler or similar code for the callers.


internal/store/bigtable/key.go
```diff
@@ -83,143 +79,37 @@ const (
...
-// BuildTriplesKey builds bigtable key for triples cache
-func BuildTriplesKey(dcids []string) bigtable.RowList {
-	rowList := bigtable.RowList{}
-	for _, dcid := range dcids {
-		rowList = append(rowList, fmt.Sprintf("%s%s", BtTriplesPrefix, dcid))
-	}
-	return rowList
+// Accessor represents data used to access bigtable row.
+type Accessor struct {
+	// import group table index.
```
Spaceenter, line 90: nit:  import -> Import

internal/server/v1/propertyvalues/state.go
```diff
@@ -71,29 +71,37 @@ func (s *state) init(
...
+		prefix = bigtable.BtPagedPropValIn
+	}
+	accs := []*bigtable.Accessor{}
 	for _, c := range cursorGroup.Cursors {
 		if c != nil {
-			rowList := buildSimpleRequestRowList(entity, property, arcOut, c.GetPage())
-			rowListMap[int(c.GetImportGroup())] = rowList
+			accs = append(accs, &bigtable.Accessor{
+				ImportGroup: int(c.GetImportGroup()),
+				Body:        [][]string{{entity}, {property}, {strconv.Itoa(int(c.GetPage()))}},
```
Spaceenter, line 83: nit: slight more readable if making this multiple lines:

```
[][]string{
  {
    ...
  },
  ...
}
```

Spaceenter (approved):

internal/server/v1/propertyvalues/state.go
shifucun: Done

internal/store/bigtable/key.go
shifucun: Done

shifucun (commented):
Thanks!

PR #848, opened 2022-05-18
Update ssl cert

shifucun:
The old manually created ssl has expired. @pradh created new cert and deployed to all mixer instances.


pradh (approved):
Thanks!

PR #849, opened 2022-05-19
Bulk paginated property values API

shifucun:
This generalizes the property values "state" struct to hold 2 dimensional data, keyed by "property" then by "entity". 

The pagination applies to each <property, entity> pair. The "nextIn()" and nextOut()" function are upgraded to consider all the <property, entity> pairs too.


proto/v1/property_values.proto
```diff
@@ -47,4 +47,32 @@ message PropertyValuesResponse {
...
+  // The pagination token for getting the next set of entries. This is empty
+  // for the first request and needs to be set in the subsequent request.
+  // This is the value returned from a prior call to PropertyValuesResponse
+  string next_token = 4;
+  // Direction can only be "in" and "out"
+  string direction = 5;
+}
+
+message BulkPropertyValuesResponse {
+  message OneResult {
```
Spaceenter, line 71: I feel we could use a better name? "OneResult" feels a bit too general?

e.g. EntityPropertyValues? See if other folks have better names...

internal/server/v1/propertyvalues/wrapper.go
```diff
@@ -0,0 +1,118 @@
...
+	limit := int(in.GetLimit())
+	token := in.GetNextToken()
+	direction := in.GetDirection()
+
+	// Check arguments
+	if property == "" {
+		return nil, status.Errorf(
+			codes.InvalidArgument, "missing required argument: property")
+	}
+	if direction != "out" && direction != "in" {
```
Spaceenter, line 88: nit: Vaguely remember that we had a discussion on using enums for "IN" and "OUT"?

Spaceenter (approved):

proto/v1/property_values.proto
shifucun: Done

internal/server/v1/propertyvalues/wrapper.go
shifucun: The enum proto won't work well here since the "out" and "in" is parsed as string from the uri. I have made these as constant string now.

PR #850, opened 2022-05-23
[V1 API] Implement paginated Triples API

shifucun:
This version is a huge improvement over the current triples API, in that:

- Allow pagination which makes it possible to retrieve full results
- Use a more organized response, making it easy to get result by property from in/out arcs

Also the implementation uses the property values cache instead of the triples cache. The triples cache can then be removed after V1 migration.


proto/v1/triples.proto
```diff
@@ -0,0 +1,35 @@
...
+message TriplesRequest {
+  string entity = 1;
+  // [Optional]
+  // The pagination token for getting the next set of entries. This is empty
+  // for the first request and needs to be set in the subsequent request.
+  // This is the value returned from a prior call to TriplesRequest
+  string next_token = 2;
+}
+
+message TriplesResponse {
```
Spaceenter, line 30: For discussions: Shall we make InTriples and OutTriples as 2 separate API calls?

shifucun: Now split out in and out triples.

internal/server/v1/triples/triples.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		return nil, err
+	}
+	for property := range data {
+		res.Data[property] = &pb.EntityInfoCollection{
+			Entities: data[property][entity],
+		}
+	}
+	if pi != nil {
+		var err error
```
Spaceenter, line 79: This line can be removed.

internal/server/v1/propertyvalues/state.go
```diff
@@ -220,18 +220,19 @@ func (s *outState) init(
 	return nil
 }
 
-func (s *state) getPagination() *pb.PaginationInfo {
-	res := &pb.PaginationInfo{}
+func (s *state) getPagination(direction string) *pb.PaginationInfo {
```
Spaceenter, line 223: Arg "direction" is not used in this function?

Spaceenter (changes requested):

Spaceenter (approved):

internal/server/v1/propertyvalues/state.go
shifucun: Good catch! updated.

internal/server/v1/triples/triples.go
shifucun: Done

shifucun (commented):
thanks for review!

PR #854, opened 2022-05-24
For Triples API, use v1 proxy v0 implementation

shifucun:
This allows removing the triples cache from Prophet and storage


shifucun:
thanks!

PR #856, opened 2022-05-24
[V1 API] Implement linked property values API

shifucun:
Changed the simple property values response outmost field name to "values" to be consistent with other API. We should only use "data" for bulk API, but specific field name for simple API.


internal/server/v1/propertyvalues/linked.go
```diff
@@ -0,0 +1,83 @@
...
+	valueEntityType := in.GetValueEntityType()
+	// Check arguments
+	if property != "containedInPlace" {
+		return nil, status.Errorf(
+			codes.InvalidArgument, "only support property 'containedInPlace'")
+	}
+	if !util.CheckValidDCIDs([]string{entity}) {
+		return nil, status.Errorf(
+			codes.InvalidArgument, "invalid entity %s", entity)
+	}
```
Spaceenter, line 45: also check if valueEntityType is empty?

Spaceenter (approved):

internal/server/v1/propertyvalues/linked.go
shifucun: Done

PR #857, opened 2022-05-24
[V1 API] bulk linked property values API

shifucun:


internal/server/v1/propertyvalues/linked.go
```diff
@@ -85,3 +85,73 @@ func LinkedPropertyValues(
...
+		return nil, err
+	}
+	result := &pb.BulkPropertyValuesResponse{}
+	for e, children := range resp {
+		tmp := &pb.BulkPropertyValuesResponse_EntityPropertyValues{
+			Entity: e,
+		}
+		for _, dcid := range children {
+			var name string
+			if tmp, ok := data["name"][dcid]; ok {
```
Spaceenter, line 144: There is already a "tmp" outside the for-loop scope, so use a different name here? Otherwise it'll be a bit confusing to read.

Also, instead of "tmp", could we have a more meaningful name?

Spaceenter (approved):

internal/server/v1/propertyvalues/linked.go
shifucun: Done. Use meaningful names in both locations.

PR #860, opened 2022-05-31
data release

n-h-diaz:


deploy/storage/bigquery.version
```diff
@@ -1 +1 @@
-datcom-store.dc_kg_2022_04_25_23_45_30
\ No newline at end of file
+datcom-store.dc_kg_2022_05_23_23_14_39
```
pradh, line 1: Could we revert the BQ version, since we've not yet gotten a full build since the bug from last week?

Screen Shot 2022-05-31 at 12 30 16 PM

pradh (approved):

deploy/storage/bigquery.version
n-h-diaz: done

shifucun (approved):

PR #864, opened 2022-06-16
Add ranking for USFEMA_NationalRiskIndex

enjoythecode:
Prefer `USDollar` over `FemaNationalRiskScore` for `Annual_ExpectedLoss_NaturalHazardImpact` (aggregate) and `Annual_ExpectedLoss_NaturalHazardImpact_*` (individual hazard types).

Could you please confirm that these changes would also rank `USDollar` over `FemaNationalRiskScore` for `Annual_ExpectedLoss_NaturalHazardImpact_AvalancheEvent`, etc.?

### DIFF
--- a/internal/server/ranking/ranking.go
+++ b/internal/server/ranking/ranking.go
@@ -79,6 +79,10 @@ var StatsRanking = map[RankKey]int{
 	// Expected data 2004-2010: 1393, 1379, 1297, 1400, 1381, 1015, 1092
 	{"FBIHateCrimePublications", "*", "*"}: 0, // FBI Hate Crime Publications
 	{"FBIHateCrime", "*", "*"}:             1, // FBI Hate Crime Aggregations
+	
+	// Prefer USDollar over Risk Score for Expected Annual Loss in FEMA National Risk Index (NRI)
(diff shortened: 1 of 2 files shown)


internal/server/ranking/ranking.go
```diff
@@ -79,6 +79,10 @@ var StatsRanking = map[RankKey]int{
 	// Expected data 2004-2010: 1393, 1379, 1297, 1400, 1381, 1015, 1092
 	{"FBIHateCrimePublications", "*", "*"}: 0, // FBI Hate Crime Publications
 	{"FBIHateCrime", "*", "*"}:             1, // FBI Hate Crime Aggregations
+	
+	// Prefer USDollar over Risk Score for Expected Annual Loss in FEMA National Risk Index (NRI)
+	{"USFEMA_NationalRiskIndex", "Annual_ExpectedLoss_NaturalHazardImpact*", "USDollar"} : 0, // Unit: USDollar
```
pradh, line 84: Oh I think the second and third args are mmethod and observationPeriod.  So to do what you want (prefer one unit over the other), we might need some more changes I think.

Let's ask @shifucun

shifucun: Right now we don't have ranking by unit yet, the definition in the early section of this file:

type RankKey struct {
	ImportName        string
	MeasurementMethod string
	ObservationPeriod string
}

So for this, you want to add a new field "Unit", add "*" to all the struct here, and make it into the comparison function below.

PR #866, opened 2022-07-01
Redesign ranking scoring algorithm

enjoythecode:
The previous implementation has exponential time (and code lines) complexity w.r.t to the number of fields of `RankKey`.

New implementation is only linear, because we can assume the number of `<RankKey, score>` mappings per import to be constant (currently, all import names have only 1 mapping, besides two of them which have 2 pairs each).

The new implementation is supported by extracting `ImportName` from RankKey to an outer key of `StatsRanking`. `ImportName` is used in every instance of `RankKey` and is the main axis of variability, so this retains the performance benefits of using a map.

### DIFF
--- a/internal/server/ranking/ranking.go
+++ b/internal/server/ranking/ranking.go
@@ -28,8 +28,8 @@ import (
 // Import Name should be non-empty.
 // Can use "*" for wildcard match for MeasurementMethod and ObservationPeriod
 type RankKey struct {
-	MeasurementMethod string
-	ObservationPeriod string
+    MeasurementMethod string
+    ObservationPeriod string
 }
 
 // StatsRanking is used to rank multiple source series for the same
@@ -84,7 +84,6 @@ var StatsRanking = map[string]map[RankKey]int{
     "FBIHateCrimePublications": {{"*", "*"}: 0,}, // FBI Hate Crime Publications
     "FBIHateCrime": {{"*", "*"}:             1,}, // FBI Hate Crime Aggregations
 }
-
 // BaseRank is the base ranking score for sources. If a source is prefered, it
 // should be given a score lower than BaseRank in StatsRanking. If a source is not
 // prefered, it should be given a score higher than BaseRank in StatsRanking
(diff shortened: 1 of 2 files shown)


internal/server/ranking/ranking.go
```diff
@@ -105,53 +108,59 @@ type CohortByRank []*pb.SourceSeries
...
+    importNameStatsRanking, ok := StatsRanking[importName]
+    if ! ok {
+        return BaseRank
+    }
+    scoreOfMostExactMatch := BaseRank
+    mostMatches := -1
+    for k, score := range importNameStatsRanking {
+        matches := 0
+        isMatch := true
+        if k.MeasurementMethod != "*" {
```
shifucun, line 121: if k.MeasurementMethod == "*", then should matches++?

internal/server/ranking/ranking.go
```diff
@@ -105,53 +108,59 @@ type CohortByRank []*pb.SourceSeries
 // score, otherwise can also match to wildcard options (indicated by *).
 //
 // If no entry is found, a BaseRank is assigned to the source series.
+func GetScoreRk(importName string, rk RankKey) int {
+    importNameStatsRanking, ok := StatsRanking[importName]
+    if ! ok {
+        return BaseRank
+    }
+    scoreOfMostExactMatch := BaseRank
```
shifucun, line 116: nit: scoreOfMostExactMatch -> rankScore

internal/server/ranking/ranking.go
```diff
@@ -105,53 +108,59 @@ type CohortByRank []*pb.SourceSeries
...
+                isMatch = false
+            }
+        }
+
+        isMoreExactMatch := matches > mostMatches
+        isSameExactnessMatchWithBetterScore := score < scoreOfMostExactMatch && matches == mostMatches
+        if isMatch && (isMoreExactMatch || isSameExactnessMatchWithBetterScore) {
+            scoreOfMostExactMatch = score
+            mostMatches = matches
+        }
```
shifucun, line 142: If I understand correctly, line 137:142 can be written as before:

if !isMatch {
  continue
}
if matches < mostMatches {
  continue
}
if matches == mostMatches && score > rankScore {
  continue
}
bestScore = score
mostMatches = matches

shifucun: Basically you want to use concise conditions and linear logic instead of complex conditions.

Also, I am not sure if we have the case of "matches == mostMatches && score < rankScore"

internal/server/ranking/ranking.go
shifucun: If you will handle this, and also make sure exact match is better than * match, then can increment * match by 0.9, and exact match by 1 (changing matches from int to float).

enjoythecode: No, we use "*" as a placeholder and do not want to increment matches for "*" fields. We use "*" as a placeholder and do not count it towards a match specifically because we want more exact matches to be scored higher.

Incrementing "*" matches would not change behavior because in case of at least 1 non-placeholder field, that gets matched and becomes the score. In case all RankKey fields are "*"; the first field will match because line 117: `mostMatches := -1`.

internal/server/ranking/ranking.go
enjoythecode: > Also, I am not sure if we have the case of "matches == mostMatches && score < rankScore"

Line 138: `isSameExactnessMatchWithBetterScore := score < scoreOfMostExactMatch && matches == mostMatches` handles this.

internal/server/ranking/ranking.go
enjoythecode: moving forward with changing `RankKey` fields to be string pointers and checking for nil-ness.

shifucun:
Thanks for the update! Looks much better now!

shifucun (approved):

PR #868, opened 2022-07-06
Format update for ranking.go

shifucun:

PR #869, opened 2022-07-06
Add Unit to Ranking and add ranking for FEMA NRI

enjoythecode:
Replaces #864
Merge after #868 (to include the formatting changes here; I did not duplicate that work here)

---

Add Unit as a field to ranking.go;
- Added field to RankKey
- Updated GetScoreRk with logic for the Unit field
- Updated GetScore* adapters to pass in the Unit as a pointer to RK

Add FEMA NRI ranking;
- Add to StatsRanking
- Add test in ranking_test.go

Fix nil pointer dereference bug where if the input RankKey to
GetScoreRk is nil, it would still get dereferenced if the StatsRanking
included that field;
- Check for input RK field non-nilness before referencing


shifucun (approved):

shifucun:
Nice!

PR #870, opened 2022-07-07
Periodic builds of Data Commons repositories

enjoythecode:
- build/ci/cloudbuild.allrepos.yaml is the entry point to the scripts
  in tools/periodic_builds. It is a Google Cloud Build job that will
  launch build jobs for all repos and notify users
- tools/periodic_builds/
  - README.md outlines usage
  - build_all_repos.sh sends build jobs for all repos (in parallel) and
    stores results in `/workspace` for the next step, which is;
  - notify_results.sh reads the build job output files from the previous
    step and sends out an email

TODO:
- email functionality is stubbed out with a printf because project
  access needs to be coordinated.

### DIFF
--- a/build/ci/cloudbuild.allrepos.yaml
+++ b/build/ci/cloudbuild.allrepos.yaml
@@ -1,0 +1,27 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


enjoythecode:
Moved to https://github.com/datacommonsorg/tools/pull/124

PR #871, opened 2022-07-11
Replaces bleve with sqlite FTS5.

nopper:


internal/server/statvar/statvar_match.go
```diff
@@ -16,39 +16,293 @@ package statvar
...
 
-func buildExplanationString(in *search.Explanation, sb *strings.Builder, level int) {
-	sb.WriteString(fmt.Sprintf("%.2f => %s\n", in.Value, in.Message))
-	if len(in.Children) > 0 {
-		for _, cchild := range in.Children {
-			for i := 0; i < level+1; i++ {
-				sb.WriteString(" ")
+const defaultLimit = 5
+
+const sqlCreateTables = `
```
shifucun, line 38: Can you add some comments for these commands?

shifucun: Maybe some higher level comments on how the sqlite search does?

shifucun (approved):
Can you add some comments for the rational for the change. Also what are the performance difference between the two? Is the index from sqlite smaller?

internal/server/statvar/statvar_match.go
nopper: Done.

PR #873, opened 2022-07-21
Data release: 2022-07-21

beets:


internal/server/statvar/golden/search_statvar/count_person.json
```diff
@@ -3899,106 +3999,6 @@
     {
       "name": "Population: 25 - 34 Years, Native Born Outside The United States",
       "dcid": "Count_Person_25To34Years_NativeBornOutsideTheUnitedStates"
-    },
```
beets, line 3902: Not sure if this truncation is expected..

internal/server/place/golden/get_locations_ranking/california.json
```diff
@@ -5,292 +5,292 @@
         "info": [
           {
             "rank": 1,
-            "value": 9943046,
+            "value": 10039107,
```
beets, line 8: This is from PEP 2019, and though we have 2020 data from ACS 5 year and Decennial. Previous number came from previous PEP import (PEP, P1Y). According to that, population did drop from 2019 to 2020.

https://autopush.datacommons.org/browser/geoId/06037?statVar=Count_Person

pradh: Probably because of the ranking here: https://github.com/datacommonsorg/mixer/blob/master/internal/server/ranking/ranking.go#L44-L46 ?
@shifucun 

India folks are updating PEP to have 2020.

internal/server/statvar/golden/search_statvar/count_person.json
pradh: I don't understand this.  There was no change to these imports (subject table S270*). My guess might be that these somehow went into Uncategorized vertical, but they do show under Demographics.

Also, the first time I went to the link below, I got nothing...

https://autopush.datacommons.org/tools/statvar#Count_Person_25To34Years_NoHealthInsurance

internal/server/place/golden/get_locations_ranking/california.json
shifucun: Can you update the ranking to be below and see it's getting better:

"CensusPEP": {{MM: s("CensusPEPSurvey"), OP: s("P1Y")}: 0}, // Population

internal/server/statvar/golden/search_statvar/count_person.json
pradh: Got nothing, I mean like: https://autopush.datacommons.org/tools/statvar#foo

@shifucun  you might have some insights... what cache does this use? does it do merge?

internal/server/place/golden/get_locations_ranking/california.json
shifucun: There are two CensusPEPSurvey import now: https://autopush.datacommons.org/browser/geoId/06?statVar=Count_Person

And we should prefer the one with "P1Y". @pradh after the data is in, I guess we can remove the other import?

pradh: Ah, that update is in.  So yes P1Y SG.

beets: @Spaceenter @shifucun should we prefer a more recent date for this if available?

shifucun: yes, the new CensPEPSurvey + P1Y would have 2020 data as well.

PR #875, opened 2022-07-25
Update V1 REST URI so entity dcid can be in the middle for better readability

shifucun:
From the feedback of initial users, it's not intuitive to separate out "entity" and "property" dcid in the request URI. Similarly, it's better to put "entity" before "variable" for observation API. The current design was due to the limit of gRPC transcoding that a multi-segment resource (with "/") can be only put in the end.

This PR improves it by programmatically checking the path in the API, rather than through the gRPC transcoding.


proto/mixer.proto
```diff
@@ -374,23 +374,23 @@ service Mixer {
   rpc PropertyValues(datacommons.v1.PropertyValuesRequest)
```
shifucun, line 374: @juliawu PTAL the URI pattern change here.

juliawu: LGTM!

juliawu (approved):
Great change! Should be much more intuitive for users now :)

PR #880, opened 2022-08-03
update GetStatVarGroup and GetStatVarGroupNode place -> entity

n-h-diaz:
wanted to double check that it's okay to make this change? 
(I'll update website next to switch places -> entities for this API)


shifucun (approved):
Thanks for the update!

proto/stat_var.proto
```diff
@@ -180,16 +181,16 @@ message StatVarGroupNode {
...
   // The stat var or stat var group ID to query for.
   string stat_var_group = 1;
-  // The places that the stat var group is associated with. The response should
-  // only contain the children stat var (group) if it has data for at least one
-  // of the places.
-  repeated string places = 2;
+  // The entities that the stat var group is associated with. The response
+  // should only contain the children stat var (group) if it has data for at
+  // least one of the entities.
+  repeated string entities = 2;
```
pradh, line 193: To confirm, stanford doesn't use this API, right?  They wanted to use some SV hierarchy APIs in the past.

shifucun: They don't have public traffic. This is the requests in the past 30 days https://screenshot.googleplex.com/9kihiGV3wFwYmag

pradh: Great, thanks for checking!

n-h-diaz: thank you both :)

PR #881, opened 2022-08-03
Require API key for V1 APIs

shifucun:


Fructokinase:
I don't have an explicit list of all v0 apis. LGTM assuming that the explicitly listed apis in the templates in this change are all the v0 apis.

Fructokinase (approved):

pradh (approved):

PR #882, opened 2022-08-06
Use typed property value cache for entity related APIs

shifucun:
The cache merging now happens at <entity, property, type> level. For many data structures, an additional level of map keyed by "type" is added.

To dedup the entities with multiple types, the code checks if any other type has processed the current entity. This processing could happen in previous page or in current page.

The meaning of limit changed for pagination now. It restricts to "per type limit" instead of the total count.


internal/server/v1/propertyvalues/core.go
```diff
@@ -34,50 +35,55 @@ import (
...
 	} else {
 		pi, err := pagination.Decode(token)
 		if err != nil {
-			return nil, nil, status.Errorf(
-				codes.InvalidArgument, "invalid pagination token: %s", token)
+			return nil, nil, status.Errorf(codes.InvalidArgument, "invalid pagination token: %s", token)
 		}
 		cursorGroups = pi.CursorGroups
 	}
 	if limit == 0 || limit > defaultLimit {
```
Spaceenter, line 64: nit: To make it more error-prone:

if limit <=0 || limit > defaultLimit {

?

Spaceenter (approved):

internal/server/v1/propertyvalues/core.go
shifucun: Done

shifucun:
@pradh I will submit this for now and feel free to review later when you are back.

PR #889, opened 2022-08-23
Add a loadtesting tool based on BulkObservationsSeriesLinked rpc

Fructokinase:
Adding the tool I used to loadtest some of the APIs. I think it will be useful for future testing. 

The easiest way to loadtest apis that are not exposed to the internet was to deploy the locust pod and specify the k8s service name as the host.

Below is a screenshot of the locust UI. 

Screen Shot 2022-08-23 at 8 39 14 AM


shifucun:
> * Scaled replicas need to build the initial index (another ~10sec).

Only need to build the index for stat var search API (3 or 4 APIs which can be grouped as one serice). So this won't be an issue for other services.

tools/loadtesting/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+This tool is meant for loadtesting mixer APIs in live clusters. It is based on [locust](https://docs.locust.io/en/stable/).
+
+The specific APIs that are called are defined by a [locustfile](https://docs.locust.io/en/stable/writing-a-locustfile.html), located under `tools/loadtesting/locustfiles`. 
+
+## Usage
+
+If the mixer instance to be tested lives as a k8s service that is not exposed to the internet, we can run a containerized version in the same cluster.
+
+1.  Deploy the pod into the same namespace as Mixer.
```
shifucun, line 11: mention set the context to the cluster first?

tools/loadtesting/README.md
```diff
@@ -0,0 +1,51 @@
...
+| Spawn rate  | Number of users added per second. This value is recommended to be <5. |
+| Host  | http:// + result of `kubectl -n <mixer namespace> get svc`. | 
+
+## Making changes to this tool
+
+1.  Set up the local environment.
+
+`virtualenv env && source env/bin/active && pip install -r requirements.txt`
+
+2.  After desired changes (most likely edit/add of a locustfile), you may want to change the default locust file inside `Dockerfile`.
```
shifucun, line 45: is it possible to not building a new image each time, but control it in the UI? So ideally the locustfile would have all tests..

tools/loadtesting/README.md
Fructokinase: Done.

tools/loadtesting/README.md
Fructokinase: Unfortunately that's not provided out of the box. We can always make local changes and launch this tool locally, but it can't be used to loadtest k8s services that are not exposed to the internet.

shifucun:
Please fix lint errors.

shifucun (approved):

PR #892, opened 2022-08-25
Fix v0 endpoint rule selector issue

shifucun:
From https://cloud.google.com/endpoints/docs/grpc/grpc-service-config#rules_and_selectors, wildcard "*" can only be in the last part of a selector.

Updated to replace the domain in deploy_gke.sh

### DIFF
--- a/scripts/deploy_gke.sh
+++ b/scripts/deploy_gke.sh
@@ -84,7 +84,6 @@ kubectl apply -f kustomize-build.yaml
 
 # Deploy Cloud Endpoints
 cp $ROOT/esp/endpoints.yaml.tmpl endpoints.yaml
-echo $DOMAIN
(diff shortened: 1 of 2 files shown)


scripts/deploy_gke.sh
```diff
@@ -84,6 +84,8 @@ kubectl apply -f kustomize-build.yaml
 
 # Deploy Cloud Endpoints
 cp $ROOT/esp/endpoints.yaml.tmpl endpoints.yaml
+echo $DOMAIN
```
beets, line 87: remove this before merging?

beets (approved):
thanks for the fix!

scripts/deploy_gke.sh
shifucun: Done

Fructokinase (approved):

PR #893, opened 2022-08-25
Fix misconfiged API endpoints selector

shifucun:
The selector needs to use the "apis.name" not "endpoints.name" field


Spaceenter (approved):

shifucun (commented):
thanks Wei!

PR #895, opened 2022-08-25
Update mixer developer guide

shifucun:
- Prefer local go server vs docker
- Add tags for fts5

### DIFF
--- a/internal/server/statvar/statvar_match.go
+++ b/internal/server/statvar/statvar_match.go
@@ -209,7 +209,7 @@ func BuildSQLiteIndex(
 	for index, doc := range docSet {
 		_, err = stmt.Exec(index, doc.ID, doc.Title, doc.Signature, doc.NumConstraints, doc.KeyValueText)
 		if err != nil {
-			fmt.Printf("Ignoring statvar with DCID=%s as its signature is not unique.\n", doc.ID)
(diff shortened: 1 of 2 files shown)


pulkit-s (approved):

internal/server/statvar/statvar_match.go
```diff
@@ -209,7 +209,7 @@ func BuildSQLiteIndex(
 	for index, doc := range docSet {
 		_, err = stmt.Exec(index, doc.ID, doc.Title, doc.Signature, doc.NumConstraints, doc.KeyValueText)
 		if err != nil {
-			fmt.Printf("Ignoring statvar with DCID=%s as its signature is not unique.\n", doc.ID)
+			// fmt.Printf("Ignoring statvar with DCID=%s as its signature is not unique.\n", doc.ID)
```
shifucun, line 212: @nopper I am seeing a lot of error logs from this. Is that expected?

nopper: I am not sure. It looks like multiple statvars have the same set of of constraints and k/v properties. Is this expected?
The signature is computed here:

https://github.com/shifucun/mixer/blob/60da88e6693a1cf4253c25549567b32dad2efce7/internal/server/statvar/statvar_match.go#L169

PR #897, opened 2022-09-02
update dcbranch version

Spaceenter:


shifucun (approved):
thanks!

pradh (approved):

test/setup.go
```diff
@@ -106,7 +106,7 @@ func setupInternal(
 		log.Fatalf("failed to create Bigquery client: %v", err)
 	}
 
-	branchTableName := "dcbranch_2022_03_06_18_18_49"
+	branchTableName := "dcbranch_2022_08_31_18_28_28"
```
pradh, line 109: Wonder if we should use a "pinned" version?  Because when GC happens, this will fail in future...

shifucun: I realized as we have newer non-branch cache, this branch cache is useless in the testing anymore. Let me try to send a PR to remove it and see if that works

Spaceenter: OK - thanks @shifucun -- I'll deprecate this PR then.

PR #904, opened 2022-09-07
Mixer API split - part 3: Add env speicific values yaml files

Fructokinase:
This is a follow up from [part 1](https://github.com/datacommonsorg/mixer/pull/902).

This PR migrates [overlaps](https://github.com/datacommonsorg/website/tree/master/deploy/overlays) to value yaml files used by the Helm chart.

- local instance is excluded.
- recon instances are excluded because they are marked to be deleted.
- custom folder is excluded because it is a template.

### DIFF
--- a/deploy/charts/envs/mixer_private.yaml
+++ b/deploy/charts/envs/mixer_private.yaml
@@ -1,0 +1,122 @@
+# Helm config
+mixer:
+  gcpProjectID: datcom-mixer-statvar
+  serviceName: mixer.endpoints.datcom-mixer-statvar.cloud.goog
+
+  useTMCFCSVData: true
+  tmcfCSVBucket: datcom-public
+  tmcfCSVFolder: food
+
+memdbJSON: |
+  {
+    "importName": "Feeding America",
+    "provenanceUrl": "https://www.feedingamerica.org/",
+    "dataDownloadUrl": "https://www.feedingamerica.org/",
+    "rootSvg": "g/Feeding_America",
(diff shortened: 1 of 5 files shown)


deploy/charts/envs/mixer_private.yaml
```diff
@@ -0,0 +1,122 @@
+# Helm config
+mixer:
+  gcpProjectID: datcom-mixer-statvar
+  serviceName: mixer.endpoints.datcom-mixer-statvar.cloud.goog
+
+  useTMCFCSVData: true
+  tmcfCSVBucket: datcom-public
+  tmcfCSVFolder: food
+
+memdbJSON: |
```
shifucun, line 10: If the file gets larger, is there a way to load it from a json file as before?

Fructokinase: yes, for example, the mcfs can be set like the following when calling helm.

--set-file mixer.configmaps."encode.mcf"=deploy/mapping/encode.mcf

Value of mixer.configmaps.encode.mcf is set from deploy/mapping/encode.mcf

Thinking out loud, I think for large files, we should consider loading from gcs instead.
(current deployment yaml  "--memdb_path=/datacommons/memdb")

shifucun: Got it. These files won't be too large, but maybe too long (hundreds of lines) to set inline in yaml

shifucun (approved):

PR #905, opened 2022-09-08
Consolidate stat var (group) related API

shifucun:
* Change VariableGroups API into BulkVariableGroupInfo API. This is to be consistent with VariableGroupInfo API and Variables API. Basically, "info" API returns the meta info of a node, where "*s" API returns a list of queried nodes.

* Change */variable/group/* to */variable-group/* as this has been confusing when coexist with */variables/* and */variable/ancestor/*. Now it's clear if the API is about an SV or SVG.


proto/v1/info.proto
```diff
@@ -60,4 +60,19 @@ message VariableGroupInfoRequest {
   // only contain the child variable (group) if it has data for at least one
   // of the entities.
   repeated string constrained_entities = 2;
+}
```
pradh, line 63: For my understanding, is there no VariableGroupInfoResponse?

pradh (approved):

proto/v1/info.proto
```diff
@@ -60,4 +60,19 @@ message VariableGroupInfoRequest {
...
+
+
+message BulkVariableGroupInfoRequest {
+  // The variable group dcids to query for.
+  // If not specified, then return all the variable group info.
```
pradh, line 68: nit: If we're using "(optional)" convention, use for this field too.

proto/v1/info.proto
shifucun: It returns StatVarGroupNode directly

proto/v1/info.proto
shifucun: Done

proto/v1/info.proto
pradh: That feels a bit inconsistent with other calls, like VariableInfoResponse, which return node + response?

Also, if we have a VariableGroupInfoResponse, perhaps BulkVariableGroupInfoResponse (like other responses) be a repeated of the former?

shifucun: That's a good point. I was hoping to keep original format but good to make it right at this round.

PR #906, opened 2022-09-09
Mixer API split - part 4 : Add a recon service group

Fructokinase:
Add a recon service group which will only serve Recon services (currently all service groups serve recon services).

Tested on dev instance

```
helm upgrade --install mixer-dev deploy/helm_charts/mixer \
  --atomic \
  -f deploy/helm_charts/envs/mixer_Dev.yaml \
  --set mixer.image.tag=45bd8bb \
  --set mixer.githash=45bd8bb \
  --set-file mixer.schemaConfigs."base\.mcf"=deploy/mapping/base.mcf \
  --set-file mixer.schemaConfigs."encode\.mcf"=deploy/mapping/encode.mcf \
  --set-file mixer.schemaConfigs."dailyweather\.mcf"=deploy/mapping/dailyweather.mcf \
  --set-file mixer.schemaConfigs."monthlyweather\.mcf"=deploy/mapping/monthlyweather.mcf \
  --set-file kgStoreConfig.bigqueryVersion=deploy/storage/bigquery.version \
  --set-file kgStoreConfig.bigtableImportGroupsVersion=deploy/storage/bigtable_import_groups.version
```

All deployments were successful

https://pantheon.corp.google.com/kubernetes/workload/overview?mods=-monitoring_api_staging&project=datcom-mixer-dev-316822

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -118,13 +118,6 @@ func main() {
 	// Create grpc server.
 	srv := grpc.NewServer(opts...)
 
-	// Metadata.
-	metadata, err := server.NewMetadata(
-		*bqDataset, *storeProject, branchBtInstance, *schemaPath)
-	if err != nil {
-		log.Fatalf("Failed to create metadata: %v", err)
-	}
-
 	branchCachePubsubTopic := "proto-branch-cache-reload"
 	branchCacheVersionFile := "latest_proto_branch_cache_version.txt"
 
@@ -196,6 +189,13 @@ func main() {
 			if err != nil {
 				log.Fatalf("Failed to create cache: %v", err)
 			}
+		}
+
+		// Metadata.
+		metadata, err := server.NewMetadata(
+			*bqDataset, *storeProject, branchBtInstance, *schemaPath)
+		if err != nil {
+			log.Fatalf("Failed to create metadata: %v", err)
 		}
 
 		// Create server object
--- a/deploy/helm_charts/envs/mixer_prod.yaml
+++ b/deploy/helm_charts/envs/mixer_prod.yaml
@@ -23,6 +23,18 @@ serviceGroups:
       resources:
         memoryRequest: "8G"
         memoryLimit: "8G"
+  - recon: 
+      urlPaths:
+        - "/entity/*"
+        - "/coordinate/*"
+        - "/id/*"
+      replicas:
+        default: 4
+        min: 4
+        max: 8
+      resources:
+        memoryRequest: "3G"
+        memoryLimit: "3G"
   - mixer-default-service:
       urlPaths:
         - "/*"
--- a/deploy/helm_charts/mixer/templates/deployment.yaml
+++ b/deploy/helm_charts/mixer/templates/deployment.yaml
@@ -71,13 +71,18 @@ spec:
             requests:
               memory:  {{ .resources.memoryRequest }}
           args:
+            # Import group tables are need for both Mixer and Recon services.
+            - --import_group_tables=$(IMPORT_GROUP_TABLES)
+            {{- if eq $serviceName "recon"}}
+            - --serve_recon_service=true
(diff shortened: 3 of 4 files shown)


Fructokinase:
@pradh Once this change reaches prod, entity resolve API can be reached by `https://api.datacommons.org/entity/resolve`. Then all recon projects can be decommissioned.

Trying to find all the places that call recon API. So far I only see API_ROOT from 
import repo:`util/src/main/java/org/datacommons/util/ExternalIdResolver.java` referring to the recon service. 

Can I get some help with finding out all the places where recon services are called please?

pradh:
> @pradh Once this change reaches prod, entity resolve API can be reached by `https://api.datacommons.org/entity/resolve`. Then all recon projects can be decommissioned.
> 
> Trying to find all the places that call recon API. So far I only see API_ROOT from import repo:`util/src/main/java/org/datacommons/util/ExternalIdResolver.java` referring to the recon service.
> 
> Can I get some help with finding out all the places where recon services are called please?

DC import tool:  https://github.com/datacommonsorg/import/search?q=recon.datacommons.org
Data repo:  https://github.com/datacommonsorg/data/search?q=recon.datacommons.org

cmd/main.go
```diff
@@ -198,6 +191,13 @@ func main() {
 			}
 		}
 
+		// Metadata.
+		metadata, err := server.NewMetadata(
```
shifucun, line 195: why is the change of this code section position?

deploy/helm_charts/envs/mixer_prod.yaml
```diff
@@ -23,6 +23,18 @@ serviceGroups:
...
+      urlPaths:
+        - "/entity/*"
+        - "/coordinate/*"
+        - "/id/*"
+      replicas:
+        default: 4
+        min: 4
+        max: 8
+      resources:
+        memoryRequest: "3G"
```
shifucun, line 36: @Spaceenter do you expect the recon API uses high memory? Or we can reduce the memory but increase the number of pods?

deploy/helm_charts/mixer/templates/deployment.yaml
```diff
@@ -71,13 +71,18 @@ spec:
...
+            # Import group tables are need for both Mixer and Recon services.
+            - --import_group_tables=$(IMPORT_GROUP_TABLES)
+            {{- if eq $serviceName "recon"}}
+            - --serve_recon_service=true
+            - --serve_mixer_service=false
+            {{ else }}
             - --mixer_project=$(MIXER_PROJECT)
             - --store_project=$(STORE_PROJECT)
             - --bq_dataset=$(BIG_QUERY)
-            - --serve_recon_service=true
```
shifucun, line 77: should the recon conditions in its own if else clause? like

```
{{- if eq $serviceName "recon"}}
- --serve_recon_service=true
- --serve_mixer_service=false
{{ else }}
- --serve_recon_service=false
- --serve_mixer_service=true
{{end}}
```

Fructokinase: yes, that sees much more clear, fixed. Thanks!

cmd/main.go
Fructokinase: Was trying to separate out the arguments for mixer vs recon services in depoyment.yaml template. schema_path is not needed for recon but without it, recon containers were throwing errors. Moved metadata declaration to mixer only, also closer to where it is first used.

shifucun: Got it

shifucun (approved):

deploy/helm_charts/envs/mixer_prod.yaml
Fructokinase: As discussed offline with Bo, will keep this at 3G for now and observe what happens.

PR #907, opened 2022-09-09
Add golden test for variable group info APIs

shifucun:


internal/server/v1/info/golden/bulk_variable_group_info/economy.json
```diff
@@ -0,0 +1,27 @@
...
+        ],
+        "childStatVarGroups": [
+          {
+            "id": "dc/g/EconomicActivity",
+            "specializedEntity": "Economic Activity"
+          }
+        ]
+      }
+    },
+    {}
```
juliawu, line 25: For my own understanding, the expected behavior is just to return empty {} for the node "" instead of throwing an error. Is that right?

internal/server/v1/info/variable_group.go
```diff
@@ -52,9 +52,16 @@ func BulkVariableGroupInfo(
...
 ) (*pb.BulkVariableGroupInfoResponse, error) {
-	// TODO(shifucun): handle "nodes" in request, now always return all the
-	// varible group data.
+	// TODO (shifucun):
+	// The response here is directly from the raw SVG group, not consistent
+	// with the VariableGroupInfo() API.
+	// Ideally, both APIs need to filter out the child variable (group) that has
+	// no data, but this is indicated with a "has_data" field, to
+	// accomocate the "Show all statistical variables" in UI widget. The UI
+	// should call this API twice, w/o constrained_entities to achieve that.
```
juliawu, line 61: For my own understanding, for now VariableGroupInfo does filter out child variable groups with no data, but BulkVariableGroupInfo does not. Is that  correct?

juliawu (approved):
LGTM!

internal/server/v1/info/golden/bulk_variable_group_info/economy.json
shifucun: I updated it to be a non-empty string now. So for input without information, it returns the node id only.

internal/server/v1/info/variable_group.go
shifucun: Yes, the behavior is not consistent now. I would like both to behave like the documentation.

PR #908, opened 2022-09-09
Add support for returning a category list (for the category menu bar …on place explorer) to the place page API

pulkit-s:
Context: b/206032102

Related prophet CL : cl/473084606


proto/stat.proto
```diff
@@ -410,4 +411,8 @@ message GetStatDateWithinPlaceRequest {
 message GetStatDateWithinPlaceResponse {
   // Keyed by stat var.
   map<string, StatDateList> data = 1;
-}
\ No newline at end of file
+}
+
+message Categories {
+  repeated string category = 1;
+}
```
shifucun, line 418: should this also be in place.proto?

internal/server/v0/internalplace/golden/get_place_page_data/asm.sample.json
```diff
@@ -25629,6 +25629,14 @@
       }
     }
   },
+  "validCategories": {
```
shifucun, line 25632: Why are these all empty?

pulkit-s: Because we don't have a category in the request for these test cases. So this uses the old cache. The file ca_economics.sample.json has a category included and we can see desired behavior in that file.

proto/stat.proto
pulkit-s: The response object used right now is in internal.proto, so have changed GetPlacePageDataResponse there.  Is there a specific place that you have in mind in place.proto that we should change?

shifucun: Yea, maybe this can also be in internal.proto. "stat.proto" should contain proto for stat related API. The folder is more or less based on the API functional grouping.

internal/server/v0/internalplace/golden/get_place_page_data/asm.sample.json
shifucun: That makes sense, thanks!

proto/stat.proto
pulkit-s: Done

proto/stat.proto
```diff
@@ -134,6 +134,7 @@ message PlaceStat {
 message StatVarObsSeries {
```
pradh, line 134: Is this equivalent to http://google3/datacommons/proto/prophet/cache_data.proto;l=280-286;rcl=473084606 ?

I wonder why we don't call it LandingPageData?

pradh (approved):

proto/stat.proto
shifucun: We shouldn't need this new field here?

pulkit-s: Do you mean the categories field @shifucun ?

We need it because the LandingPageData object which holds the cache data from prophet has a categories field.

shifucun (approved):
thanks for the update!

PR #909, opened 2022-09-12
Mixer API split - part 5 : Switch over to helm in deploy_gke.sh

Fructokinase:
Switch over from kustomize to helm.

Changes:
- Change kustomize commands to helm commands in deploy_gke.sh. After this PR is merged, changes to the clusters will be based on deploy/helm_charts/envs/<env file>.yaml rather than overlays.
- Create a binary before each deployment, and bind the created image to what is deployed ass opposed to always using latest. Tags are easily overwritten and fixing tags to releases stabilizes this.
- Small fix on backend config template: make it release specific.

Tested:
./scripts/deploy_gke.sh mixer_dev 
./scripts/deploy_gke.sh mixer_autopush

### DIFF
--- a/scripts/deploy_gke.sh
+++ b/scripts/deploy_gke.sh
@@ -80,14 +80,14 @@ gcloud container clusters get-credentials $CLUSTER_NAME --region $REGION
 RELEASE=${ENV//_/-}
 
 # Create a release specific image for the deployment.
-./scripts/push_binary.sh $TAG
+./scripts/push_binary.sh "$TAG"
 
 # Upgrade or install Mixer helm chart into the cluster
-helm upgrade --install $RELEASE deploy/helm_charts/mixer \
+helm upgrade --install "$RELEASE" deploy/helm_charts/mixer \
   --atomic \
-  -f deploy/helm_charts/envs/$ENV.yaml \
-  --set mixer.image.tag=$TAG \
-  --set mixer.githash=$TAG \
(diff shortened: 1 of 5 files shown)


scripts/deploy_gke.sh
```diff
@@ -64,23 +64,36 @@ if [[ $ENV == "autopush" ]]; then
...
-cp kustomization.yaml kustomize-deployed.yaml
 gcloud config set project $PROJECT_ID
 gcloud container clusters get-credentials $CLUSTER_NAME --region $REGION
-kubectl apply -f kustomize-build.yaml
+
+# Change "mixer_prod" for example, to "mixer-prod"
+RELEASE=${ENV//_/-}
+
+# Create a release specific image for the deployment.
+./scripts/push_binary.sh $TAG
```
shifucun, line 83: is this needed, at this stage, a version with $TAG should already exist?

shifucun:
Please also fix Codacy error

scripts/deploy_gke.sh
Fructokinase: It didn't when I was running locally. Sometimes I forget to run push_binary.sh beforehand, thought I'd just add it here. But it should only run if image does not exist, added that condition. Thanks!

shifucun (approved):
Thanks for the update. Please follow closely on this week's release with the new process.

PR #917, opened 2022-09-14
Merge categories across import groups to get a complete set of categories

pulkit-s:
b/206032102

Also add a function to remove duplicates from the list of categories.


internal/server/placepage/place_page.go
```diff
@@ -258,6 +258,9 @@ func fetchBtData(
 			placePageData := row.Data.(*pb.LandingPageCache)
 			if _, ok := mergedPlacePageData[place]; !ok {
 				mergedPlacePageData[place] = placePageData
+			} else {
+				mergedPlacePageData[place].Categories = append(
```
shifucun, line 262: You can use util.MergeDedupe() function in util.go

pulkit-s: Done.

internal/server/placepage/place_page.go
```diff
@@ -275,8 +279,8 @@ func fetchBtData(
 	for place, data := range mergedPlacePageData {
 		finalData := &pb.StatVarSeries{Data: map[string]*pb.Series{}}
 		categoryData[place] = &pb.Categories{}
-		categoryData[place].Category = append(categoryData[place].Category,
-			data.Categories...)
+		categoryData[place].Category = util.MergeDedupe(
+			categoryData[place].Category, data.Categories)
```
shifucun, line 283: are these still needed, or can just use data.Categories?

pulkit-s: Done. Added a sort call to ensure that the output is sorted.

internal/server/placepage/place_page.go
```diff
@@ -275,8 +279,8 @@ func fetchBtData(
 	for place, data := range mergedPlacePageData {
 		finalData := &pb.StatVarSeries{Data: map[string]*pb.Series{}}
 		categoryData[place] = &pb.Categories{}
-		categoryData[place].Category = append(categoryData[place].Category,
-			data.Categories...)
+		categoryData[place].Category = data.Categories
+		sort.Strings(categoryData[place].Category)
```
shifucun, line 283: Oh, so MergeDedupe() already sorts the result. So can just do 

```
cateogryData[place] = &pb.Categories{Category: data.Categories}
```

shifucun (approved):

internal/server/placepage/place_page.go
pulkit-s: The case where I saw an issue was where only one import group had categories for a place. In that case, because the code never passes through merge dedupe, it would cause output to be unsorted. I can move the sort to the if block to counter this though, if you think that would be better.

shifucun: Maybe can remove the "else" condition in line 262, so it always do MergeDedup

shifucun (approved):

internal/server/placepage/place_page.go
pulkit-s: Makes sense, done.

shifucun:
thanks for the update!

shifucun (approved):

PR #920, opened 2022-09-15
Add /v1 to VariableAncestors URL; Fix selector for GetEntityStatVarsUnionV1

shifucun:


Fructokinase:
Should we also delete "/variable/ancestors/*" from values.yaml, and mixer_<prod|autopush|staging>.yaml?

Fructokinase (approved):
Should we also delete "/variable/ancestors/*" from values.yaml, and mixer_<prod|autopush|staging>.yaml?

shifucun:
> Should we also delete "/variable/ancestors/*" from values.yaml, and mixer_<prod|autopush|staging>.yaml?

Good catch. Done

PR #922, opened 2022-09-20
Update cache data versions

shifucun:
All diffs looks reasonable to me. Would be good if the provenance id is stable.


pradh:
> All diffs looks reasonable to me. Would be good if the provenance id is stable.

Ah sorry, didn't get around to it last week.  Added it to my todo this week...

pradh (approved):

pradh:
Adding a few more pairs of eyes :)

deploy/storage/bigtable_import_groups.version
```diff
@@ -1,4 +1,4 @@
-frequent_2022_08_24_01_14_28
-ipcc_2022_08_15_21_36_44
-biomedical_2022_08_23_09_33_09
-infrequent_2022_09_08_19_08_12
+frequent_2022_09_19_01_13_28
+ipcc_2022_09_01_03_46_38
```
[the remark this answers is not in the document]
shifucun: Yes, this is for catching any other issues, which do not exist :)

internal/server/place/golden/get_locations_ranking/california.json
```diff
@@ -5,292 +5,292 @@
         "info": [
           {
             "rank": 1,
-            "value": 9829544,
+            "value": 10039107,
```
beets, line 8: This is reverting to Census PEP or ACS 1 year from 2019 instead of PEP P1Y from 2021 or 5 year from 2020 which are also available.

https://screenshot.googleplex.com/9Nk8QNDvqo82WmY

shifucun: @pradh @Spaceenter is this related to any Flume change?

pradh: Nice catch!

The import with the new dataset is USCensusPEP_Annual_Population. The triggering change was likely cl/471904782 which removed USCensusPEP_Annual_Population from frequent.  As a result, I think we moved the final ranking from Mixer to Flume.

Two issues I find:
(1) The ranking config differs between [Mixer](https://github.com/datacommonsorg/mixer/blob/master/internal/server/ranking/ranking.go#L44) and [pipeline](http://google3/datacommons/prophet/flume_generator/ranking_helper.cc;l=50;rcl=475182259).  That is, the change to add USCensusPEP_Annual_Population is missing in flume.

(2) The ranking uses a side-table of population stats (since not every import-group will have population data), and that might need to be refreshed [here](https://g3doc.corp.google.com/datacommons/prophet/related_places/README.md?cl=head).

These should be fixed, but neither of these I think explain the diff, because the ranking pipeline uses an older custom ranking based just on the dates.  Might need to try running with these to debug further...

pradh: Longer term Q: The ranking pipeline adds a bunch of complexity with import-groups, and it's worth thinking whether we should deprecate it in favor of computing rankings at runtime.

shifucun: Should we rebuild all the cache with the updated ranking config?

We should evaluate the latency for computing ranking at runtime, especially for the place page overview ranking table. The API traffic routing allows us to independently scale APIs, which gives more room to speed up. @Fructokinase

beets: +1 for computing ranking at runtime! It will also allow us to do it for many more variables.

In the meantime, can we put a big note around both configs in (1) to update the other? A manual IFTT :)

pradh: I think I understand the issue now (repro'd with cl/475707058), this is a third issue.

(3) The series choice in the ranking pipeline predates [prophet](http://google3/datacommons/prophet/flume_generator/ranking_helper.cc;l=266;rcl=475182259), and uses a heuristic a different heuristic:  within a series, it picks the "latest overlapping date" across all comparable places, and then compares this date across different series.  For the USCensusPEP_Annual_Population import, there are a few counties whose latest date is 1999 or 2009 ([screenshot](https://screenshot.googleplex.com/A65tfgJbAEdpuXQ)).  That demotes this series.

We should make the series selection logic in the ranking pipeline consistent with prophet+mixer, but it will be a non-trivial change and perhaps not worth making if we're going to deprecate the pipeline altogether.

This is what I'm considering doing:
a) Create a tracking bug to deprecate ranking pipeline, or if we decide it needs to stay, fix series selection to be consistent with Prophet+Mixer.
b) Move USCensusPEP_Annual_Population to "frequent", so Mixer will prefer that over infrequent (as it was before). This should fix Issue (3).
c) Send out CLs for fixing issues (1) and (2) (which don't actively cause any known issue)
d) Rebuild frequent (will happen as part of nightly build)

Let me know if you think otherwise...

shifucun: Sg for the analysis and fixes! Look forward to understanding the complexity of ranking at run-time.

pradh: Frequent got built, and it restores old stats.

Confirmed via `./check_bt 'd/q/*^County^geoId/06^Count_Person' frequent | ./cache_parse`

But autopush Mixer still seems to prefer infrequent, per result in

`curl -X POST 'https://autopush.api.datacommons.org/node/ranking-locations' -d '{ "place_type": "County", "within_place": "geoId/06", "stat_var_dcids": ["Count_Person"] }'`

Any idea @shifucun ?

internal/server/v0/statvarpath/golden/get_statvar_path/school.json
```diff
@@ -1,6 +1,6 @@
 {
   "path": [
     "dc/g/Person_EducationalAttainment",
-    "dc/g/Education"
+    "dc/g/Demographics"
```
shifucun, line 4: @pradh is this expected?

internal/server/place/golden/get_locations_ranking/california.json
shifucun: There is some issue with mixer deployment, but the data looks good after I use the latest version in this PR

internal/server/v0/statvarpath/golden/get_statvar_path/school.json
pradh: On autopush it shows up under both Demographics and Education?

internal/server/place/golden/get_locations_ranking/california.json
pradh: Sounds good, thanks.

internal/server/v0/statvarpath/golden/get_statvar_path/school.json
shifucun: autopush has some deployment issue so it's stale. For now we should rely on this PR..

pradh: what cache is this?

pradh: SVG hierarchy cache in frequent also has Person_EducationalAttainment under both Demographics and Education.

`./check_bt 'd/1' frequent | ./cache_parse > /tmp/svg.txt`

shifucun: I see, in BigQuery it is only specializationOf dc/g/Education

https://screenshot.googleplex.com/3Rgv9Ar6zVvPjxr

pradh: Should we roll this to staging and validate?

shifucun: sg, I will merge this PR and tested on staging

PR #924, opened 2022-09-22
Fix triples API: add observation result and update bulk API signature

shifucun:


internal/server/v1/triples/golden/bulk_triples_in/result1.json
```diff
@@ -4,7 +4,7 @@
       "node": "Country",
       "triples": {
         "rangeIncludes": {
-          "entities": [
+          "nodes": [
```
pradh, line 7: To confirm, this API is not used by anyone?

pradh (approved):

internal/server/v1/triples/golden/bulk_triples_in/result1.json
shifucun: Nope, this is not used.

PR #926, opened 2022-09-22
Fix autopush config, also add debug flag for helm

Fructokinase:
This is a follow up for https://github.com/datacommonsorg/mixer/pull/925

Autopush had too much replicas, resulting in a scheduling failure. Add debug flag to helm to show the exact issues from now on.


Fructokinase:
Tested locally with ./scripts/deploy_gke.sh mixer_autopush

shifucun (approved):

PR #927, opened 2022-09-23
Change facet id from uint32 to string

shifucun:
JSON does not allow to use integer as key. The response changes facet id to string when it is key of map field, which creating inconsistency.

This PR explicitly make facet id as string.

Website client needs update post this PR. RFF does not use the facet field https://github.com/billybabis/billybabis.github.io/blob/main/static/src/fetchData.js, so not affected.


pradh (approved):
Glad we caught it now before people start using the APIs :)

shifucun:
Thanks!

PR #928, opened 2022-09-23
improve perf of coordinate recon

Spaceenter:


pradh (approved):
Thanks for the quick change!

Fructokinase (approved):
wow, you're so fast! Thanks for the changes.

PR #933, opened 2022-09-29
update base.mcf for Provenance/Dataset/Source

n-h-diaz:


deploy/mapping/base.mcf
```diff
@@ -170,6 +170,7 @@ description: C:Provenance->provenance_description
 importUrl: C:Provenance->mcf_url
 importTime: C:Provenance->timestamp_secs
 importDuration: C:Provenance->duration_secs
+isPartOf: C:Provenance->is_part_of
```
pradh, line 173: I think this should be `E:Provenance->E5`, where E5 is:

```
Node: E:Provenance->E5
typeOf: Dataset
dcid: C:Provenance->is_part_of
functionalDeps: dcis
```

deploy/mapping/base.mcf
```diff
@@ -208,13 +209,28 @@ typeOf: Source
...
 dcid: C:Source->prov_id
 functionalDeps: dcid
 
+Node: E:Dataset->E1
+typeOf: Dataset
+dcid: C:Dataset->id
+provenance: E:Dataset->E2
+name: C:Dataset->name
+url: C:Dataset->url
+isPartOf: C:Dataset->is_part_of
```
pradh, line 227: Like above this should be E3 where E3 is a Source.

deploy/mapping/base.mcf
pradh: Sorry I missed this in the other mapping.

n-h-diaz: got it thanks! will update

deploy/mapping/base.mcf
n-h-diaz: done

deploy/mapping/base.mcf
```diff
@@ -188,6 +188,11 @@ typeOf: Provenance
 dcid: C:Provenance->prov_id
 functionalDeps: dcid
 
+Node: E:Provenance->E5
```
pradh, line 191: We need an `isPartOf: E:Provenance->E5` added to the E:Provenance->E1 node?

deploy/mapping/base.mcf
```diff
@@ -216,6 +223,25 @@ dcid: C:Source->prov_id
 functionalDeps: dcid
 
 
+Node: E:Dataset->E1
+typeOf: Dataset
+dcid: C:Dataset->id
+provenance: E:Dataset->E2
+name: C:Dataset->name
+url: C:Dataset->url
```
pradh, line 231: ditto.  add `isPartOf` here?

n-h-diaz: done

n-h-diaz:
> Thanks!
> 
> Lets just sanity check this with @shifucun for a sec before merging :)

ahhh sorry I didn't see this - I can revert the PR if there are any concerns with this

PR #934, opened 2022-10-04
Make endpoint group for observations API

shifucun:
There have been a lot of webdriver tests failure, which could be due to resource issue in autopush mixer. Create a new group for observations API to narrow down / fix the issue.

### DIFF
--- a/deploy/helm_charts/envs/mixer_autopush.yaml
+++ b/deploy/helm_charts/envs/mixer_autopush.yaml
@@ -11,16 +11,16 @@ ingress:
     }
 
 serviceGroups:
-  - mixer-svg-service:
+  - svg:
       urlPaths:
         - "/place/stat-var-group/*"
         - "/stat-var/*"
         - "/v1/info/variable/group/*"
         - "/v1/variable/*"
       replicas:
-        default: 2
-        min: 2
-        max: 8
+        default: 5
+        min: 5
+        max: 10
       resources:
         memoryRequest: "8G"
         memoryLimit: "8G"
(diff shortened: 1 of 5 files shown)


deploy/helm_charts/envs/mixer_autopush.yaml
```diff
@@ -34,6 +34,17 @@ serviceGroups:
       resources:
         memoryRequest: "400M"
         memoryLimit: "400M"
+  - mixer-default-service:
```
Fructokinase, line 37: mixer-observations-service?

deploy/helm_charts/mixer/values.yaml
```diff
@@ -120,6 +120,17 @@ serviceGroups:
       resources:
         memoryRequest: "400M"
         memoryLimit: "400M"
+  - mixer-default-service:
```
Fructokinase, line 123: mixer-observations-service?

deploy/helm_charts/envs/mixer_prod.yaml
```diff
@@ -11,16 +11,16 @@ ingress:
...
       urlPaths:
         - "/place/stat-var-group/*"
         - "/stat-var/*"
         - "/v1/info/variable/group/*"
         - "/v1/variable/*"
       replicas:
-        default: 30
-        min: 30
-        max: 50
+        default: 10
```
shifucun, line 21: These API is very low cost in memory key value read and should be serve high qps per pod, hence reducing the replicas.

deploy/helm_charts/envs/mixer_autopush.yaml
shifucun: updated all the name to use just the label.

Fructokinase (approved):

PR #938, opened 2022-10-08
Fix a few release config issue

shifucun:
With this, I have redeployed mixer_private

### DIFF
--- a/deploy/helm_charts/envs/mixer_encode.yaml
+++ b/deploy/helm_charts/envs/mixer_encode.yaml
@@ -26,5 +26,6 @@ serviceGroups:
 ip: 34.107.228.109
 region: us-central1
 api_title: DataCommons API (Encode)
-api: datacommons.Mixer
+api:
+  - datacommons.Mixer
 nodes: 1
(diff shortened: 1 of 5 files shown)


deploy/helm_charts/envs/mixer_encode.yaml
```diff
@@ -26,5 +26,6 @@ serviceGroups:
 ip: 34.107.228.109
 region: us-central1
 api_title: DataCommons API (Encode)
-api: datacommons.Mixer
+api:
+  - datacommons.Mixer
```
beets, line 30: just for my understanding - is this purely a formatting change, or does the `-` have an effect?

beets (approved):

deploy/helm_charts/envs/mixer_encode.yaml
shifucun: This represents a list, while the old way should be a single value https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html#yaml-basics

PR #940, opened 2022-10-13
Add V1 API for getting observation dates based on entity count

shifucun:
Compared with previous version, the major change here is the response shape, where it's oriented by the date with the entity frequency for each facet. This aligns well with its usage, where user wants to know which facet to use for each date.

The dates and count for facets are both sorted.

Also updated the uri and args to match V1 patterns.


internal/server/v1/observationdates/observation_dates_linked.go
```diff
@@ -0,0 +1,130 @@
...
+	store *store.Store,
+) (
+	*pb.BulkObservationDatesLinkedResponse, error) {
+	linkedEntity := in.GetLinkedEntity()
+	entityType := in.GetEntityType()
+	linkedProperty := in.GetLinkedProperty()
+	variables := in.GetVariables()
+	if linkedEntity == "" {
+		return nil, status.Errorf(codes.InvalidArgument,
+			"Missing required argument: ancestor_place")
```
chejennifer, line 46: the argument is linked_entity right?

chejennifer (approved):
This looks good to me!

internal/server/v1/observationdates/observation_dates_linked.go
shifucun: Fixed

proto/v1/observation_dates.proto
```diff
@@ -0,0 +1,50 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+option go_package = "./proto";
+package datacommons.v1;
+
+import "stat.proto";
+
+message EntityCount {
```
pradh, line 21: Add some comments to these protos?  e.g., count of what?

pradh (approved):

proto/v1/observation_dates.proto
shifucun: Added. Thanks!

shifucun:
Thanks for review!

PR #944, opened 2022-10-15
Increase helm timeout from 5min to 10min

Fructokinase:
Some builds are timing out. Helm waits for a default of 5min. Immediate solution is to increase this.

Example 
https://pantheon.corp.google.com/cloud-build/builds;region=global/bdd7bfb9-d824-47f9-b7a8-a71c7b35e47c;step=0?mods=-monitoring_api_staging&project=datcom-ci

helm waits for all service groups in series, there's a way to change that.

### DIFF
--- a/deploy/helm_charts/README.md
+++ b/deploy/helm_charts/README.md
@@ -6,12 +6,12 @@ A single helm chart organizes a collection of templatized k8s yaml files.
 
 ## Installing a Helm chart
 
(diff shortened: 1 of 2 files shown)


shifucun (approved):

shifucun:
It still times out. I wonder why it's taking much long than before for this to finish? There may be some other real issues.

PR #945, opened 2022-10-20
Allow point observation to filter by facet; Return all facets data in all response

shifucun:
This allows users to achieve the following functions:

1. Query observation point with "all_facets=False" and still get all the facets information.
2. Able to use facet id as query param to get the desired facet data, instead of fetching ALL facets.

This would help improve clients like map tool, where fetching all facets data of IPCC stat vars is just too much.

This change won't affect the existing API, but is only an feature addition.


proto/v1/observations.proto
```diff
@@ -33,6 +33,7 @@ message EntityObservations {
 message VariableObservations {
   string variable = 1;
   repeated EntityObservations observations_by_entity = 2;
+  map<string, datacommons.StatMetadata> all_available_facets = 3;
```
chejennifer, line 36: I think it's very valuable to be able to get all_available_facets for a stat var for certain entitites, but I wonder if it might be confusing to put it here in this response. Could it be confusing to a user that the facets here do not have to do with the observations in observations_by_entity? And especially if user chooses to filter by a specific facet, but we're still giving them information on all_available_facets? I also notice that we don't populate this when user asks for all_facets, I understand that the same information will be in the `facets` field, but could the inconsistency be confusing

internal/server/v1/observations/bulk_point_linked.go
```diff
@@ -204,12 +231,13 @@ func BulkPointLinked(
 		}
 	}
 	// Get the preferred facet
-	if !allFacets {
-		for _, varibleObservation := range result.ObservationsByVariable {
-			for _, entityObservation := range varibleObservation.ObservationsByEntity {
-				if len(entityObservation.PointsByFacet) == 0 {
-					continue
-				}
+	for _, varibleObservation := range result.ObservationsByVariable {
```
chejennifer, line 234: nit: variableObservation

proto/v1/observations.proto
shifucun: Another option is to have a new API that only returns the facet information.

@pradh any thoughts on this too?

shifucun: @Spaceenter A new cache on all facets would be helpful

shifucun:
Closing this in favor of upcoming V2 API

PR #949, opened 2022-10-24
fix series picking logic bug when same import in multiple groups

shifucun:


internal/server/stat/stat_util.go
```diff
@@ -255,7 +255,9 @@ func CollectDistinctSourceSeries(seriesList ...[]*pb.SourceSeries) []*pb.SourceS
 	for _, series := range seriesList {
 		for _, s := range series {
 			metahash := getSourceSeriesHash(s)
-			if _, ok := resultMap[metahash]; ok && len(s.Val) < len(resultMap[metahash].Val) {
+			// The input series list is sorted, so the for loop is from the msot
```
pradh, line 258: nit: most

pradh, line 258:
```
// For same source series, the one with more data points is used. In most cases,
// this is the series with the latest data as well.
```

For what it's worth, it might be worth tweaking this slightly.  For example, I recall seeing that WDI refresh in some cases dropping some of the older data.  So preferring import-group with a later date, and if the latest date is the same, using the import-group preference order...

pradh (approved):
Thanks for the quick fix!

internal/server/stat/stat_util.go
shifucun: Done

shifucun: Updated

pradh (approved):
Thanks for the change!

PR #950, opened 2022-10-24
Unquote schema mapping

Fructokinase:
This is currently causing autopush to fail to be deployed in GCP.


deploy/helm_charts/mixer/templates/config_maps.yaml
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# All things that can vary for the same mixer code should be 
+# All things that can vary for the same mixer code should be
 # in config maps. This includes BT, BQ versions, schema mcfs, so on.
 #
 # Note: Please prefix the name of each config map with
```
shifucun, line 18: how about line 55 below?

shifucun (approved):

deploy/helm_charts/mixer/templates/config_maps.yaml
Fructokinase: Unquoted everything except for githash.

PR #958, opened 2022-10-31
Use place page categories cache to populate validCategories field

shifucun:


internal/server/v1/page/golden/place_page/zuid-nederland.sample.json
```diff
@@ -18133,10 +18133,41 @@
     }
   },
   "validCategories": {
-    "country/NLD": {},
-    "nuts/NL4": {},
-    "nuts/NL41": {},
-    "nuts/NL42": {}
+    "country/NLD": {
+      "category": [
```
pradh, line 18137: why were these empty on the LHS?

internal/server/placepage/place_page.go
```diff
@@ -244,11 +247,42 @@ func fetchBtData(
...
+			}
+			return &categories, nil
+		},
+	)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	categoryData := map[string]*pb.Categories{}
+	for _, place := range places {
+		categoryData[place] = &pb.Categories{Category: []string{}}
```
pradh, line 269: I don't know where else Categories is used, but it seems it is internal to mixer?

The proto corresponding to this is http://google3/knowledge/answers/stats/proto/chart_store.proto;l=80-84;rcl=485168034

Also, should we mark this reserverd: https://github.com/datacommonsorg/mixer/blob/34989caa16f7ba773fd2cfc090e78e5a155f5801/proto/internal.proto#L83 ?

pradh (commented):
Thanks!

internal/server/placepage/place_page.go
shifucun: I updated the proto to match the internal one.

internal/server/v1/page/golden/place_page/zuid-nederland.sample.json
shifucun: They should contain the actual valid categories from the definition. To me, it's bug from previous implementation.

pradh (approved):
Thanks.  Just to confirm, I don't see the proto file diff?

shifucun:
> Thanks. Just to confirm, I don't see the proto file diff?

The push was halted. Now it's there.

pradh (approved):

PR #959, opened 2022-11-01
Add a new method for resolving place names.

Spaceenter:


proto/recon.proto
```diff
@@ -112,6 +112,24 @@ message ResolveIdsResponse {
   repeated Entity entities = 1;
 }
 
+message ResolvePlaceNamesRequest {
+  message Place {
+    string name = 1;
+    string type = 2;
+  }
+  repeated Place places = 1;
+}
```
pradh, line 121: Is it worth having a non-repeated version for simple use?

/v1/find/entity?description="capital city in California"
/v1/find/entity?description="santa clara"&type=County

proto/recon.proto
```diff
@@ -112,6 +112,24 @@ message ResolveIdsResponse {
   repeated Entity entities = 1;
 }
 
+message ResolvePlaceNamesRequest {
+  message Place {
+    string name = 1;
```
pradh, line 117: description?

internal/util/util.go
```diff
@@ -501,3 +508,39 @@ func StringListIntersection(list [][]string) []string {
 
 	return res
 }
+
+// MapsClient gets the client for Maps API.
+func MapsClient(ctx context.Context) (*maps.Client, error) {
+	var projectID string
+	credentials, err := google.FindDefaultCredentials(ctx, compute.ComputeScope)
+	if err != nil || credentials == nil || credentials.ProjectID == "" {
+		projectID = "datcom-store"
```
pradh, line 517: nit: use global constant `defaultProjectID`

internal/server/recon/place_name.go
```diff
@@ -0,0 +1,271 @@
...
+		name := place.GetName()
+		if name == "" {
+			return nil, fmt.Errorf("empty place name")
+		}
+		nameSet[name] = struct{}{}
+		placeInfoSet[placeInfo{name, place.GetType()}] = struct{}{}
+	}
+
+	// Get place IDs.
+	nameToPlaceIDs, placeIDSet, err := resolvePlaceIDs(ctx, mapsClient, nameSet)
```
pradh, line 70: When a type is provided its typically useful to append that to the geocoding name.  e.g., NY State, NY City, Santa Clara County, etc.

internal/server/recon/place_name.go
```diff
@@ -0,0 +1,271 @@
...
+	mapsClient *maps.Client,
+	nameSet map[string]struct{},
+) (*sync.Map /* name -> place IDs */, *sync.Map /* place ID set */, error) {
+	type placeInfo struct {
+		name     string
+		placeIDs []string
+	}
+	var wg sync.WaitGroup
+
+	// Send place names to channel.
```
pradh, line 142: I confess I only vaguely follow go synchronization.... wonder if it would help to provide a preface to how we're parallelizing?

My understanding....  Set up a "request" channel of names to be resolved, set up `maxMapsAPICallsInParallel` "response" channels for resolved place IDs, run `maxMapsAPICallsInParallel` goroutines that each reads from the "request" channel, calls place API and writes to the "response" channel ?

internal/server/recon/place_name.go
```diff
@@ -0,0 +1,271 @@
...
+		Input:     name,
+		InputType: maps.FindPlaceFromTextInputTypeTextQuery,
+		Fields:    []maps.PlaceSearchFieldMask{maps.PlaceSearchFieldMaskPlaceID},
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	placeIDs := []string{}
+	for _, candidate := range resp.Candidates {
```
pradh, line 208: If this is the candidate type - https://github.com/googlemaps/google-maps-services-go/blob/391036138cfed529e87aaad49922e96da4a393da/places.go#L331-L332 - looks like it might also include the type.  So in future, we could consider using that to prune unexpected types.

internal/server/recon/place_name.go
```diff
@@ -0,0 +1,271 @@
...
+		placeIDs = append(placeIDs, candidate.PlaceID)
+	}
+
+	return placeIDs, nil
+}
+
+func resolveDCIDs(
```
pradh, line 215: Document what it returns?

internal/server/recon/place_name.go
```diff
@@ -0,0 +1,271 @@
...
+			}
+		}
+
+		resp.Places = append(resp.Places, place)
+	}
+
+	return resp, nil
+}
+
+func resolvePlaceIDs(
```
pradh, line 131: Document what it returns?

internal/server/recon/place_name.go
```diff
@@ -0,0 +1,271 @@
...
+		}
+
+		if placeIDs, ok := nameToPlaceIDs.Load(info.placeName); ok {
+			allDCIDs := []string{}
+			for _, placeID := range placeIDs.([]string) {
+				if dcids, ok := placeIDToDCIDs[placeID]; ok {
+					allDCIDs = append(allDCIDs, dcids...)
+				}
+			}
+			sort.Strings(allDCIDs)
```
pradh, line 102: Hmm... here it seems this sort might make use lose the order of placeIDs returned from the Maps API?

That might be valuable to keep.  From what I've seen, the first placeId is usually the most relevant one, and subsequent ones may not be right.

PR #961, opened 2022-11-01
Only build search cache for svg service groups

Fructokinase:


cmd/main.go
```diff
@@ -61,6 +61,8 @@ var (
 	importGroupTables = flag.String("import_group_tables", "", "Newline separated list of import group tables")
 	// Branch Bigtable Cache
 	useBranchBt = flag.Bool("use_branch_bt", true, "Use branch bigtable cache")
+	// Stat-var search cache
+	useSearch      = flag.Bool("use_search", true, "Uses stat-var search. Will build search indexes.")
```
shifucun, line 65: should this be false by default?

Fructokinase: I put it to true because I'm not sure I was afraid of breaking existing instances in case it gets deployed from the latest master, where service groups do not exist (like website deployments). Only service group yaml deployments can make it false. But I think you should make the call! What do you think?

shifucun: That makes sense. I see in the default deploy file it is set to false. So this works.

Please check the memory usage etc to make sure this works. Also can you adjust the default resource after this is merged?

shifucun (approved):

cmd/main.go
Fructokinase: Will merge this and watch autopush, then adjust the resources in a followup CL.

PR #971, opened 2022-11-04
Remove two deprecated API: GetStatSetSeries, GetStatSetSeriesWithinPlace

shifucun:
These two APIs are never made public and were only used for the website app. As the website migrated to V1 APIs, these can be removed.

To confirm on the usage. Here are the traffic:

Mixer (30 days):
image

Website (7 days, as the usage was removed since last release):
image

The 4 calls to /stat/set/series/within-place have IP from Google crawlers.


pradh:
To confirm, does the mixer have some usage for within-place?  And to confirm, that previous Stanford folks are not using it?

pradh (approved):

shifucun:
> To confirm, does the mixer have some usage for within-place? And to confirm, that previous Stanford folks are not using it?

Does not look like they have any recent usage. And these were not in the list of APIs that we provide to them either.

PR #972, opened 2022-11-07
[BulkFindEntities] Only keep the most likely place IDs

Spaceenter:


pradh (approved):
Does it not impact the golden tests?

PR #976, opened 2022-11-09
Fix failing recon API test

juliawu:
This PR updates `resolvePlaceIDs()` so that `bulk_find_entities_test.go` passes.

With this change, descriptions that do not map to any dcids are still returned, just without matches (e.g. the "non-existing place wow" case in [/internal/server/recon/bulk_find_entities_test.go:51](https://github.com/datacommonsorg/mixer/blob/ade6e189a0ff2efb251dd32517208cfe5f73d123/internal/server/recon/golden/bulk_find_entities_test.go#L51)

### DIFF
--- a/internal/server/recon/find_entities.go
+++ b/internal/server/recon/find_entities.go
@@ -170,7 +170,6 @@ func resolvePlaceIDs(
 					resolveResultChan <- resolveResult{
 						entityInfo: &entityInfo,
 						placeIDs:   []string{}}
-
 				} else {
 					// Only keep the first place ID, as the rest ones are usually much less accurate.
 					resolveResultChan <- resolveResult{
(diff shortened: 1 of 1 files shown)


internal/server/recon/find_entities.go
```diff
@@ -166,10 +166,17 @@ func resolvePlaceIDs(
 				if err != nil {
 					return err
 				}
-				// Only keep the first place ID, as the rest ones are usually much less accurate.
-				resolveResultChan <- resolveResult{
-					entityInfo: &entityInfo,
-					placeIDs:   []string{placeIDs[0]}}
+				if len(placeIDs) == 0 {
```
shifucun, line 169: nit:
`
usedPlaceIds = []string{}
if len(placeIDs) > 0 {
usedPlaceds = []string{placeIDs[0]}
}
resolveResultChan <- resolveResult{..., placeIDs: usedPlaceids}

juliawu: This is much more elegant. Fixed!

shifucun (approved):

PR #977, opened 2022-11-09
Update custom import group and test

shifucun:
Previously a feature to have a custom import group in the request was added. It was meant to support a user dashboard and import workflow, which is halted now.

With the recent config change, mixer allows custom import group to be configured and created during server start up.

I am removing the API request level import group for now before we have a clear idea on how it would be done.

This PR also adds a custom import group to the golden test.


pradh (approved):
Nice cleanup!

PR #986, opened 2022-11-17
[Place Page] Remove v0 API; Always set category in place page golden test

shifucun:
Since the place page cache is only per category now.


internal/server/v1/page/golden/place_page_test.go
```diff
@@ -59,7 +59,7 @@ func TestPlacePage(t *testing.T) {
 				1,
```
pradh, line 59: Do we need to handle v0 test file too?

shifucun: They are not used anymore; Removing the v0 code and golden tests.

proto/v1/page.proto
```diff
@@ -32,3 +60,17 @@ message PlacePageRequest {
   // The category for which data needs to be fetched.
   string category = 4;
 }
+
+// Response to get place page info for a place.
+message PlacePageResponse {
```
pradh, line 65: Hard to review the diffs... is the expectation that the proto fields exactly match v0 so that website code doesn't have to change? Or is there a corresponding website change?

internal/server/v1/page/place.go
```diff
@@ -30,7 +30,7 @@ func PlacePage(
 	ctx context.Context,
 	in *pb.PlacePageRequest,
 	store *store.Store,
-) (*pb.GetPlacePageDataResponse, error) {
+) (*pb.PlacePageResponse, error) {
 	node := in.GetNode()
 	if !util.CheckValidDCIDs([]string{node}) {
 		return nil, status.Errorf(codes.InvalidArgument, "Invalid entity")
```
pradh, line 36: should we fail on empty category (eg line 39)?

pradh (approved):

internal/server/v1/page/place.go
pulkit-s: We could also default to overview in case of empty category

shifucun: Since this is at API level, it should be more direct. I would opt for return error if category is not set.

Updated that and removed the old cache in mixer as well.

internal/server/v1/page/golden/place_page_test.go
```diff
@@ -83,47 +83,47 @@ func TestPlacePage(t *testing.T) {
...
+				"Equity",
 			},
 			{
 				"zuid-nederland.sample.json",
 				"nuts/NL4",
 				1,
 				[]string{},
 				5,
-				"",
+				"Overview",
```
beets, line 118: would be great to have all the categories represented for goldens -- health, equity, economics, ..

internal/server/v1/page/place.go
beets: +1 for immediate failure

beets (commented):
thanks for the cleanup!

internal/server/v1/page/golden/place_page_test.go
shifucun: I will update the golden tests in a follow up PR.

PR #987, opened 2022-11-17
Cover all categories in place page golden test

shifucun:
Also merged all the place page code into one file, as the v0 handler and wrapper has been deleted.


beets (approved):
thanks bo!

PR #995, opened 2022-12-13
Add a new API method EventCollection.

Spaceenter:


proto/v1/event.proto
```diff
@@ -0,0 +1,62 @@
...
+    message Point {
+      optional double latitude = 1;
+      optional double longitude = 2;
+    }
+    message Polygon { optional string geo_json = 1; }
+    oneof geo {
+      Point point = 1;
+      Polygon polygon = 2;
+    }
+  }
```
shifucun, line 30: I wonder should this proto be outside "EventCollection" since it feels like a fairly standalone concept?

shifucun: Maybe make PropVal and Event also top level proto, so they can potentially be reused for more APIs

proto/v1/event.proto
```diff
@@ -0,0 +1,62 @@
...
+
+  message Event {
+    // Data Commons places associated with the event.
+    repeated string places = 1;
+    // Geospatial location of the event.
+    repeated GeoLocation geo_locations = 2;
+    // NOTE: The format should be ISO 8601.
+    repeated string dates = 3;
+    repeated PropVal salient_pvs = 4;
```
shifucun, line 44: what is salient_pvs?

proto/mixer.proto
```diff
@@ -526,6 +527,19 @@ service Mixer {
...
 
+  // Get event collection for {eventType, affectedPlaceDcid, date}.
+  // NOTE:
+  // - The affectedPlaceDcid is only for top-level places such as Earth,
+  // continents, and countries.
+  // - The date format should be: YYYY-MM.
+  rpc EventCollection(datacommons.v1.EventCollectionRequest)
+      returns (datacommons.v1.EventCollectionResponse) {
+    option (google.api.http) = {
+      get : "/v1/event/collection"
```
shifucun, line 538: maybe just "/v1/events", then it matches other API naming to get a list of triples/observations/etc..

internal/server/v1/page/golden/place_page/asm.Education.json
```diff
@@ -454,51 +304,6 @@
     }
   },
   "latestPopulation": {
-    "country/ASM": {
```
shifucun, line 457: The deletion is suspicious.

internal/server/v1/event/event.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	resp := &pb.EventCollectionResponse{}
+
+	// Go through (ordered) import groups one by one, stop when data is found.
+	for _, btData := range btDataList {
+		for _, row := range btData {
+			resp.EventCollection = row.Data.(*pb.EventCollection)
```
shifucun, line 63: break?

proto/v1/event.proto
```diff
@@ -0,0 +1,62 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+option go_package = "./proto";
+package datacommons.v1;
+
+message EventCollection {
```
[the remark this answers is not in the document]
Spaceenter: done

proto/v1/event.proto
Spaceenter: This is to exactly match the the protos in .../proto/prophet/cache_data.proto, so that we can directly pass the value in cache to API client... We could try to make a change in the cache first before modifying here.

proto/v1/event.proto
Spaceenter: Added comment.

proto/mixer.proto
Spaceenter: done

internal/server/v1/event/event.go
Spaceenter: Good catch... done

internal/server/v0/statvarsummary/golden/get_statvar_summary/females.json
```diff
@@ -1091,24 +1091,8 @@
                   "minValue": 0,
                   "maxValue": 72821
                 },
-                "Place": {
```
shifucun, line 1094: This deletion is suspicious..

internal/server/v1/page/golden/place_page/ca.Environment.json
```diff
@@ -21741,43 +21741,6 @@
             "unit": "MetricTonCO2e"
           }
         },
-        "Count_CycloneEvent": {
```
shifucun, line 21744: why is this removed?

internal/server/v1/propertyvalues/golden/bulk_property_values_in/containedIn1.json
```diff
@@ -22556,31 +22556,31 @@
             "PowerPlant"
           ],
           "dcid": "dc/kg6g6mkqg9j2g",
-          "provenanceId": "dc/df6wc4"
+          "provenanceId": "dc/kl1tbs3"
```
shifucun, line 22559: @pradh is prov change expected?

proto/mixer.proto
```diff
@@ -526,6 +527,19 @@ service Mixer {
...
 
+  // Get event collection for {eventType, affectedPlaceDcid, date}.
+  // NOTE:
+  // - The affectedPlaceDcid is only for top-level places such as Earth,
+  // continents, and countries.
+  // - The date format should be: YYYY-MM.
+  rpc EventCollection(datacommons.v1.EventCollectionRequest)
+      returns (datacommons.v1.EventCollectionResponse) {
+    option (google.api.http) = {
+      get : "/v1/events"
```
shifucun, line 538: @beets @pradh Wonder if this is for general "Event" or for "Disaster/Climate Event"?

proto/v1/event.proto
```diff
@@ -0,0 +1,65 @@
...
+message EventCollectionRequest {
+  string event_type = 1;
+  // Only for top-level places such as Earth, continents, and countries.
+  string affected_place_dcid = 2;
+  // YYYY-MM.
+  string date = 3; 
+}
+
+message EventCollectionResponse {
+  EventCollection event_collection = 1;
```
chejennifer, line 64: Could it make sense to include the source information (i.e., name, source, and url) for each unique provenance that is included in event_collection? It would save having to do another call in Flask to get this information

proto/v1/event.proto
```diff
@@ -0,0 +1,65 @@
...
+    oneof geo {
+      Point point = 1;
+      Polygon polygon = 2;
+    }
+  }
+
+  message PropVal {
+    optional string prop = 1;
+    repeated string val = 2;
+  }
```
beets, line 36: is it possible to use a map here? so we can just have:

{
"magnitude": "3.5",
"depth": "Kilometer9.8"
}
this would match our newer v1 APIs. should we also match the triples API here and return provenances for each pv.

proto/v1/event.proto
beets: are we calling it salient_pvs because it is not all the pv's attached to the event?

proto/v1/event.proto
```diff
@@ -0,0 +1,65 @@
...
+    repeated string places = 1;
+    // Geospatial location of the event.
+    repeated GeoLocation geo_locations = 2;
+    // NOTE: The format should be ISO 8601.
+    repeated string dates = 3;
+    // These are the properties and values that well characterize the event.
+    // For example, for EarthquakeEvent, one of the salient property is the magnitude.
+    repeated PropVal salient_pvs = 4;
+    optional string dcid = 5;
+    optional string provenance = 6;
```
beets, line 49: nit: provenanceId to match triples API

proto/v1/event.proto
beets: I had a similar question regarding facets - we would need to add something similar here prior to publishing this API

proto/v1/event.proto
```diff
@@ -0,0 +1,65 @@
...
+    repeated string val = 2;
+  }
+
+  message Event {
+    // Data Commons places associated with the event.
+    repeated string places = 1;
+    // Geospatial location of the event.
+    repeated GeoLocation geo_locations = 2;
+    // NOTE: The format should be ISO 8601.
+    repeated string dates = 3;
```
beets, line 44: How should a user interpret this date? Is it the start / end date?

proto/v1/event.proto
```diff
@@ -0,0 +1,65 @@
...
+
+  repeated Event events = 1;
+}
+
+message EventCollectionRequest {
+  string event_type = 1;
+  // Only for top-level places such as Earth, continents, and countries.
+  string affected_place_dcid = 2;
+  // YYYY-MM.
+  string date = 3;
```
beets, line 60: How is this date interpreted in the query? Do we return all dates that intersect? e.g. if a flood lasts from 2022-01-01 to 2022-02-01, and the user requests 2022-01-10, would we return that flood

beets (commented):
Mostly have lots of questions around the API design. Sorry for missing out on earlier discussions while we were designing the cache.

proto/v1/event.proto
Spaceenter: Right, not all properties are selected, "salient_pv" is a name suggested by @pradh , who might have more sayings here.

pradh: Yes, not all properties in an event are in the cache, just the more meaningful ones (for example, there are things like some internal IDs, date of review, etc)

I do recognize that this distinction could be useful in the implementation-side (cache proto), but confusing to the user (api proto).  If there's a benefit to keeping the protos the same, we can call it properties (and add a comment in the cache_data.proto).

proto/v1/event.proto
Spaceenter: This is using the current/old event schema, so there is no such a thing of "facet" for event. But I get it that you need more info re provenance (import name, source, url) -- I'm adding those info in the event cache in backend prophet in cpp, and will send out for reviews.

Spaceenter:
Deprecating this PR in favor of https://github.com/datacommonsorg/mixer/pull/997 -- the feedback has been implemented in both backend cache and the new PR.

PR #999, opened 2022-12-21
Add EventCollectionDate API method.

Spaceenter:
Also modified the golden test for EventCollection to use CA instead of USA, so that the golden file is not that large.


proto/v1/event.proto
```diff
@@ -70,5 +70,21 @@ message EventCollectionRequest {
...
+}
+
+message EventCollectionDateRequest {
+  string event_type = 1;
+  // Only for top-level places: Earth, continent, country, state, adminArea1.
+  string affected_place_dcid = 2;
+}
+
+message EventCollectionDateResponse {
+  EventCollectionDate event_collection_date = 1;
```
chejennifer, line 89: curious why we need the EventCollectionDate type when it's just a single `repeated string`?

chejennifer (approved):
thanks for adding this!

proto/v1/event.proto
Spaceenter: For one thing, we need the EventCollectionDate to do type casting for the data read from BT, see event.go. 
Regarding here, I guess either is fine, but the intention is to make the API a transparent proxy to the BT cache.

PR #1000, opened 2022-12-22
Update /v1/bulk/info/variable-group with richer response data

shifucun:
Calls the simple version of the API VariableGroupInfo() concurrently to get the bulk response.


jehangiramjad (approved):
Thanks!

PR #1002, opened 2023-01-05
Add a new API to check if observation exist for <variable, entity> collection

shifucun:


internal/server/v1/observationexistence/golden/observation_existence/existence.json
```diff
@@ -0,0 +1,25 @@
+{
+  "by_variable": {
```
shifucun, line 2: @beets @chejennifer This json data will be consumed by the frontend

beets (approved):
lgtm

my nit would be for the response to drop the `by_` prefix, and simply use `variables` and `entities` like the request. but ok to skip in the interest of time.

shifucun:
> lgtm
> 
> my nit would be for the response to drop the `by_` prefix, and simply use `variables` and `entities` like the request. but ok to skip in the interest of time.

Updated

PR #1003, opened 2023-01-06
[Event API] Merge country data to earth; Add filter mechanism

shifucun:


proto/v1/event.proto
```diff
@@ -67,9 +67,10 @@ message EventCollectionRequest {
   // - discoveryDate
   // - occurrenceTime
   string date = 3;
+  double fire_event_area_limit = 4;
```
shifucun, line 70: @chejennifer Please set this value in frontend request. It filters out events with smaller area.

Spaceenter (approved):

proto/v1/event.proto
```diff
@@ -67,9 +67,14 @@ message EventCollectionRequest {
   // - discoveryDate
   // - occurrenceTime
   string date = 3;
+  // The property used for filtering
+  string filter_prop = 4;
+  // Event is not returned if value below the limit
+  double filter_limit = 5;
```
chejennifer, line 73: Will this API only handle filtering by a lower limit? Currently in the UI, user can also choose an upper limit..

proto/v1/event.proto
```diff
@@ -67,9 +67,14 @@ message EventCollectionRequest {
   // - discoveryDate
   // - occurrenceTime
   string date = 3;
+  // The property used for filtering
+  string filter_prop = 4;
```
chejennifer, line 71: Since an event can have multiple prop vals that have to do with severity (e.g., directInjuries and directDeaths), do we only want to allow filtering by one prop?

internal/server/v1/event/event.go
```diff
@@ -27,6 +30,27 @@ import (
...
+type filterSpec struct {
+	prop  string
+	unit  string
+	limit float64
+}
+
+func keepEvent(event *pb.EventCollection_Event, spec filterSpec) bool {
+	for prop, vals := range event.GetPropVals() {
+		if prop == spec.prop {
+			valStr := strings.TrimPrefix(vals.Vals[0], spec.unit)
```
chejennifer, line 42: just checking, will the unit always be the prefix?

internal/server/v1/event/event.go
```diff
@@ -27,6 +30,27 @@ import (
...
+			if v, err := strconv.ParseFloat(valStr, 64); err == nil {
+				return v > spec.limit
```
chejennifer, line 44: should it be >=?

internal/server/v1/event/event.go
```diff
@@ -55,16 +105,42 @@ func Collection(
...
+					}) {
+						continue
+					}
+				}
+				resp.EventCollection.Events = append(
+					resp.EventCollection.Events,
+					event,
+				)
+			}
+			for provID, info := range data.ProvenanceInfo {
```
chejennifer, line 138: not sure how likely this would be to happen, but if all events with a specific provID gets filtered out, probably don't want to keep that provID in the provenanceInfo?

chejennifer (commented):
Thanks for the changes, just a few questions

internal/server/v1/event/event.go
shifucun: It should be from current schema. If it deviates, we can set up a config here so the function is based on the unit

internal/server/v1/event/event.go
shifucun: good catch!

proto/v1/event.proto
shifucun: Do you foresee the use case of filtering by multiple property? I wonder if a combination of filtering in mixer (reduce data size) and in client (for UI actions) is preferred. Anyway, if we do need to filter by multiple props, can update this into a nested structure.

proto/v1/event.proto
shifucun: My original intention was to use this reduce the size of the data. For a more comprehensive filtering, maybe we should start a doc and decide from that?

internal/server/v1/event/event.go
shifucun: good point. set provId in event loop now.

proto/v1/event.proto
beets: could we have two optional limits (upper_limit, lower_limit)? if neither is set, all data is returned.

chejennifer (approved):
thanks for the changes, just one small nit!

internal/server/v1/event/event.go
```diff
@@ -39,12 +63,38 @@ func Collection(
 	if !util.CheckValidDCIDs([]string{in.GetAffectedPlaceDcid()}) {
 		return nil, status.Errorf(codes.InvalidArgument, "Invalid DCID")
 	}
+	filterProp := in.GetFilterProp()
+	filterLimit := in.GetFilterLimit()
+	filterUnit := in.GetFilterUnit()
+
+	// Merge all the affected places in the final result.
+	affectedPlaces := []string{}
+	if in.GetAffectedPlaceDcid() == "Earth" {
```
beets, line 72: we should do this for continents here too.

proto/v1/event.proto
chejennifer: I think this depends on how many severity props events will have and how much we want to allow a user to play around with setting filters. Having just one filter is ok for now

internal/server/v1/event/event.go
shifucun: Done

internal/server/v1/event/event.go
```diff
@@ -55,16 +105,41 @@ func Collection(
...
+						limit: filterLimit,
+					}) {
+						continue
+					}
+				}
+				resp.EventCollection.Events = append(
+					resp.EventCollection.Events,
+					event,
+				)
+				proveID := event.ProvenanceId
```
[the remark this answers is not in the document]
shifucun: Done

proto/v1/event.proto
shifucun: Done

chejennifer (approved):
thanks for the updates!

internal/server/v1/event/event.go
```diff
@@ -27,6 +30,40 @@ import (
...
+}
+
+func keepEvent(event *pb.EventCollection_Event, spec filterSpec) bool {
+	for prop, vals := range event.GetPropVals() {
+		if prop == spec.prop {
+			valStr := strings.TrimPrefix(vals.Vals[0], spec.unit)
+			v, err := strconv.ParseFloat(valStr, 64)
+			if err != nil {
+				// Can not convert the value, keep.
+				return true
```
beets, line 58: this is fine for now if the idea here is to reduce data size. can we add a note of that principle with the request or here in the function?

beets (approved):
thanks for the updates!

internal/server/v1/event/event.go
shifucun: Added function comment.

PR #1004, opened 2023-01-13
[Event Date Api] Merge country data to Earth

chejennifer:
- the event date cache only goes up to Country level so need to merge dates when requesting Earth or continents.


shifucun (approved):

internal/server/recon/golden/bulk_find_entities/result.json
```diff
@@ -26,10 +26,7 @@
       ]
     },
     {
-      "description": "mmountan view",
-      "dcids": [
-        "geoId/0649670"
-      ]
+      "description": "mmountan view"
```
shifucun, line 29: this test is flaky due to maps api, maybe we should remove it @Spaceenter ?

Spaceenter: Right, Maps Place API is flaky... Do we completely remove the golden test for (bulk)findEntities?
Also add @pradh for visibility.

Spaceenter (approved):

internal/server/recon/golden/bulk_find_entities/result.json
shifucun: Sure, let's remove it for now to unblock.

chejennifer: done - removed

PR #1009, opened 2023-01-19
[Event API] trim space when reading filter values

chejennifer:
- trim space between unit and value when reading property values for filtering

e.g., https://autopush.datacommons.org/browser/fire/irwinId/b62ca75d-b14b-4c5d-b792-6f0d0d7a443a has a space between the unit and the value for the property value "area"


internal/server/v1/event/golden/event_collection_test.go
```diff
@@ -74,6 +74,16 @@ func TestEventCollection(t *testing.T) {
 				8,
 				"FireEvent_CA_202210.json",
 			},
+			{
+				"WildlandFireEvent",
+				"Earth",
+				"2022-10",
```
shifucun, line 80: The added test file has 100K lines. As this is only for testing purpose, can we make the filter here even more like (7,8) to reduce the size?

shifucun (approved):

internal/server/v1/event/golden/event_collection_test.go
chejennifer: done - updated filter & changed from Earth to California to cut down test file size

PR #1011, opened 2023-01-20
Filter out event without the desired property

shifucun:


Spaceenter (approved):

shifucun:
@beets @chejennifer please review the change too.

PR #1015, opened 2023-02-01
Update stat var group cache merging logic

shifucun:


pradh (approved):
Given that we expect the custom import group to be its own SVG hierarchy (with all distinct SVG nodes), I think using an SVG content from the preferred cache should be good.

PR #1016, opened 2023-02-02
[SVG Cache] Link custom root to dc/g/Root

shifucun:
Custom import group does not contains base schemas, so the SVG cache won't have "dc/g/Custom_Root" as child svg of "dc/g/Root".

Here explicitly add this link.

### DIFF
--- a/internal/server/statvar/statvar_group.go
+++ b/internal/server/statvar/statvar_group.go
@@ -32,7 +32,7 @@ const (
 	// SvgRoot is the root stat var group of the hierarchy. It's a virtual entity
 	// that links to the top level category stat var groups.
 	SvgRoot       = "dc/g/Root"
-	CustomSvgRoot = "dc/g/Custom_Root"
+	customSvgRoot = "dc/g/Custom_Root"
 )
 
 // Note this function modifies validSVG inside.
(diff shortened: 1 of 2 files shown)


internal/server/statvar/statvar_group.go
```diff
@@ -160,18 +161,35 @@ func GetStatVarGroup(
...
 			}
 		}
+		if customRootNode != nil {
+			result.StatVarGroups[SvgRoot].ChildStatVarGroups = append(
+				result.StatVarGroups[SvgRoot].ChildStatVarGroups,
+				&pb.StatVarGroupNode_ChildSVG{
+					Id:                customSvgRoot,
+					SpecializedEntity: customRootNode.AbsoluteName,
+				},
+			)
```
pradh, line 190: Just to confirm, should we check that the SvgRoot does not already have customSvgRoot?  For example, in old Stanford custom DC if schema was part of custom DC that can happen.

pradh (approved):
nice!

internal/server/statvar/statvar_group.go
shifucun: That's good to know. Added a check.

PR #1017, opened 2023-02-03
Revive Minikube for local mixer development

shifucun:
Mixer GKE deployment has migrated to Helm from Kustomize. 

As the original minikube setup is based on Kustomize, I forked a deployment file under `local` folder. Since the mixer deployment config rarely changes, this fork is okay to maintain.

### DIFF
--- a/deploy/local/custom_bigtable_info.yaml
+++ b/deploy/local/custom_bigtable_info.yaml
@@ -1,0 +1,5 @@
+project: <project_id>
+instance: <instance_id>
+tables:
+  - <table1>
+  - <table2>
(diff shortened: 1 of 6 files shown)


Fructokinase (approved):
The API compiler is neat! Out of curiosity, how does the ESP container know to wait for the compiler to finish before starting?

shifucun:
> The API compiler is neat! Out of curiosity, how does the ESP container know to wait for the compiler to finish before starting?

The api complier runs during docker build, so in deploy stage, it already exists in the volume

PR #1024, opened 2023-02-10
[EventCollection] fix default value (0) latitudes and longitudes

chejennifer:
- when `optional` is not set for a proto field, default values (in this case, 0) will be dropped

https://protobuf.dev/programming-guides/field_presence/#presence-in-proto3-apis
TODO: we probably want to set other numeric values in the protos also to `optional`


shifucun:
Just to confirm, this means all the lat,long would be set in the cache. Otherwise it can be incorrectly represented as 0.

shifucun (approved):

proto/v1/event.proto
```diff
@@ -20,8 +20,8 @@ package datacommons.v1;
 message EventCollection {
   message GeoLocation {
     message Point {
-      double latitude = 1;
-      double longitude = 2;
+      optional double latitude = 1;
```
beets, line 23: please add a note here so we don't drop it in the future

beets (commented):
thx for the fix!

proto/v1/event.proto
chejennifer: done in a followup PR: https://github.com/datacommonsorg/mixer/pull/1025

PR #1027, opened 2023-02-10
Make place import group rank higher

Spaceenter:


beets (commented):
thanks! could you also add a test in [https://github.com/datacommonsorg/mixer/blob/master/internal/server/v1/propertyvalues/golden/bulk_property_values_out_test.go](this file) for typeOf geoId/06 == state?

PR #1031, opened 2023-02-15
fix typo

n-h-diaz:


Fructokinase (approved):

beets:
curious: why is the change only in a generated file?

n-h-diaz:
> curious: why is the change only in a generated file?

sorry I didn't realize this was a generated file before so had manually modified it

PR #1033, opened 2023-02-17
Update infrequent version, as well as place group rank order.

Spaceenter:


shifucun (approved):

internal/store/bigtable/group.go
```diff
@@ -32,9 +32,9 @@ var groupRank = map[string]int{
 	"ipcc":       2,
 	"biomedical": 3,
 	"disaster":   4,
-	"place":   		100,
 	"borgcron":   10000, // Used for legacy base cache
 	"infrequent": 10000,
+	"place":      11000,
```
shifucun, line 37: Is this the final step for the migration?

Spaceenter: Correct.

Spaceenter: I guess the final final step is to roll out to prod... @pradh

PR #1034, opened 2023-02-17
Fix bulk linked observation bug

shifucun:
Make the code more robust by not relying on cache build behavior as different import groups (place, stat1, stat2) can use different data snapshot


pradh (approved):
Thank you for the quick fix!  Could you please describe the issue and fix maybe?

PR #1036, opened 2023-02-23
Support fetching all stat var group info

shifucun:
This is not a full solution since it does not take into account of constrained entities. I will come back to that later.


juliawu (approved):
Thanks for the quick fix!

PR #1038, opened 2023-02-24
Add minimum number of entities for svg nodes

chejennifer:
Add field to indicate number of entities that should have data for the stat vars and stat var groups returned for VariableGroupInfo


internal/server/statvar/statvar_group.go
```diff
@@ -271,31 +273,37 @@ func GetStatVarGroupNode(
 		if err != nil {
 			return nil, err
 		}
+		// We want at least 1 entity to have data.
+		if numEntitiesExistence == 0 {
+			numEntitiesExistence = 1
+		}
```
shifucun, line 279: move this to line 239?

shifucun (approved):

internal/server/statvar/statvar_group.go
chejennifer: done

PR #1042, opened 2023-02-27
[Custom DC] Fix custom SVG hierarchy

Fructokinase:
Currently, if multiple import groups have overlapping schemas, the hierarchy is not merged correctly. 
In addition, even when there is only 1 custom import group, the hierarchy is wrong (See [current stanford instance](https://screenshot.googleplex.com/9EGZoY9CTbhTG3p)).

I can't seem to figure out what determines when a svg shows up in cache (ex: dc/g/Custom_COVID-19_Incidence shows up in deepsolar import?), so a fix for this involves in merging all overlapping stat-vars and child svgs for a particular node, and then recounting all the nodes top down starting from the custom root. 

This fixes current Stanford instance's problem for multiple import groups.

TESTED=[local website, local mixer on envoy](https://screenshot.googleplex.com/3ShFhyPUdPpEM8W)

### DIFF
--- a/internal/server/statvar/statvar_group.go
+++ b/internal/server/statvar/statvar_group.go
@@ -116,48 +116,47 @@ func filterSVG(in *pb.StatVarGroups, statVars []string) *pb.StatVarGroups {
 	return result
 }
 
-// fixCustomRootDescendentStatVarCount returns the DescendentStatVarCount for node curID.
+// adjustDescendentStatVarCount returns the DescendentStatVarCount for node curID.
 // Mutates DescendentStatVarCount for all nodes in id2Node starting at customSvgRoot.
-func fixCustomRootDescendentStatVarCount(id2Node map[string]*pb.StatVarGroupNode, curID string) int32 {
+func adjustDescendentStatVarCount(id2Node map[string]*pb.StatVarGroupNode, curID string) int32 {
 	curNode, ok := id2Node[curID]
 	if !ok {
 		return 0
 	}
-	var descendentStatVarCount int32
+	curNode.DescendentStatVarCount = int32(len(curNode.ChildStatVars))
 	for _, childSVG := range curNode.GetChildStatVarGroups() {
 		// Child SVG protos have its own DescendentStatVarCount that must be set.
-		childSVG.DescendentStatVarCount = fixCustomRootDescendentStatVarCount(id2Node, childSVG.GetId())
-		descendentStatVarCount += childSVG.DescendentStatVarCount
+		childSVG.DescendentStatVarCount = adjustDescendentStatVarCount(id2Node, childSVG.GetId())
+		curNode.DescendentStatVarCount += childSVG.DescendentStatVarCount
 	}
-	curNode.DescendentStatVarCount = descendentStatVarCount + int32(len(curNode.ChildStatVars))
 	return curNode.GetDescendentStatVarCount()
 }
 
-// mergeCustomSVGNode merges n2's svs and child svgs into n1.
+// mergeCustomSVGNode merges node2's svg and svs into node1.
 // Merge here refers to set operation.
-// Warning: n1's DescendentStatVarCount is unchanged.
+// Warning: node1's DescendentStatVarCount is unchanged.
 // Caller is responsible for calling fixCustomRootDescendentStatVarCount
 // at the end of all merges.
-func mergeCustomSVGNode(n1, n2 *pb.StatVarGroupNode) {
-	n1SVGs := map[string]bool{}
-	for _, childSVG := range n1.GetChildStatVarGroups() {
-		n1SVGs[childSVG.GetId()] = true
+func mergeCustomSVGNode(node1, node2 *pb.StatVarGroupNode) {
+	node1SVGs := map[string]bool{}
+	for _, childSVG := range node1.GetChildStatVarGroups() {
+		node1SVGs[childSVG.GetId()] = true
 	}
-	for _, childSVG := range n2.GetChildStatVarGroups() {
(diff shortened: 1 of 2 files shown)


internal/server/statvar/statvar_group.go
```diff
@@ -114,6 +116,52 @@ func filterSVG(in *pb.StatVarGroups, statVars []string) *pb.StatVarGroups {
...
+	for _, childSVG := range curNode.GetChildStatVarGroups() {
+		// Child SVG protos have its own DescendentStatVarCount that must be set.
+		childSVG.DescendentStatVarCount = fixCustomRootDescendentStatVarCount(id2Node, childSVG.GetId())
+		descendentStatVarCount += childSVG.DescendentStatVarCount
+	}
+	curNode.DescendentStatVarCount = descendentStatVarCount + int32(len(curNode.ChildStatVars))
+	return curNode.GetDescendentStatVarCount()
+}
+
+// mergeCustomSVGNode merges n2's svs and child svgs into n1.
```
shifucun, line 136: might just use node1 and node2?

internal/server/statvar/statvar_group.go
```diff
@@ -114,6 +116,52 @@ func filterSVG(in *pb.StatVarGroups, statVars []string) *pb.StatVarGroups {
...
+// Merge here refers to set operation.
+// Warning: n1's DescendentStatVarCount is unchanged.
+// Caller is responsible for calling fixCustomRootDescendentStatVarCount
+// at the end of all merges.
+func mergeCustomSVGNode(n1, n2 *pb.StatVarGroupNode) {
```
shifucun, line 141: Can you use this function https://github.com/datacommonsorg/mixer/blob/master/internal/util/util.go#L289 to merge child SVG and child SV

internal/server/statvar/statvar_group.go
```diff
@@ -114,6 +116,52 @@ func filterSVG(in *pb.StatVarGroups, statVars []string) *pb.StatVarGroups {
 	return result
 }
 
+// fixCustomRootDescendentStatVarCount returns the DescendentStatVarCount for node curID.
+// Mutates DescendentStatVarCount for all nodes in id2Node starting at customSvgRoot.
+func fixCustomRootDescendentStatVarCount(id2Node map[string]*pb.StatVarGroupNode, curID string) int32 {
```
shifucun, line 121: This in theory works even for non-custom hierarchy. So maybe name this to 

`adjustDescendentStatVarCount(nodeMap map[string]*pb. StatVarGroupNode, node string) int32`

internal/server/statvar/statvar_group.go
```diff
@@ -114,6 +116,52 @@ func filterSVG(in *pb.StatVarGroups, statVars []string) *pb.StatVarGroups {
...
+	curNode, ok := id2Node[curID]
+	if !ok {
+		return 0
+	}
+	var descendentStatVarCount int32
```
shifucun, line 126: curNode. DescendentStatVarCount := int32(len(curNode.ChildStatVars))
..., then can save line 132

internal/server/statvar/statvar_group.go
Fructokinase: done.

internal/server/statvar/statvar_group.go
Fructokinase: I tried but I'm not sure if it was cleaner.  The MergeDedupe function merges strings but here we need to creating a string slice list of just the ids, a reference of the id -> child SVG and id -> child SV, and then loop through eveything again. We could try to make MergeDedup to use Generics but that may be for later.

shifucun: I see, here try to merge object using ID to dedup.

shifucun (approved):

internal/server/statvar/statvar_group.go
```diff
@@ -114,6 +116,51 @@ func filterSVG(in *pb.StatVarGroups, statVars []string) *pb.StatVarGroups {
...
+		childSVG.DescendentStatVarCount = adjustDescendentStatVarCount(id2Node, childSVG.GetId())
+		curNode.DescendentStatVarCount += childSVG.DescendentStatVarCount
+	}
+	return curNode.GetDescendentStatVarCount()
+}
+
+// mergeSVGNodes merges node2's svg and svs into node1.
+// Merge here refers to set operation.
+// Warning: node1's DescendentStatVarCount is unchanged.
+// Caller is responsible for calling fixCustomRootDescendentStatVarCount
```
pradh, line 138: Where is fixCustomRootDescendentStatVarCount, or you mean adjustDescendentStatVarCount?

internal/server/statvar/statvar_group.go
```diff
@@ -114,6 +116,51 @@ func filterSVG(in *pb.StatVarGroups, statVars []string) *pb.StatVarGroups {
...
+// at the end of all merges.
+func mergeSVGNodes(node1, node2 *pb.StatVarGroupNode) {
```
pradh, line 140: I wonder if this should be recursive?

internal/server/statvar/statvar_group.go
Fructokinase: Oops I need to update the comment, it should be adjustDescendentStatVarCount

internal/server/statvar/statvar_group.go
Fructokinase: This is not recursive because the calling function happens while the id -> [child node] map is being updated. This is caller per update for the merging of two nodes. 

As spoken offline, for main DC, we can take the max. On second though, I think we should take the sum instead of max? ex: node1[count=2] and node2[count=3], the actual count is between 3 and 5.

PR #1043, opened 2023-02-27
Increase envoy buffer

Fructokinase:
Some requests are massive, ex: http://localhost:8081/v1/bulk/info/variable-group

### DIFF
--- a/esp/envoy-config.yaml
+++ b/esp/envoy-config.yaml
@@ -19,6 +19,8 @@ static_resources:
   - name: restlistener
     address:
       socket_address: { address: 0.0.0.0, port_value: 8081 }
+    # Reuqest buffer size limit
+    per_connection_buffer_limit_bytes: 1073741824 # 1GB
     filter_chains:
     - filters:
       - name: envoy.filters.network.http_connection_manager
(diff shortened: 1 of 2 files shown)


esp/envoy-config.yaml
```diff
@@ -19,6 +19,8 @@ static_resources:
   - name: restlistener
     address:
       socket_address: { address: 0.0.0.0, port_value: 8081 }
+    # Reuqest buffer size limit
```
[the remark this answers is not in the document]
Fructokinase: fixed, thanks!

esp/envoy-config.yaml
```diff
@@ -19,6 +19,8 @@ static_resources:
   - name: restlistener
     address:
       socket_address: { address: 0.0.0.0, port_value: 8081 }
+    # Request buffer size limit
+    per_connection_buffer_limit_bytes: 1073741824 # 1GB
```
shifucun, line 23: How do we make this work also in ESP v1? Or shall we try to move to ESP v2?

shifucun (approved):

esp/envoy-config.yaml
Fructokinase: I think we should, envoy is much better built than nginx as a proxy. 

For espv2, this is the docker lag:  --envoy_connection_buffer_limit_bytes
https://github.com/GoogleCloudPlatform/esp-v2/pull/344/files

PR #1047, opened 2023-03-02
Log requests in envoy; Some envoy config/docs formatting

shifucun:


Fructokinase (approved):
This is useful, thanks!

PR #1048, opened 2023-03-02
Fix duplicate count of descendent stat vars after merging svg nodes

shifucun:
Also apply the merging to all the SVG (not only the custom SVG)

The count adjustment takes 3-4 seconds locally.

@pradh With this, we won't need to set the descendent count in Flume


internal/server/statvar/statvar_group.go
```diff
@@ -220,7 +228,7 @@ func GetStatVarGroup(
 						}
 						if _, ok := result.StatVarGroups[k]; !ok {
 							result.StatVarGroups[k] = v
-						} else if strings.HasPrefix(k, customSVGPrefix) {
+						} else {
 							// For custom SVGs, merge all SVGs regardless of the import group rank.
```
pradh, line 232: Update comment?

internal/server/statvar/statvar_group.go
```diff
@@ -270,6 +276,8 @@ func GetStatVarGroup(
 			},
 		)
 	}
+	// Recount all descendent stat vars after merging
+	adjustDescendentSVCount(result.StatVarGroups, SvgRoot)
```
pradh, line 280: If we're computing  the counts here, is it okay to refer to those above, like:

`result.StatVarGroups[SvgRoot].DescendentStatVarCount += customRootNode.DescendentStatVarCount`

internal/server/statvar/statvar_group.go
```diff
@@ -116,20 +115,29 @@ func filterSVG(in *pb.StatVarGroups, statVars []string) *pb.StatVarGroups {
...
+	for _, sv := range curNode.ChildStatVars {
+		descendentSV[sv.Id] = struct{}{}
 	}
-	curNode.DescendentStatVarCount = int32(len(curNode.ChildStatVars))
 	for _, childSVG := range curNode.GetChildStatVarGroups() {
+		childDescendentSV := adjustDescendentSVCount(id2Node, childSVG.GetId())
 		// Child SVG protos have its own DescendentStatVarCount that must be set.
-		childSVG.DescendentStatVarCount = adjustDescendentStatVarCount(id2Node, childSVG.GetId())
-		curNode.DescendentStatVarCount += childSVG.DescendentStatVarCount
+		childSVG.DescendentStatVarCount = int32(len(childDescendentSV))
```
pradh, line 134: Slightly puzzled why the setting in line 139 doesn't handle this?  I'm assuming `childSVG` is same as `id2Node[childSVG.GetId()]` ?

pradh:
> Also apply the merging to all the SVG (not only the custom SVG)
> 
> The count adjustment takes 3-4 seconds locally.
> 
> @pradh With this, we won't need to set the descendent count in Flume

Nice simplification!  Does this cover all API entries from SV hierarchy?

pradh (approved):

internal/server/statvar/statvar_group.go
```diff
@@ -116,20 +115,29 @@ func filterSVG(in *pb.StatVarGroups, statVars []string) *pb.StatVarGroups {
...
-		return 0
+		return nil
+	}
+	descendentSV := map[string]struct{}{}
+	for _, sv := range curNode.ChildStatVars {
+		descendentSV[sv.Id] = struct{}{}
 	}
-	curNode.DescendentStatVarCount = int32(len(curNode.ChildStatVars))
 	for _, childSVG := range curNode.GetChildStatVarGroups() {
+		childDescendentSV := adjustDescendentSVCount(id2Node, childSVG.GetId())
```
Fructokinase, line 132: Each recursion creates multiple levels of the id->child map in memory and I'm afraid this is O(d * len(c)), d=depth and c being the max child per node. Could this cause OOM on a container?

Alternatively, descendentSV := map[string]struct{}{} can be declared before the first recursion call and descendentSV can be passed down to every recursive call.

internal/server/statvar/statvar_group.go
shifucun: The structure is like:
```
{
  DescendentStatVarCount: 20,
  GetChildStatVarGroups: [
  {
     "Id": "1"
     "DescendentStatVarCount": 2
   },
  {
     "Id": "2"
     "DescendentStatVarCount": 3
   },
  ]
}
```

So there are two DescendentStatVarCount to set, this one being the children count.

internal/server/statvar/statvar_group.go
shifucun: Done

internal/server/statvar/statvar_group.go
shifucun: good catch, that one is not needed anymore

internal/server/statvar/statvar_group.go
shifucun: I think the set from a parent SVG can not be passed to all the child SVG, so each child would always need to create its own set (and hopefully GC will kick in when the recursion returns from children SVG).

shifucun:
> 

Yes, it covers all entries.

pradh:
> > 
> 
> Yes, it covers all entries.

What about existence check cache?  http://google3/datacommons/proto/prophet/cache_data.proto;l=347-357;rcl=513107256

internal/server/statvar/statvar_group.go
pradh: Ahh ok.  I think I was looking at the cache proto, but the mixer proto has additional field :)

shifucun:
> 

This is in a separate API. I can check afterwards about it.

PR #1049, opened 2023-03-02
Remove API /place/metadata which is replaced by /v1/info/place

shifucun:
This API has no usage in the past 14 days in both website and mixer.


Fructokinase (approved):
Approval on the condition that we update all existing service definitions in ESP. For example, if we don't update endpoints first in autopush before committing, will things break?

shifucun:
Thanks!

shifucun:
> Approval on the condition that we update all existing service definitions in ESP. For example, if we don't update endpoints first in autopush before committing, will things break?

The autopush will build grpc binary as well and release a new esp version before GKE update, so it should be fine?

PR #1052, opened 2023-03-09
Remove unused field related_places

Spaceenter:


Spaceenter:
This field is removed from cache proto too.

shifucun (approved):

PR #1053, opened 2023-03-10
[V2] Query parser: add split function

shifucun:


internal/server/v2/parser.go
```diff
@@ -0,0 +1,42 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package v2
+package v2
+
+import "fmt"
+
+// split a string by "->" and "<-"
+func split(s string) ([]string, error) {
```
Spaceenter, line 21: I think we also need to split by "," and/or ":" etc., when parsing other parts of the query. So maybe make the function name more specific?

internal/server/v2/parser.go
```diff
@@ -0,0 +1,42 @@
...
+	if len(s) < 2 {
+		return nil, fmt.Errorf("invalid query string: %s", s)
```
Spaceenter, line 23: Wonder if we want to use more formal error syntax. For example:

status.Errorf(codes.InvalidArgument, ".....")

https://pkg.go.dev/google.golang.org/grpc/codes

Spaceenter (approved):

internal/server/v2/parser.go
shifucun: changed to splitArc

internal/server/v2/parser.go
shifucun: Done

PR #1057, opened 2023-03-15
[V2 API] Parse arc into a Graph object

shifucun:


internal/server/v2/parser.go
```diff
@@ -45,3 +47,91 @@ func splitArc(s string) ([]string, error) {
...
+	if len(s) == 0 {
+		return arc, nil
+	}
+	if s[0] == '[' {
+		if s[len(s)-1] != ']' {
+			return nil, status.Errorf(
+				codes.InvalidArgument, "invalid filter string: %s", s)
+		}
+		s = s[1 : len(s)-1]
+		// [prop1, prop2]
```
Spaceenter, line 76: At this point, [ and ] are already removed, so:   prop1, prop2 ?

internal/server/v2/parser.go
```diff
@@ -45,3 +47,91 @@ func splitArc(s string) ([]string, error) {
...
+		props := []string{}
+		pos := -1
+		for i := 0; i < len(s)-1; i++ {
+			if s[i] == ',' {
+				props = append(props, strings.TrimSpace(s[pos+1:i]))
+				pos = i
+			}
+		}
+		props = append(props, strings.TrimSpace(s[pos+1:]))
+		arc.props = props
```
Spaceenter, line 86: A shorter version...

```
arc.props = strings.Split(strings.ReplaceAll(s, " ", ""), ",")
```

internal/server/v2/parser.go
```diff
@@ -45,3 +47,91 @@ func splitArc(s string) ([]string, error) {
...
+		return arc, nil
+	}
+	for i := 0; i < len(s); i++ {
```
Spaceenter, line 89: Add more comments on what each block is doing?

internal/server/v2/parser.go
```diff
@@ -45,3 +47,91 @@ func splitArc(s string) ([]string, error) {
...
+		if s[i] == '+' {
+			arc.prop = s[0:i]
+			arc.wildcard = "+"
+			s = s[i+1:]
+			break
```
Spaceenter, line 94: Is it possible to have the following pattern?

->prop1+{prop2: val2}

If so, probably 'break' is wrong here. If not, probably add a comment somewhere that the above pattern is not supported?

internal/server/v2/parser.go
```diff
@@ -45,3 +47,91 @@ func splitArc(s string) ([]string, error) {
...
+		}
+	}
+	if len(s) > 0 && s[0] == '{' {
+		if s[len(s)-1] != '}' {
+			return nil, status.Errorf(
+				codes.InvalidArgument, "invalid filter string: %s", s)
+		}
+		// {prop1:val1, prop2:val2}
+		filter := map[string]string{}
+		parts := strings.Split(s[1:len(s)-1], ",")
```
Spaceenter, line 108: parts := strings.Split(strings.ReplaceAll(s[1:len(s)-1], " ", ""), ",")

Then, you can remove all the strings.TrimSpace() calls below.

internal/server/v2/parser.go
```diff
@@ -45,3 +47,91 @@ func splitArc(s string) ([]string, error) {
...
+			if len(kv) != 1 {
+				return nil, status.Errorf(
+					codes.InvalidArgument, "invalid filter string: %s", p)
+			}
+			k := strings.TrimSpace(p[0:kv[0]])
+			v := strings.TrimSpace(p[kv[0]+1:])
+			if len(k) == 0 || len(v) == 0 {
+				return nil, status.Errorf(
+					codes.InvalidArgument, "invalid filter string: %s", p)
+			}
```
Spaceenter, line 126: Can we just use  strings.Split(p, ":") here? Then check if there are 2 non-empty items.

Spaceenter (changes requested):

internal/server/v2/parser.go
shifucun: Moved the comment to the beginning of the block

internal/server/v2/parser.go
shifucun: much better!

internal/server/v2/parser.go
shifucun: Done

internal/server/v2/parser.go
shifucun: The above pattern should work. This block process upto "prop1+". Then from line101, it handles {prop2: val2}

internal/server/v2/parser.go
shifucun: updated!

internal/server/v2/parser.go
shifucun: that's definitely much better

internal/server/v2/graph.go
```diff
@@ -0,0 +1,30 @@
...
+
+// Package v2 is the version 2 of the Data Commons REST API.
+package v2
+
+// Arc represents an arc in the graph.
+type Arc struct {
+	// Whether it's out or in arc.
+	out bool
+	// The property of the arc. This is when property is specified without []
+	prop string
```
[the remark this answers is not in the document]
shifucun: Use "singleProp" and "bracketProps" to make the initial query intention clear. For "containedInPlace+" it uses single prop but implies a list of props, so it's better to have two terms to make this clear.

internal/server/v2/parser.go
```diff
@@ -45,3 +47,73 @@ func splitArc(s string) ([]string, error) {
...
+	if len(s) > 0 && s[0] == '{' {
+		if s[len(s)-1] != '}' {
+			return nil, status.Errorf(
+				codes.InvalidArgument, "invalid filter string: %s", s)
+		}
+		filter := map[string]string{}
+		parts := strings.Split(strings.ReplaceAll(s[1:len(s)-1], " ", ""), ",")
+		for _, p := range parts {
+			kv := strings.Split(p, ":")
+			if len(kv) != 2 {
```
Spaceenter, line 105: if len(kv) != 2 || kv[0] == "" || kv[1] == ""

Spaceenter (approved):

internal/server/v2/parser.go
shifucun: Done

PR #1058, opened 2023-03-20
[V2 API] Property Values routing and simple query handling

shifucun:
- Use EntityInfo in the generic query response. Keep "types" and "name" fields for now, but I think they should be removed eventually to make this response generic.
- Make routing for simple property value query and added a few golden tests from V1 test cases.


proto/v2/v2.proto
```diff
@@ -19,11 +19,18 @@ package datacommons.v2;
...
-message PropertyValuesV2Request {
+message Arc {
+  string property = 1;
+  repeated datacommons.EntityInfo nodes = 2;
+}
+
+message QueryV2Request {
   repeated string nodes = 1;
-  string property = 2;
+  string graph = 2;
```
Spaceenter, line 29: Add some comments and examples for what is "graph"?

internal/server/handler_v2.go
```diff
@@ -18,14 +18,57 @@ package server
...
-// PropertyValuesV2 implements API for mixer.PropertyValuesV2.
-func (s *Server) PropertyValuesV2(
-	ctx context.Context, in *pb.PropertyValuesV2Request,
-) (*pb.PropertyValuesV2Response, error) {
-	return v2pv.API(ctx, in, s.store)
+// QueryV2 implements API for mixer.QueryV2.
+func (s *Server) QueryV2(
+	ctx context.Context, in *pb.QueryV2Request,
+) (*pb.QueryV2Response, error) {
+	arcStrings, err := v2.SplitArc(in.Graph)
```
Spaceenter, line 32: Style thing: always prefer using Get function, to avoid nil pointer..., so:

in.GetGraph()

internal/server/handler_v2.go
```diff
@@ -18,14 +18,57 @@ package server
...
+	arcs := []*v2.Arc{}
+	for _, s := range arcStrings {
+		arc, err := v2.ParseArc(s)
+		if err != nil {
+			return nil, err
+		}
+		arcs = append(arcs, arc)
+	}
+	// Simple Property Values
+	if len(arcs) == 1 {
```
Spaceenter, line 45: I assume you'll add the case "len(arcs) != 1" in a future PR. So the logic here becomes pretty long, can we abstract out to a function? Ideally hander.go only does routing and does not contain complex logic?

Spaceenter (approved):

proto/v2/v2.proto
```diff
@@ -19,11 +19,22 @@ package datacommons.v2;
...
+message Arc {
+  string property = 1;
+  repeated datacommons.EntityInfo nodes = 2;
+}
+
+message QueryV2Request {
+  // Nodes to query for.
   repeated string nodes = 1;
+  // The expression for the property or the chained property.
   string property = 2;
```
pradh, line 31: Should we include a summary of the language we decided here?

proto/v2/v2.proto
```diff
@@ -19,11 +19,22 @@ package datacommons.v2;
...
+  // Max number of entries to be returned.
```
pradh, line 32: What is an "entry"?  Is it nodes or triples?

pradh (approved):
Nice!  Mostly some comments about comments :)

internal/server/handler_v2.go
```diff
@@ -18,14 +18,58 @@ package server
...
+	}
+	// TODO: abstract this out to a router module.
+	// Simple Property Values
+	if len(arcs) == 1 {
+		arc := arcs[0]
+		direction := util.DirectionOut
+		if !arc.Out {
+			direction = util.DirectionIn
+		}
+		if arc.SingleProp != "" && arc.Wildcard == "" {
```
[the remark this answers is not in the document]
shifucun: That would the case of "containedInPlace+", which needs to be handled separately.

proto/v2/v2.proto
```diff
@@ -19,11 +19,22 @@ package datacommons.v2;
 
 import "entity.proto";
 
-message PropertyValuesV2Request {
+message Arc {
+  string property = 1;
```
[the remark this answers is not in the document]
shifucun: This is a single property. Added comment

proto/v2/v2.proto
shifucun: Done

proto/v2/v2.proto
shifucun: It's nodes.

proto/v2/v2.proto
```diff
@@ -19,11 +19,22 @@ package datacommons.v2;
...
+  // Max number of entries to be returned.
+  int32 limit  = 3;
+  // Pagination token
+  string next_token = 4;
 }
 
-message PropertyValuesV2Response {
-  repeated datacommons.NodePropertyValues data = 1;
+message QueryV2Response {
+  map<string, Arc> data = 1;
```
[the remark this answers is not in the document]
shifucun: Added comments.

shifucun:
Thank you both for the review!

PR #1060, opened 2023-03-22
Load place info in store for place recognition.

Spaceenter:


proto/recon.proto
```diff
@@ -145,3 +145,15 @@ message RecognizePlacesResponse {
...
   repeated Item items = 1;
 }
+
+// RecogPlace contains info of a place for RecognizePlaces method.
+message RecogPlace {
+  repeated string words = 1;
+  string dcid = 2;
+  repeated string containing_places = 3;
+  int64 population = 4;
+}
```
pradh, line 155: Just wonder why this is proto and not struct?  Also, is recon.proto an internal proto

proto/recon.proto
```diff
@@ -145,3 +145,15 @@ message RecognizePlacesResponse {
   }
   repeated Item items = 1;
 }
+
+// RecogPlace contains info of a place for RecognizePlaces method.
+message RecogPlace {
+  repeated string words = 1;
```
pradh, line 151: These parts of the name right?  We may soon want to add multiple names.  So either make it `repeated string names = 1;` that we split,
or 

```
message NameParts {
  repeated string words = 1;
}

repeated NameParts names = 1;
```

pradh (approved):
Thanks!

proto/recon.proto
Spaceenter: This is to make unit test easier. Deep comparison of proto messages is cleaner than deep comparison of structs.
recon.proto could be public I think? I don't see a reason why we want to keep it private.

proto/recon.proto
Spaceenter: done

PR #1062, opened 2023-03-23
Support multiple names for RecogPlaceMap.

Spaceenter:


internal/store/files/recog_place_map.go
```diff
@@ -63,19 +63,24 @@ func RecogPlaceMap() (map[string]*pb.RecogPlaces, error) {
...
+		if strings.TrimSpace(record[2]) == "" {
 			return nil, status.Errorf(codes.FailedPrecondition,
-				"Empty name parts for CSV record: %v", record)
+				"Empty names for CSV record: %v", record)
 		}
-		name := &pb.RecogPlace_Name{}
-		for _, namePart := range nameParts {
-			name.Parts = append(name.Parts,
-				strings.ToLower(strings.TrimSpace(namePart)))
+		for _, name := range strings.Split(record[2], ",") {
```
pradh, line 71: I slightly wonder if we should split with double quotes escape?

For example: 
image

pradh (commented):
Thanks!

internal/store/files/recog_place_map.go
Spaceenter: Do you mean the potential double quotes like "name1,name2,name3"? If so, Go's CSV reader should handle that already:
e.g.
https://go.dev/play/p/0oCACwV1gzf

pradh: > 

Sorry I meant, the names already have a comma in them (per screenshot).  So maybe we need better escaping with double-quotes (like we recommend for our import CSV [here](https://github.com/datacommonsorg/data/blob/master/docs/formatting_tips.md#emitting-dc-repeated-values)), or should do one name per row?

Spaceenter: I think the CSV package can handle commas within double quotes, e.g. https://go.dev/play/p/dhTFA0j0STS (Note "Pike1,Pike2,Pike3") -- is this what you mean?

Spaceenter: @pradh

pradh: Have a look at https://datacommons.org/browser/nuts/TR82 (in this case, its a parent place taking the name of 3 children places).  The name itself has commas.  But we probably don't want to treat that as different names, but parts of the same name.

The option we do in the import script is:
`"\"Kastamonu, Çankiri, Sinop\", Another name, \"Another name, with comma\""`
That's 3 names, two of them have commas.

But another option is to perhaps remove/replace the commas in the KG while generating the CSV.  If so, then this PR LG.

Does that make sense?

Spaceenter: Wow, OK, that's a very interesting special case...
I think a deeper question is, whether we want to include the commas as part of the name tokens? The current place recognition code will split by comma, so a name like this cannot be recognized by the current place recognition algorithm?

pradh: Yeah, you're right.  Maybe we should be sure to ignore commas in a place name before adding to CSV.

pradh (approved):

PR #1063, opened 2023-03-23
Add util functions tokenize and findPlaceCandidates for place recogni…

Spaceenter:
…tion.

### DIFF
--- a/internal/server/recon/recognize_places.go
+++ b/internal/server/recon/recognize_places.go
@@ -17,6 +17,7 @@ package recon
 
 import (
 	"context"
+	"strings"
 
 	pb "github.com/datacommonsorg/mixer/internal/proto"
 	"github.com/datacommonsorg/mixer/internal/store"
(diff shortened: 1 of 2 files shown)


internal/server/recon/recognize_places.go
```diff
@@ -29,3 +30,97 @@ func RecognizePlaces(
 	// TODO(ws): Implement.
 	return nil, nil
 }
+
+func tokenize(query string) []string {
+	tokens := []string{}
+
+	// Split by space.
+	for _, partBySpace := range strings.Split(query, " ") {
```
pradh, line 38: nit: just curious, does golang not allow iteration by char?

pradh (approved):
Thanks for the tests!

internal/server/recon/recognize_places.go
Spaceenter: Yes it can iterate by char -- just feels using 'split' is closer to how human thinks about tokenize things, and maybe slightly more readable?

PR #1064, opened 2023-03-23
Update V2 API response proto

shifucun:
Make response recursive so it can be used when multiple levels of properties are specified in the request.


proto/v2/v2.proto
```diff
@@ -19,11 +19,13 @@ package datacommons.v2;
...
-  // chained properties
-  string property = 1;
-  repeated datacommons.EntityInfo nodes = 2;
+message Nodes {
+  repeated datacommons.EntityInfo nodes = 1;
+}
+
+message Arcs {
+  map<string, Nodes> arcs = 2;
+  map<string, Arcs> neighbor = 3;
```
Spaceenter, line 28: Where is the poor 1? :)

Spaceenter: The naming is a bit confusing, especially the first one, the map val type is 'Nodes', but the variable name is 'arcs'. How about:

  map<string, Nodes> neighbor_nodes = 1;
  map<string, Arcs> neighbor_arcs = 2;

Or, just:

  map<string, Nodes> nodes = 1;
  map<string, Arcs> arcs = 2;

proto/v2/v2.proto
```diff
@@ -42,5 +44,5 @@ message QueryV2Request {
 
 message QueryV2Response {
   // Keyed by the query node dcid.
-  map<string, Arc> data = 1;
+  map<string, Arcs> data = 1;
```
Spaceenter, line 47: I slight wonder if we can just do:

Arcs data = 1;

Because inside 'Arcs' there is also a message 'map<string, Arcs>'

proto/v2/v2.proto
shifucun: I initially tried this too, but it does not look right. I just used this to update the golden and pushed and you can see the actual response. By comparing with the previous version, you can tell it was more readable.

Basically, "neighbor" is to recursive to next level of "arcs" and "arcs" is linked to leaf "nodes".

internal/server/v2/propertyvalues/golden/simple/california.json
```diff
@@ -0,0 +1,48 @@
+{
+  "data": {
+    "geoId/06": {
+      "nodes": {
+        "administrativeCapital": {
```
shifucun, line 5: note here: we would say "geoId/06" has some "arcs" with keys "administrativeCapital", "name" etc. The value of each arc is "a list of nodes" @shifucun

shifucun: Note this is not the optimal version where using "nodes" does not convey the idea of "arcs" as in my initial PR version

Spaceenter (approved):

proto/v2/v2.proto
```diff
@@ -19,11 +19,13 @@ package datacommons.v2;
...
-message Arc {
-  // The property of one single arc. This is only a single property, rather than
-  // chained properties
-  string property = 1;
-  repeated datacommons.EntityInfo nodes = 2;
+message Nodes {
+  repeated datacommons.EntityInfo nodes = 1;
+}
+
+message LinkedGraph {
```
shifucun, line 26: Change this to LinkedGraph, which is much better.

pradh: This LG.  Can we describe the keys and what the fields mean, maybe with an example?

pradh: Also, I forget what the consensus was on using map vs. list of protos (for json purposes), do we use maps in V1?

pradh (approved):
Thanks!

proto/v2/v2.proto
shifucun: Since V2 holds a subgraph in the response, it's better to use map than list. We use list in V1 and can still provide util functions to convert the raw v2 response into list representations, especially for observations.

PR #1065, opened 2023-03-24
Complete implementation of RecognizePlaces().

Spaceenter:


Spaceenter:
Doc:

https://docs.google.com/document/d/1z4sD0-UJP3TLW1L5kQZY4KdH2sZkqP4udmxrKwnVqLI

internal/server/recon/golden/recognize_places_test.go
```diff
@@ -0,0 +1,90 @@
+// Copyright 2023 Google LLC
```
pradh, line 1: nit: its a bit confusing to have two files with the same name? maybe call this integration test?

internal/server/recon/golden/recognize_places_test.go
```diff
@@ -0,0 +1,90 @@
...
+			if err != nil {
+				t.Errorf("RecognizePlaces(%s) = %s", query, err)
+				continue
+			}
+
+			if latencyTest {
+				continue
+			}
+
+			goldenFile := fmt.Sprintf("expected%d.json", idx)
```
pradh, line 64: nit: wdyt of putting the query in the `.json` file too, so we don't have to line up the index to validate? :)

As this grows we may add more tests, and it might be easier to maintain.

proto/recon.proto
```diff
@@ -148,15 +142,21 @@ message RecognizePlacesResponse {
...
   repeated string containing_places = 3;
   int64 population = 4;
 }
 
-message RecogPlaces {
-  repeated RecogPlace places = 1;
-}
+message RecogPlaces { repeated RecogPlace places = 1; }
+
+message TokenSpans {
```
pradh, line 154: This is for internal use right?  For my understanding, curious why this is a proto instead of a struct?

pradh (approved):
Thanks for the tests!

proto/recon.proto
Spaceenter: Right, this is to make the comparisons in unit test easier: 

```
cmpOpts := cmp.Options{
  protocmp.Transform(),
}
if diff := cmp.Diff(got, c.want, cmpOpts); diff != "" {
  ...
}
```

internal/server/recon/golden/recognize_places_test.go
Spaceenter: I see the confusion. Actually, I think it's probably meaningful to include the query in the response, so that we can do batch queries in the future, and it's easier to see which query input corresponds to which query response.

Changed.

internal/server/recon/golden/recognize_places_test.go
Spaceenter: The golden test file name usually corresponds to the RPC method name - so I changed the name of the code files from recognize_places.go to recognize.go, because we might need to recognize other things in the future, and those code can all go there.

PR #1071, opened 2023-03-31
[V2 API] Implement direct observations query from time series cache

shifucun:


internal/server/v2/observation/series.go
```diff
@@ -17,18 +17,115 @@ package observation
...
+						entityObservation.OrderedFacetObservations = append(
+							entityObservation.OrderedFacetObservations,
+							&pbv2.FacetObservation{
+								FacetId:      facetID,
+								Observations: obsList,
+							},
+						)
+					}
+				}
+			} else if store.MemDb.HasStatVar(variable) {
```
pradh, line 93: If memstore has variable, should that trump what's in KG?

internal/server/v2/observation/series.go
```diff
@@ -17,18 +17,115 @@ package observation
...
+					sort.SliceStable(obsList, func(i, j int) bool {
+						return obsList[i].Date < obsList[j].Date
+					})
+					if queryDate == LATEST {
+						obsList = obsList[len(obsList)-1:]
+					}
+					if len(obsList) > 0 {
+						result.Facets[facetID] = series.Metadata
+						entityObservation.OrderedFacetObservations = append(
+							entityObservation.OrderedFacetObservations,
```
pradh, line 118: If this for loop code is the same maybe can use helper?

pradh (approved):
Nice!

shifucun:
Thanks for review!

internal/server/v2/observation/series.go
pradh: That is nice!

PR #1074, opened 2023-04-01
[V2 API] Add handler for containedInPlace observation query; Update query expression parsing

shifucun:


internal/server/v2/parser.go
```diff
@@ -118,8 +116,52 @@ func ParseArc(s string) (*Arc, error) {
...
+	}
+	arcs := []*Arc{}
+	for i := 0; i < len(parts)/2; i++ {
+		arc, err := parseArc(parts[i*2+1], parts[i*2+2])
+		if err != nil {
+			return nil, err
+		}
+		arcs = append(arcs, arc)
+	}
+	g.Arcs = arcs
```
Spaceenter, line 165: Don't need to define 'arcs' here, just re-use g.Arcs:

	for i := 0; i < len(parts)/2; i++ {
		arc, err := parseArc(parts[i*2+1], parts[i*2+2])
		if err != nil {
			return nil, err
		}
		g.Arcs = append(arcs, arc)
	}

internal/server/v2/parser_test.go
```diff
@@ -178,29 +197,115 @@ func TestParseArc(t *testing.T) {
...
+				t.Errorf("ParseProperty(%s) expect error, but got nil", c.expr)
+			}
+			continue
+		}
+		if err != nil {
+			t.Errorf("ParseProperty(%s) got error %v", c.expr, err)
+			continue
+		}
+		if diff := cmp.Diff(result, c.arc); diff != "" {
+			t.Errorf("v(%s) got diff %v", c.expr, diff)
```
Spaceenter, line 271: nit:   v(%s)  -->  ParseProperty(%s)

internal/server/v2/graph.go
```diff
@@ -28,3 +28,9 @@ type Arc struct {
 	// The filter of the arc.
 	Filter map[string]string
 }
+
+// Graph represents a local graph with subject and connected arcs.
+type Graph struct {
```
Spaceenter, line 33: The name of 'Graph' seems to be too general - could it be more specific?

Spaceenter (approved):

PR #1075, opened 2023-04-03
Handle expressions with reasonable spaces in-between.

Spaceenter:


internal/server/v2/parser.go
```diff
@@ -27,10 +27,10 @@ func splitWithDelim(expr string, delim string) []string {
 	res := []string{}
```
shifucun, line 27: maybe trim all space/newline at the beginning of function splitWithDelim?

Spaceenter: Oh that's more aggressive, but I guess we can be more aggressive. Done.

PR #1076, opened 2023-04-03
minor refactor to check obsList earlier in FetchFromSeries.

Spaceenter:


internal/server/v2/observation/series.go
```diff
@@ -60,40 +60,45 @@ func FetchFromSeries(
...
+				result.Facets[facetID] = metadata
+				entityObservation.OrderedFacetObservations = append(
+					entityObservation.OrderedFacetObservations,
+					&pbv2.FacetObservation{
+						FacetId:      facetID,
+						Observations: obsList,
+					},
+				)
 			}
 			result.ObservationsByVariable[variable].ObservationsByEntity[entity] = entityObservation
```
shifucun, line 103: When len(series)==0, still need to add entityObservation to the result (line 98). If "continue", then this entry will be missing in the result.

PR #1082, opened 2023-04-04
Implement <-containedInPlace+{typeOf:X} API

Spaceenter:


internal/server/handler_v2.go
```diff
@@ -49,26 +49,51 @@ func (s *Server) V2Node(
...
+			}
+			if arc.SingleProp != "containedInPlace" {
+				return nil, status.Errorf(codes.InvalidArgument,
+					"only containedInPlace is supported for wildcard '+'")
+			}
+			if arc.Out {
+				return nil, status.Errorf(codes.InvalidArgument,
+					"only in arc is supported for containedInPlace+")
+			}
+			placeType, ok := arc.Filter["typeOf"]
```
shifucun, line 82: shall we put all the working scenarios at the beginning, so it's clear to see all the exact working scenarios. Then return errors for the rest of the case?

internal/server/handler_v2.go
```diff
@@ -49,26 +49,51 @@ func (s *Server) V2Node(
...
+			if !ok {
+				return nil, status.Errorf(codes.InvalidArgument,
+					"must provide typeOf filter for <-containedInPlace+")
+			}
+			return v2pv.LinkedPropertyValues(
```
shifucun, line 87: I suggest to make the LinkedPropertyValues API general:

LinkedPropertyValues(...., nodes, linkedProperty, filters) and then handle the placesIn case specially inside it. That way, we can later extend it to other prop/filters by doing multiple BT read.

internal/server/v2/propertyvalues/golden/linked_property_values_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	_, filename, _, _ := runtime.Caller(0)
+	goldenPath := path.Join(path.Dir(filename), "linked")
+
+	testSuite := func(mixer pbs.MixerClient, latencyTest bool) {
+		for _, c := range []struct {
+			goldenFile string
+			nodes      []string
+			property   string
+		}{
```
shifucun, line 42: Maybe use the same test cases as V1? And later we can delete the V1/V0 tests entirely without losing information.

shifucun (approved):
thanks for the update, looks very nice!

PR #1083, opened 2023-04-05
Clean up docs about minikube, skaffold and kustomize

shifucun:
Bringing up local Kubernetes cluster with Minikube is slow, subject to memory limit and is very slow. Using envoy is a much better choice for testing API locally. So no need to keep minikube (with skaffold) as an option for local development.

Also remove reference to Kustomize as it's replaced by Helm in mixer already.


docs/developer_guide.md
```diff
@@ -42,11 +36,11 @@ deploy and port forwarding.
   gcloud auth application-default login
   ```
 
-## Develop mixer locally as a Go server (Recommended)
+## Develop mixer locally as a Go server
 
 **NOTE** This can only develop and test the gRPC server. Since the
```
Fructokinase, line 41: nit: Maybe remove this note all together?

Fructokinase (approved):

docs/developer_guide.md
shifucun: Good point. Now I made the entire guide for local development

PR #1084, opened 2023-04-05
logging for DCID is invalid

shifucun:


internal/util/util.go
```diff
@@ -227,6 +227,7 @@ func CamelToSnake(str string) string {
 func CheckValidDCIDs(dcids []string) bool {
 	for _, dcid := range dcids {
 		if dcid == "" || strings.Contains(dcid, " ") || strings.Contains(dcid, ",") {
+			log.Printf("CheckValidDCIDs failed with %s", dcid)
 			return false
 		}
 	}
```
Spaceenter, line 233: A more go-style way is to:

func CheckValidDCIDs(dcids []string) error {
  if ... {
    return fmt.Errorf(...)
  }
  return nil
}

Then check if CheckValidDCIDs() is nil in the calller.

Spaceenter (approved):

internal/util/util.go

PR #1086, opened 2023-04-07
[V2 API] A unified observation API for fetching stat values and non-stat properties

shifucun:
In V2, I try to have a unified API to query observation stat value and  metrics. This PR is an attempt to upgrade V1 GetPlaceVariables and CheckObservationExistence API.

**GetPlaceVariables**
In V1, the response is a map keyed by stat var with values being a list of entity dcids.

**CheckExistence**
In V1, the response is a two level map with keys being variable and entity, and value being a boolean.

In V2, they are all converted into GetObservation() and make the API argument with SQL style.

1. The request has a new "select" fields to specify what to return. This makes it possible to use one API to fetch stat values and other properties like date and variables.

2. To get place variables, need to loop through the keys of the response, instead of a plain list. The complexity for client is similar. When multiple entities are provided, there is a "union" bit in V1 which is inherently supported (and with finer granularity) in V2: Each variable contains all the places that has observations (see the golden response for detail)

This also updated the response field from "observations_by_variable" to "by_variable" so it also suits to the Observation Metric API. It's concise and still meaning full with "observations" as the inner fields in GetObservation() API response.


internal/server/handler_v2.go
```diff
@@ -140,3 +141,21 @@ func (s *Server) V2Observation(
 	}
 	return &pbv2.ObservationResponse{}, nil
 }
+
+// V2ObservationMetric implements API for mixer.V2ObservationMetric.
+func (s *Server) V2ObservationMetric(
+	ctx context.Context, in *pbv2.ObservationRequest,
+) (*pbv2.ObservationResponse, error) {
+	if in.GetVariablesExpression() == "?" { // Get all variables for entities
```
pradh, line 149: Given the baggage of `?` in regex, using it for get all variables feels a bit odd, is there a reason we don't want to use `*` for it?

pradh: Per the doc do we still use `->?` for existence?

Spaceenter (approved):

internal/server/handler_v2.go
Spaceenter: Yea maybe "*" is more suitable...

Spaceenter (approved):

proto/v2/observation.proto
```diff
@@ -56,17 +65,17 @@ message ObservationRequest {
...
-  // When specified, only observation with this unit is returned.
-  string unit = 6;
-  // When specified, only observation with this measurement method is returned.
-  string measurement_method = 7;
-  // When specified, only observation with this observation period is returned.
-  string period = 8;
+  // Value of the observation
+  string value = 6;
+  // [Optional] filter for the facet
+  FacetFilter filter = 7;
```
pradh, line 71: If we are OK having nested protos for request, I slightly wonder if we want to do a common message for variable and entity:

message ExpIdArg {
  repeated dcids = 1;
  string expression = 2;
}
ExpIdArg entity = 1;
ExpIdArg variable = 2;

Then we can use singular and make it consistent with filter (to Wei's point)

proto/v2/observation.proto
```diff
@@ -56,17 +65,17 @@ message ObservationRequest {
...
+  // Fields to query for, valide values are: "variables", "entities", "date", "value"
```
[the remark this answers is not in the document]
pradh: +1 for singular

nit: Fields to return?

pradh (commented):
Thanks!

proto/v2/observation.proto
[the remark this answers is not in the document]
shifucun: Done

proto/v2/observation.proto
shifucun: Both suggestions are very good. Updated

PR #1087, opened 2023-04-07
implement V2 event

Spaceenter:


internal/server/v2/event/golden/event_collection_date/EarthquakeEvent_CA.json
```diff
@@ -0,0 +1,1197 @@
+{
```
shifucun, line 1: Can this just return a list?

internal/server/v2/event/golden/event_collection/FireEvent_CA_202210.json
```diff
@@ -0,0 +1,295 @@
+{
+  "eventCollection": {
```
shifucun, line 2: I feel the response can be simplified. For example eventCollection seems redundant?

propVals could be an array of 
{
 prop: xxx
 vals: []
}

Spaceenter: The motivation to have a wrapper like this that we can have a "bulk" method in the future.

Currently in V1, EventCollection is not a "bulk" method, and we were planning to add a "bulk" method, that's why we have this wrapper, so that in the future, "eventCollection" can be a repeated field.

internal/server/v2/event/golden/event_collection_date/EarthquakeEvent_CA.json
Spaceenter: Similar to "eventCollection", the motivation to have a wrapper like this that we can have a "bulk" method in the future.

Currently in V1, EventCollectionDate is not a "bulk" method, and we were planning to add a "bulk" method, that's why we have this wrapper, so that in the future, "eventCollectionDate" can be a repeated field.

shifucun (approved):

PR #1088, opened 2023-04-07
Recon

Spaceenter:


proto/v2/resolve.proto
```diff
@@ -0,0 +1,45 @@
...
+
+// Three modes are supported:
+//
+// 1. ID to ID:
+//   nodes = ["wikidataId/CA", "wikidataId/MA"]
+//   property = "<-wikidataId->nutsCode"
+//
+// 2. Coordinate (latitude#longitude) to DCID:
+//   nodes = ["34.56#-23.12", "-22.11#44.55"] 
+//   property = "<-geoCoordinate"
```
shifucun, line 29: Here have an implicit dcid like "<-geoCoordinate->dcid". I wonder should make that explicit, so it matches 1) where other ids are used?

Then the response is also clear that resolved_ids are the ids asked in the request.

Spaceenter: SG - done.

shifucun (approved):

PR #1090, opened 2023-04-09
[V2 API] Get svg parent chains via ->specializationOf+

shifucun:


internal/server/handler_v2.go
```diff
@@ -120,12 +114,18 @@ func (s *Server) V2Node(
 					in.NextToken,
 				)
 			}
-
-			if arc.Wildcard == "+" && !arc.Out {
+			if arc.Wildcard == "+" {
 				// Examples:
 				//   <-containedInPlace+{typeOf:City}
```
Spaceenter, line 119: nit: Add another example here?

->specializationOf+{typeOf:StatVarGroup}

Spaceenter (approved):

PR #1092, opened 2023-04-10
Implement derived series for V2 API.

Spaceenter:


proto/v2/observation.proto
```diff
@@ -57,9 +57,24 @@ message FacetFilter {
...
+  // 2. op: ObservationPeriod
+  // 3. ut: Unit
+  // 4. sf: ScalingFactor
+  //
+  // For example:
+  // 1. Person_Count - Person_Count_Female
+  // 2. Person_Count_Female / Person_Count[mm=CensusACS5yrSurvey;op=P5Y]
+  // 3. Person_Count - Person_Count_Female - Person_Count_Male[op=P5Y]
+  // 4. (Person_Count_Male - Person_Count_Female) / Person_Count
   DcidOrExpression variable = 1;
```
shifucun, line 76: Oh, can we have 3 fields in DcidOrExpression: dcids, expression, formula? 

Expression is now defined as "graph expression" (like in entity), so we should have a separate one for algebra expression (formula)? Also for variable, we can use graph expression to get a list of variables (like the child sv of an svg).

Spaceenter: SG - done

proto/v2/observation.proto
```diff
@@ -57,9 +58,24 @@ message FacetFilter {
 
 // Generic observation request
 message ObservationRequest {
-  // A list of entity DCIDs
+  // A list of statistical variable DCIDs, or a formula expression.
+  //
+  // Supported operators for the formula expression: +, -, *, /, along with parentheses.
```
shifucun, line 63: nit: maybe move the examples to line 28 where formula is defined?

shifucun (approved):
thanks for the update!

proto/v2/observation.proto
Spaceenter: done

PR #1094, opened 2023-04-11
[V2 API] Merger support for observation API

shifucun:


internal/server/handler_v2.go
```diff
@@ -16,22 +16,64 @@
...
+	"google.golang.org/protobuf/reflect/protoreflect"
 
 	pbv2 "github.com/datacommonsorg/mixer/internal/proto/v2"
 )
 
+func fetchRemote(
+	metadata *resource.Metadata,
+	apiPath string,
+	in protoreflect.ProtoMessage,
+	out protoreflect.ProtoMessage,
```
Spaceenter, line 45: Use: proto.Message

https://godoc.corp.google.com/pkg/google3/third_party/golang/protobuf/v2/reflect/protoreflect/protoreflect#ProtoMessage

"This is declared in the protoreflect package to avoid a cyclic dependency; use the proto.Message type instead, which aliases this type."

internal/server/handler_v2.go
```diff
@@ -16,22 +16,64 @@
...
+) error {
+	url := metadata.RemoteMixerUrl + apiPath
+	jsonValue, _ := protojson.Marshal(in)
+	request, _ := http.NewRequest("POST", url, bytes.NewBuffer(jsonValue))
+	request.Header.Set("Content-Type", "application/json")
+	request.Header.Set("X-API-Key", metadata.RemoteMixerApiKey)
+
+	client := &http.Client{}
```
Spaceenter, line 53: It feels like 'client' should be a field in 'server', and it should be passed in to 'fetchRemote' as an argument, otherwise, we need to init http.Client every time calling fetchRemote(), which seems an anti-pattern.

internal/server/handler_v2.go
```diff
@@ -16,22 +16,64 @@
...
+	response, err := client.Do(request)
+	if err != nil {
+		return err
+	}
+	defer response.Body.Close()
+
+	// Read response body
+	var responseBodyBytes []byte
+	if response.StatusCode == http.StatusOK {
```
Spaceenter, line 62: Shall we return some error when the status code is not http.StatusOK?

internal/server/resource/resource.go
```diff
@@ -50,12 +50,14 @@ type Cache struct {
...
 // Metadata represents the metadata used by the server.
 type Metadata struct {
-	Mappings        []*types.Mapping
-	OutArcInfo      map[string]map[string][]types.OutArcInfo
-	InArcInfo       map[string][]types.InArcInfo
-	SubTypeMap      map[string]string
-	MixerProject    string
-	BigQueryDataset string
+	Mappings          []*types.Mapping
+	OutArcInfo        map[string]map[string][]types.OutArcInfo
```
Spaceenter, line 54: use pointer:

map[string]map[string][]*types.OutArcInfo

internal/server/resource/resource.go
```diff
@@ -50,12 +50,14 @@ type Cache struct {
...
+	InArcInfo         map[string][]types.InArcInfo
```
Spaceenter, line 55: use pointer:

map[string][]*types.InArcInfo

Spaceenter (changes requested):

cmd/main.go
```diff
@@ -68,6 +68,8 @@ var (
 	tmcfCsvBucket  = flag.String("tmcf_csv_bucket", "", "The GCS bucket that contains tmcf and csv files")
 	tmcfCsvFolder  = flag.String("tmcf_csv_folder", "", "GCS folder for an import. An import must have a unique prefix within a bucket.")
 	memdbPath      = flag.String("memdb_path", "", "File path of memdb config")
+	// Remote mixer url. The API serves merged data from local and remote mixer
+	remoteMixerUrl = flag.String("remote_mixer_url", "", "Remote mixer url to fetch and merge data for API response.")
```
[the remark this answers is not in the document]
shifucun: For the planned future, this will only be for custom DC to talking to base DC API. I will make it a single remote for now to save the complexity downstream (concurrent fetch for multiple remotes, handle multiple APIs in config etc)

internal/server/handler_v2.go
shifucun: Nice!

internal/server/handler_v2.go
```diff
@@ -16,22 +16,64 @@
...
 
+func fetchRemote(
+	metadata *resource.Metadata,
+	apiPath string,
+	in protoreflect.ProtoMessage,
+	out protoreflect.ProtoMessage,
+) error {
+	url := metadata.RemoteMixerUrl + apiPath
+	jsonValue, _ := protojson.Marshal(in)
+	request, _ := http.NewRequest("POST", url, bytes.NewBuffer(jsonValue))
```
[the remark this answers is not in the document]
shifucun: Done

internal/server/handler_v2.go
shifucun: Done

internal/server/handler_v2.go
shifucun: Done

internal/server/handler_v2.go
```diff
@@ -16,22 +16,64 @@
...
+			return err
+		}
+	}
+
+	// Convert response body to string
+	err = protojson.Unmarshal(responseBodyBytes, out)
+	if err != nil {
+		return err
+	}
+	return nil
```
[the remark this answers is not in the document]
shifucun: Done

internal/server/resource/resource.go
shifucun: Nice, this is a long time issue.

internal/server/resource/resource.go
shifucun: Done

shifucun (commented):
thanks!

Spaceenter (approved):

PR #1096, opened 2023-04-12
fix a bug in obs merge, and add unit tests

Spaceenter:


internal/merger/merger.go
```diff
@@ -34,14 +34,14 @@ func MergeObservation(
...
+				o1.ByVariable[v].ByEntity[e].OrderedFacets,
+				eData.OrderedFacets...,
 			)
 		}
 	}
 	for facetID, facet := range o2.Facets {
 		o1.Facets[facetID] = facet
 	}
-	return o1, nil
+	return o1
```
shifucun, line 46: since this always return nil, no need to return error?

shifucun (approved):

internal/merger/merger.go
Spaceenter: Right, that's what this line's change is for.

PR #1099, opened 2023-04-13
Add merging for resolve results for V2 API.

Spaceenter:
Also renamed handler_bigtable.go to handler_core.go, because for Resolve(), it's not just read from Bigtable, but also Maps API, and (in the future) NL APIs.


shifucun (approved):
looking good!

PR #1100, opened 2023-04-13
make RecognizePlaces bulk

Spaceenter:


proto/recon.proto
```diff
@@ -126,18 +126,24 @@ message BulkFindEntitiesResponse {
...
   }
-  string query = 1;
-  repeated Item items = 2;
+
+  message Result {
+    string query = 1;
+    repeated Item items = 2;
+  }
+
+  repeated Result results = 1;
```
pradh, line 146: Wonder if we make it a map from query -> Items?

pradh (approved):
Thanks!

proto/recon.proto
Spaceenter: done

PR #1103, opened 2023-04-13
Add merging for events for V2 API

Spaceenter:


shifucun (approved):
nice, got a very good framework now

PR #1105, opened 2023-04-14
Refactor V2Node to only call v2pv.API (which needs token) once.

Spaceenter:


shifucun (approved):
I don't quite get where the old code not working for token?

Spaceenter:
The old code still works for token, this change is just trying to simplify the code a bit by avoid calling the same function twice.

PR #1108, opened 2023-04-18
Use RecognizePlaces in ResolveDescription.

Spaceenter:


internal/server/recon/find_entities.go
```diff
@@ -158,7 +147,159 @@ func BulkFindEntities(
...
+	error,
+) {
+	// First try to resolve DCIDs by RecognizePlaces.
+	entityInfoToDCIDSet, dcidSet, err := resolveWithRecognizePlaces(
+		ctx, store, entityInfoSet)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// See if there are any entities that cannot be resolved by RecognizePlaces.
```
pradh, line 167: 1. I think it might be useful to have set some debug info so we can tell how we matched?  This will be useful in evaluating/debugging the results?  Separate PR is fine.

2. Also, I'm wondering about doing calls parallel-y, and if both have results, prefer recognize-places, but use Maps API as signal to reorder the results.  Could add a TODO maybe?

internal/store/files/recog_place_map.go
```diff
@@ -52,23 +64,28 @@ func RecogPlaceMap() (map[string]*pb.RecogPlaces, error) {
...
-		if recogPlace.Dcid == "" {
+		dcid := strings.TrimSpace(record[0])
+		if dcid == "" {
 			return nil, status.Errorf(codes.FailedPrecondition,
 				"Empty DCID for CSV record: %v", record)
 		}
+		recogPlace := &pb.RecogPlace{Dcid: dcid}
+
+		// Containing places.
+		containingPlaces := strings.Split(strings.ReplaceAll(record[3], " ", ""), ",")
```
pradh, line 76: For the row: `<dcid>,Chennai, District, "Tamil Nadu, India",6000000`, will "Tamil Nadu" be addd as "TamilNadu"?  If so, will that match the query string?

internal/store/files/recog_place_map.go
```diff
@@ -99,5 +116,28 @@ func RecogPlaceMap() (map[string]*pb.RecogPlaces, error) {
 		}
 	}
 
-	return recogPlaceMap, nil
+	// Add more names in the pattern of "selfName containingPlaceName", e.g., "Brussels Belgium".
```
pradh, line 119: I slightly wonder about this.  IIRC, the existing algorithm already should infer that <Brussels> is a place, and then <Belgium> is a place, and further <Brussels> is in <Belgium> so we can decide that we should consume both tokens to match Brussels?

internal/store/files/recog_place_map.go
Spaceenter: Note that this line of code deals with "containingPlaces", which is a list of DCIDs (e.g. "country/IND,wikidataId/XXX,", etc." - Your comment is about names (which contains spaces).

internal/store/files/recog_place_map.go
Spaceenter: Right, I thought about the capability of existing algorithms. However, if using the capability of existing algorithms, then:
1. We need to add the expanded name to RecogizePlacesResponse, which is not needed for RecognizePlaces users.
2. The existing algorithm is already complicated, plugging in more logic there will make it a bit more complex. In comparison, memorizing dcidToNames in store is not computationally-heavy, but could leave RecognizePlaces code intact.

internal/store/files/recog_place_map.go
pradh: Ah, got it.

internal/store/files/recog_place_map.go
pradh: Ah, I missed that DcidToNames is only used in FindEntities flow.  This is good!

pradh (approved):
Thanks!

internal/server/recon/find_entities.go
Spaceenter: Added TODOs

PR #1114, opened 2023-04-21
Call local and remote in parallel for V2 resolve

Spaceenter:


internal/server/handler_v2.go
```diff
@@ -74,19 +75,40 @@ func fetchRemote(
...
-		return nil, err
-	}
-	if s.metadata.RemoteMixerDomain != "" {
-		remoteResp := &pbv2.ResolveResponse{}
-		err := fetchRemote(s.metadata, s.httpClient, "/v2/resolve", in, remoteResp)
+	errGroup, errCtx := errgroup.WithContext(ctx)
+	respChan := make(chan *pbv2.ResolveResponse, 2)
+
+	errGroup.Go(func() error {
+		resp, err := V2ResolveCore(errCtx, s.store, s.mapsClient, in)
```
shifucun, line 82: just confirm V2ResolveCore can return nil (instead of empty protobuf)

shifucun (approved):

internal/server/handler_v2.go
Spaceenter: Yes, the returned type is *pbv2.ResolveResponse, which is a pointer, so "nil" is just an empty pointer.

Spaceenter: Also, I think "nil" is in general preferred over a non-nil pointer to an empty protobuf.

PR #1115, opened 2023-04-21
Read local and remote in parallel for V2 event and observation

Spaceenter:


shifucun (approved):
framework programming!

PR #1118, opened 2023-04-21
Filter out inferior facets and remove empty entry from observation response

shifucun:


internal/server/v2/observation/collection.go
```diff
@@ -131,6 +136,11 @@ func FetchFromCollection(
 			return nil, err
 		}
 		for variable, variableData := range moreResult.ByVariable {
+			for entity, entityData := range variableData.ByEntity {
+				if len(entityData.OrderedFacets) == 0 {
+					delete(variableData.ByEntity, entity)
```
Spaceenter, line 141: Curious - instead of deleting here, is it possible to avoid adding these empty things in FetchFromSeries() above?

Spaceenter (approved):

internal/server/v2/observation/collection.go
shifucun: Good question. FetchFromSeries takes a list of entities. When it's used in an API call where a list of entities are provided, we do want to return these entities even they have empty data. But in the case of containedIn, the list of child entities are implicit (and even changing), hence not making sense to return if result is empty.

shifucun:
Thanks!

PR #1125, opened 2023-04-25
Add triples to V2 API.

Spaceenter:


shifucun (approved):
So the local and remote merge are still to be added?

PR #1127, opened 2023-04-26
Update mixer service group and resource management

shifucun:
* Move some of the most used APIs (by traffic) into the dedicated service groups (observation, node)
* Decrease the memory usage for all service group as they are over allocated now
* Decrease pods for "svg" service group as it serves APIs with data in memory with very few computations
* Decrease pods for "default" as it only serves low traffic (and fewer) APIs.


chejennifer (approved):

shifucun:
thanks for review!

PR #1128, opened 2023-04-27
Support multiple typeOf filters for ResolveDescription

Spaceenter:


Spaceenter:
Also updated go module, because new Go package "golang.org/x/exp" needs to be added for using maps.Keys().

internal/server/v2/parser.go
```diff
@@ -100,24 +104,35 @@ func parseArc(arrow, expr string) (*Arc, error) {
...
-					codes.InvalidArgument, "invalid filter string: %s", rawExpr)
+			if strings.Contains(part, ":") {
+				kv := strings.Split(part, ":")
+				if len(kv) != 2 || kv[0] == "" || kv[1] == "" {
+					return nil, status.Errorf(
+						codes.InvalidArgument, "invalid filter string: %s", rawExpr)
+				}
+				lastKey = kv[0]
+				filter[lastKey] = map[string]struct{}{kv[1]: {}}
+			} else { // No ":" means this is another val in square bracket.
```
shifucun, line 129: this is clear!

internal/server/v2/graph.go
```diff
@@ -25,8 +25,8 @@ type Arc struct {
 	Decorator string
 	// The properties of the arc. This is when property is specified with []
 	BracketProps []string
-	// The filter of the arc.
-	Filter map[string]string
+	// The filter of the arc: filter key -> filter value set.
+	Filter map[string]map[string]struct{}
```
shifucun, line 29: wonder why use set instead of just array? Then all the uses of this filter looks simpler, and only need to dedupe it once in the building (and in most of the cases, it only has one element).

Spaceenter: done

shifucun (approved):

PR #1129, opened 2023-04-27
Be able to call remote for place page

Spaceenter:


Spaceenter:
Also move fetchRemote() to handler_core.go, as it's used by hander_v1.go too.

internal/server/handler_v1.go
```diff
@@ -220,7 +220,20 @@ func (s *Server) BioPage(
 func (s *Server) PlacePage(
 	ctx context.Context, in *pbv1.PlacePageRequest,
 ) (*pbv1.PlacePageResponse, error) {
-	return page.PlacePage(ctx, in, s.store)
+	localResp, err := page.PlacePage(ctx, in, s.store)
+	if err != nil {
+		return nil, err
+	}
+	if localResp == nil && s.metadata.RemoteMixerDomain != "" {
```
shifucun, line 227: maybe check localResp.StatVarSeries here instead of setting localResp to nil?

shifucun (approved):

internal/server/handler_v1.go
Spaceenter: done

PR #1131, opened 2023-05-01
[stat var hierarchy] separate blocklist svgs for each environment

chejennifer:
- Get blocklisted svgs from environment file
- set blocklisted svgs file in each helm chart

works together with https://github.com/datacommonsorg/website/pull/2627


shifucun (approved):
looking good. And you can deploy to mixer_dev for testing

PR #1132, opened 2023-05-01
Return name for containedInPlace+ node API

shifucun:
This is to match the general property values response (also matching V1)

### DIFF
--- a/internal/server/v2/propertyvalues/api.go
+++ b/internal/server/v2/propertyvalues/api.go
@@ -157,13 +157,37 @@ func LinkedPropertyValues(
 		if err != nil {
 			return nil, err
 		}
+		// Fetch descendent names
+		descendents := []string{}
+		for _, vals := range data {
+			descendents = append(descendents, vals...)
+		}
+		nameResp, _, err := v1pv.Fetch(
+			ctx,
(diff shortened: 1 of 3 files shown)


internal/server/v2/propertyvalues/api.go
```diff
@@ -157,13 +157,37 @@ func LinkedPropertyValues(
...
 			dcids, ok := data[node]
 			if ok && len(dcids) > 0 {
 				for _, dcid := range dcids {
-					list = append(list, &pb.EntityInfo{Dcid: dcid})
+					info := &pb.EntityInfo{Dcid: dcid}
+					var name string
+					if tmp, ok := nameResp[dcid]["name"]; ok {
+						name = tmp[""][0].Value
+					}
+					info.Name = name
```
Spaceenter, line 189: Two lines shorter:

					if v, ok := nameResp[dcid]["name"]; ok {
						info.Name = v[""][0].Value
					}

Spaceenter (approved):

internal/server/v2/propertyvalues/api.go
shifucun: Done

PR #1142, opened 2023-05-10
Temp fix for missing observation collectioin data when multiple child types are present

shifucun:


internal/server/v2/observation/contained_in.go
```diff
@@ -88,13 +87,36 @@ func FetchContainedIn(
...
+				if refetch {
+					btData, err = stat.ReadStatCollection(
+						ctx, store.BtGroup, bigtable.BtObsCollection,
+						ancestor, "State", variables, queryDate,
+					)
+					if err != nil {
+						return nil, err
+					}
+				}
+			}
```
pradh, line 119: nit: Wonder if we can do like:
{
  "AA1": "State",
  "State": "AA1"
}

If childType in the map, use the value?

pradh (approved):
Thanks!

internal/server/v2/observation/contained_in.go
shifucun: As this is only for temp fix before Flume fix, will not make this a general solution for now. Also looks like we only need to check when it's AA1, since State is preferred in Flume from table_types.textproto?

Spaceenter (approved):

PR #1153, opened 2023-06-01
Add the ability to use only certain big tables

shifucun:
SV and SVG in "infrequent" and other import groups could be stale. To prevent them from showing up in the hierarchy and to avoid building the new BT for them, this introduces a way to filter big tables during cache read.

### DIFF
--- a/internal/server/v1/info/golden/variable_group_info/weather.json
+++ b/internal/server/v1/info/golden/variable_group_info/weather.json
@@ -1,0 +1,509 @@
+{
+  "node": "dc/g/Weather",
+  "info": {
+    "absoluteName": "Weather",
+    "childStatVars": [
+      {
+        "id": "ConsecutiveDryDays",
+        "searchNames": [
+          "Consecutive Dry Days of Place"
+        ],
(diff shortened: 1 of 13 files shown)


internal/server/v1/info/golden/variable_group_info/weather.json
```diff
@@ -0,0 +1,509 @@
...
+      {
+        "id": "ConsecutiveDryDays",
+        "searchNames": [
+          "Consecutive Dry Days of Place"
+        ],
+        "displayName": "Consecutive Dry Days",
+        "definition": "mp=consecutiveDryDays,pt=Place",
+        "hasData": true
+      },
+      {
```
pradh, line 15: This is because of prod tables I guess?

pradh (approved):
Very nice!  Thanks!

internal/server/v1/info/golden/variable_group_info/weather.json
shifucun: Yes, after using newest frequent, this file is much smaller.

PR #1154, opened 2023-06-01
Fix a bug of inconsistent bigtable group and accessor

shifucun:


pradh (approved):
Seems subtle.  Where does the inconsistency come from?

internal/store/bigtable/reader.go
```diff
@@ -109,12 +111,11 @@ func ReadWithFilter(
 // needed by the pagination APIs.
 func ReadWithGroupRowList(
 	ctx context.Context,
-	btGroup *Group,
+	tables []*cbt.Table,
 	prefix string,
 	accs []*Accessor,
 	unmarshalFunc func([]byte) (interface{}, error),
 ) ([][]BtRow, error) {
-	tables := btGroup.Tables(nil)
```
shifucun, line 117: note here use "nil" filter, but it should use the same filter as caller. 

But better to pass in the filtered tables directly here.

shifucun:
thanks!

PR #1159, opened 2023-06-07
Merge place metadata cache for place info API

shifucun:


internal/server/place/places.go
```diff
@@ -207,53 +207,66 @@ func GetPlaceMetadataHelper(
...
+				if currInfo, ok := metaMap[info.Dcid]; !ok {
+					metaMap[info.Dcid] = info
+				} else {
+					// Merge place info from new import group.
+					if currInfo.Name == "" {
+						currInfo.Name = info.Name
+					}
+					if currInfo.Type == "" {
+						currInfo.Type = info.Type
+					}
```
Spaceenter, line 227: Something to think about...:
Wonder if it's possible to have   currInfo.Type != info.Type? In that case, shall we raise an error? Or prefer the main DC one?
Similar comment for curInfo.Name above.

Spaceenter (approved):

internal/server/place/places.go
shifucun: For all the merging, we rank custom DC higher, and the loop here is from most ranked import group. So now if higher ranked import has empty value, then use the lower ranked one. If the values conflict, then still prefer the higher ranked import group..

PR #1163, opened 2023-06-13
For event APIs, merge results across import groups

Spaceenter:


Spaceenter:
Note that the diffs in the golden files are due to:
1. Dedup (which wasn't done before).
2. Sort (which wasn't done before).

ajaits (approved):

internal/server/v1/event/event.go
```diff
@@ -177,16 +173,38 @@ func Collection(
...
-				resp.EventCollection.Events = append(
-					resp.EventCollection.Events,
-					event,
-				)
-				provID := event.ProvenanceId
-				resp.EventCollection.ProvenanceInfo[provID] = data.ProvenanceInfo[provID]
+
+				eventID := event.GetDcid()
+				if _, ok := idToEvent[eventID]; !ok {
+					idToEvent[eventID] = event
```
shifucun, line 179: The event here is replaced by later import group? Since we rank the import groups already, so should keep the higher ranked ones, meaning continue here if already exists?

Spaceenter: NOTE that line 178 is "!ok", which means if not exist, we add, otherwise we don't add. So I think the code already achieves the logic you stated?

@shifucun

shifucun: I see

shifucun (approved):

PR #1164, opened 2023-06-14
Add new import groups of country and auto*

Spaceenter:


internal/store/bigtable/group.go
```diff
@@ -28,10 +28,14 @@ import (
...
 	"dcbranch":   0, // Used for the latest proto branch cache
 	"branch":     0, // Used for legacy branch cache
-	"frequent":   1,
-	"ipcc":       2,
-	"biomedical": 3,
-	"disaster":   4,
+	"auto1d":     1,
+	"auto1w":     2,
+	"auto2w":     3,
+	"frequent":   4,
```
pradh, line 34: What is the new plan with frequent given autoX groups?  If it continues to be daily, should it be right after auto1d?

pradh (approved):
Thanks!

internal/store/bigtable/group.go
Spaceenter: done

shifucun (approved):
thanks for updating the golden test!!

PR #1170, opened 2023-07-06
Bundle both mixer and esp in one container

shifucun:
This tests the feasibility to bring multiple process in the same container to prepare for doing this for mixer + website.

This PR also allows reading the api secret from environment directly, so a custom mixer can be brought up locally without GCP at all.


internal/util/util.go
```diff
@@ -503,23 +505,44 @@ func StringListIntersection(list [][]string) []string {
...
 }
 
-// MapsClient gets the client for Maps API.
-func MapsClient(ctx context.Context, projectID string) (*maps.Client, error) {
-	secretClient, err := secretmanager.NewClient(ctx)
+func ReadLatestSecret(
+	ctx context.Context, projectID, secretID string,
+) (string, error) {
+
+	secret := os.Getenv(strings.Replace(secretID, "-", "_", -1))
```
pradh, line 512: nit: add a note why Replace?

scripts/push_binary.sh
```diff
@@ -30,5 +30,5 @@ fi
 gcloud builds submit \
     --project=datcom-ci \
     --config=build/ci/cloudbuild.push.yaml \
-    --substitutions SHORT_SHA="$TAG" \
+    --substitutions SHORT_SHA="$TAG",_SKIP_AUTOPUSH_UPDATE=true \
```
pradh, line 33: nit: should an external var not start with `_`?

pradh (approved):
Nice!

scripts/push_binary.sh
shifucun: cloudbuild requires custom variable to start with _: https://cloud.google.com/build/docs/configuring-builds/substitute-variable-values#using_user-defined_substitutions

internal/util/util.go
shifucun: Done

shifucun (commented):
thanks!

PR #1172, opened 2023-07-11
Support SQLite as data store; Implement V2 API for fetching direct observations

shifucun:


sqlite/import.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
pradh, line 1: Do you mean to checkin the datacommons.db?

cmd/main.go
```diff
@@ -59,8 +63,13 @@ var (
 	customBigtableInfo = flag.String("custom_bigtable_info", "", "Yaml formatted text containing information for custom Bigtable")
 	// Branch Bigtable Cache
 	useBranchBigtable = flag.Bool("use_branch_bigtable", true, "Use branch bigtable cache")
+	// SQLite database
+	useSQLite = flag.Bool("use_sqlite", false, "Use SQLite as database.")
+	sqliteDB  = flag.String("sqlite_db", "./sqlite/datacommons.db", "SQLite DB file path.")
 	// Stat-var search cache
 	useSearch = flag.Bool("use_search", true, "Uses stat var search. Will build search indexes.")
+	// Include maps client
+	useMapsApi = flag.Bool("use_maps_api", true, "Uses maps API for place recognition.")
```
pradh, line 72: Is this for the recon API implementation?

internal/server/v2/observation/direct.go
```diff
@@ -48,70 +49,123 @@ func FetchDirect(
...
-							FacetId:      facetID,
-							Observations: obsList,
-						},
-					)
 				}
+				result.ByVariable[variable].ByEntity[entity] = entityObservation
 			}
-			result.ByVariable[variable].ByEntity[entity] = entityObservation
+		}
+	} else if store.SQLiteClient != nil {
```
pradh, line 125: nice.  Maybe could have a TODO to modularize the different backends a bit?

pradh (approved):
Very cool!

cmd/main.go
shifucun: Yes.

internal/server/v2/observation/direct.go
shifucun: Yea, that sounds good. I will do that after adding more APIs / need a new connector for Cloud SQL

sqlite/import.sh
shifucun: Yea, I want to expand that and use it for testing; Also users would use it as "hello world" db.

shifucun (commented):
thanks!

PR #1173, opened 2023-07-12
Fix place info merge bug and btGroup nil access bug

shifucun:
As place info is always set (but only with dcid), it would not fetch remote place info, hence giving incomplete data for custom DC.

The btGroup nil access is result from previous PR.


pradh (approved):
What is an example symptom / issue this causes?

shifucun:
> What is an example symptom / issue this causes?

Added in PR description.

PR #1180, opened 2023-07-14
[SQLite] Implement simple property values query

shifucun:


internal/server/v1/propertyvalues/core.go
```diff
@@ -32,8 +36,173 @@ import (
...
+	var resp map[string]map[string]map[string][]*pb.EntityInfo
+	var pg *pbv1.PaginationInfo
+	var err error
+	if store.BtGroup != nil {
+		resp, pg, err = fetchBT(ctx, store.BtGroup, nodes, properties, limit, token, direction)
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+	// No pagination for sqlite query, so if there are pagination token, meaning
```
pradh, line 63: nit: s/are/is/

pradh (approved):
Quick review

internal/server/v1/propertyvalues/core.go
shifucun: done

shifucun (commented):
thanks for the review!

PR #1183, opened 2023-07-15
Update RecognizePlaces place info file to include Alpha2/3 codes as c…

Spaceenter:
…ountry names


Spaceenter:
SQL updates: cl/548350675

Spaceenter:
It might not be a good solution, as some common words are now recognized as countries, like "in" for "India", "are" for "United Arab Emirates", "and" for "Andorra"

PR #1184, opened 2023-07-16
Add support for abbreviated names for place recognition.

Spaceenter:


internal/server/recon/golden/find_entities_test.go
```diff
@@ -45,6 +45,12 @@ func TestFindEntities(t *testing.T) {
 				},
 				"santa_clara_county.json",
 			},
+			{
```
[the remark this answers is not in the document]
jehangiramjad: Also, perhaps we should note that ME also refers to the US State of Maine. So some comment about why we match to country/MNE over the state in the US is important.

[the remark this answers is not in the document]
jehangiramjad: Hmm.. Upon further investigation, country/MNE has a population < 700k while state of Maine has a population > 1.3mn. So why is country/MNE preferred?

jehangiramjad (approved):
Thanks! Just a couple of clarifying questions where the decisions could be communicated in the comments somewhere (e.g. we should have a test or pointer to a test where the word "me" in lower case also appears in the query which does not match to "Maine" or "Montenegro". 

Approved from my side but please wait for @pradh's more careful review.

internal/server/v2/resolve/golden/resolve_description/abbreviated_names.json
```diff
@@ -0,0 +1,37 @@
...
+        "geoId/2111872",
+        "geoId/2412150",
+        "geoId/2910468",
+        "geoId/4210768"
+      ]
+    },
+    {
+      "node": "ME",
+      "resolvedIds": [
+        "country/MNE",
```
pradh, line 16: To confirm, why is the order here country before state, while in recognize places, state before country?

If we had to pick, I might slightly go with the state before country order...

internal/store/files/README.md
```diff
@@ -0,0 +1,21 @@
+## BQ SQL query for WorldGeosForPlaceRecognitionAbbreviatedNames.csv
```
pradh, line 1: We could also expand to `isoCode` for AA1's which has codes like "US-CA" from which we parse out the last two chars... so TN (for Tamil Nadu state), SH (for shanghai).  But we might prefer US states if we do that..

internal/server/recon/recognize.go
```diff
@@ -123,6 +127,12 @@ func (p *placeRecognition) findPlaceCandidates(
 		return 0, nil
 	}
 
+	// Check if the first token match any abbreviated name.
+	// Note: abbreviated names are case-sensitive.
+	if places, ok := p.abbreviatedNameToPlaces[tokens[0]]; ok {
+		return 1, places
+	}
```
pradh, line 134: To confirm, Will this work for "Sunnyvale CA" or "Texas US" ?

pradh (commented):
Thanks Wei!  A couple of comments.

internal/store/files/README.md
pradh: We can have TODO for now, and expand later, per Guha suggestion.

Spaceenter: Added a TODO in the README.md which contains the BQ SQL query.

internal/server/recon/recognize.go
Spaceenter: Yes it works - added a test case "Sunnyvale CA" in //internal/server/v2/resolve/golden/resolve_description_test.go

internal/server/recon/recognize.go
```diff
@@ -107,7 +108,10 @@ func tokenize(query string) []string {
 }
 
 type placeRecognition struct {
+	// The key is the first token/word (lower case) of each place.
 	recogPlaceMap map[string]*pb.RecogPlaces
+	// The key is abbreviated name of each place.
```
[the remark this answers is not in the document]
Spaceenter: done

internal/server/v2/resolve/golden/resolve_description/abbreviated_names.json
Spaceenter: I think RecognizePlaces sort places by population. 
However, ResolveDescription (which calls BulkFindEntities) doesn't sort the resolved entities. If needed, we can also sort (maybe in a future PR) by population, but I'm not sure if that's necessary, because ResolveDescription can also take placeType as an input to filter, so if the user really wants "State" or "Country", he/she can specify it in the query, like "<-description{typeOf:State}->dcid"

internal/server/recon/golden/find_entities_test.go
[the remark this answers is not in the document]
Spaceenter: Added test which includes both "ME" and "me" in: //internal/server/recon/golden/recognize_places_test.go

Now the tests for RecognizePlaces are scattered in a few places, mainly due to V1->V2 API migrations. We'll try to clean it up, deprecate V1 things, then move all tests under V2.

internal/server/recon/golden/recognize_places/result.json
```diff
@@ -204,6 +204,38 @@
         }
       ]
     },
+    "the birds in ME and USA are chirpy": {
```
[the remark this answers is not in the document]
Spaceenter: See result below, we are indeed preferring state over country, as we sort place candidates by population.

internal/server/recon/golden/find_entities_test.go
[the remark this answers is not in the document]
Spaceenter: Re country/state orders, copy another comment below:

1. RecognizePlaces sort places by population.
2. However, ResolveDescription (which calls BulkFindEntities in find_entities.go) doesn't sort the resolved entities. If needed, we can also sort (maybe in a future PR) by population, but I'm not sure if that's necessary, because ResolveDescription can also take placeType as an input to filter, so if the user really wants "State" or "Country", he/she can specify it in the query, like "<-description{typeOf:State}->dcid"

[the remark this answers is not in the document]
pradh: I would vote for making ResolveDescription result match the same logic as RecognizePlaces, but that's outside the scope of this PR.

internal/server/v2/resolve/golden/resolve_description_test.go
```diff
@@ -97,6 +97,16 @@ func TestResolveDescription(t *testing.T) {
 				"<-description->dcid",
 				"harvard.json",
 			},
+			{
+				[]string{
+					"California, USA",
+					"Mountain View, US",
+					"Sunnyvale CA",
+					"ME",
```
pradh, line 105: I guess "CA US" should work too?

pradh (approved):
Thanks Wei!

internal/server/v2/resolve/golden/resolve_description_test.go
Spaceenter: Yes it resolves to "geoId/06" -- added to resolve_description_test.go.

internal/server/recon/golden/find_entities_test.go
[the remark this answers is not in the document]
Spaceenter: So basically keep the ranking by population - cool.

PR #1186, opened 2023-07-17
[SQLite] Support places in query

shifucun:


docs/developer_guide.md
```diff
@@ -115,9 +115,11 @@ protoc --proto_path=proto \
   proto/*.proto proto/**/*.proto
 ```
 
-Assuming mixer gRPC server is at `localhost:12345`, run the following from repo
-root to spin up envoy proxy. This exposes the http mixer service at
-`localhost:8081`.
+Start mixer gRPC server (running at `localhost:12345`) by following `go run
```
shifucun, line 118: @pradh make the instruction clear here.

internal/server/placein/placein.go
```diff
@@ -40,38 +42,77 @@ func GetPlacesIn(
...
 	}
+
+	if store.SQLiteClient != nil {
+		// Only queries based on direct containedInPlace for now.
+		// This could extend to more hops and even link with BT cache data, but that
+		// might make it too complicated.
+		// In custom DC, it's reasonable to ask user to provide direct containment
+		// relation.
+		parentPlacesStr := "'" + strings.Join(parentPlaces, "', '") + "'"
+		query := fmt.Sprintf(
```
Spaceenter, line 88: Do we care about SQL injection here?
Since this is open-source code, maybe we want to avoid SQL injection?
https://go.dev/doc/database/sql-injection
@pradh

Spaceenter (changes requested):

internal/server/placein/placein.go
shifucun: Totally. I updated here and a few other places to use query args in stead of raw strings.

Spaceenter (approved):

PR #1187, opened 2023-07-17
In ResolveDescription/FindEntities, keep the sort order by population

Spaceenter:


Spaceenter:
This matches the behavior in RecognizePlaces.

jehangiramjad (approved):

PR #1189, opened 2023-07-17
Avoid resolving Bogus names in RecognizePlaces, but do so in ResolveD…

Spaceenter:
…escription.


proto/recon.proto
```diff
@@ -126,7 +126,12 @@ message BulkFindEntitiesResponse {
   repeated Entity entities = 1;
 }
 
-message RecognizePlacesRequest { repeated string queries = 1; }
+message RecognizePlacesRequest {
+  repeated string queries = 1;
+  // When set, bogus name should be resolved.
+  // This needs to be set for ResoveDescription, but not set for RecognizePlaces for sentences.
+  bool resolve_bogus_name = 2;
```
pradh, line 133: nit: maybe call it `require_context_for_common_names`?  we internally call it bogus name.

internal/store/files/recog_place_store.go
```diff
@@ -205,3 +216,23 @@ func loadAbbreviatedNames() (map[string][]string, error) {
...
+	}
+
+	res := map[string]struct{}{}
+
+	for _, record := range records {
+		if len(record) != 1 {
+			return nil, status.Errorf(codes.FailedPrecondition,
+				"Wrong BogusPlaceNames CSV record: %v", record)
+		}
+		res[strings.ToLower(record[0])] = struct{}{}
```
pradh, line 234: nit: Trim extra spaces?

pradh (approved):
Thanks!

internal/store/files/recog_place_store.go
Spaceenter: done

proto/recon.proto
Spaceenter: Per offline discussion - making this an internal flag.

PR #1191, opened 2023-07-18
Better selection algo for large amount of place candidates for PlaceR…

Spaceenter:
…ecognition

### DIFF
--- a/internal/server/recon/recognize_test.go
+++ b/internal/server/recon/recognize_test.go
@@ -340,10 +340,10 @@ func TestRankAndTrimCandidates(t *testing.T) {
 					{
 						Tokens: []string{"Mountain", "View"},
 						Places: []*pb.RecogPlace{
-							{Dcid: "geoId/MTV2", Population: 102},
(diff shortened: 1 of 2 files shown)


internal/server/recon/recognize_test.go
```diff
@@ -380,15 +379,15 @@ func TestRankAndTrimCandidates(t *testing.T) {
...
+							{Dcid: "geoId/MTV2", Population: 9102},
+							{Dcid: "geoId/MTV1", Population: 9101},
 						},
 					},
 				},
 			},
 		},
 	} {
-		got := pr.rankAndTrimCandidates(c.tokenSpans)
+		got := pr.rankAndTrimCandidates(c.tokenSpans, 2)
```
pradh, line 390: should we have an e2e test for Franklin County?

pradh (approved):
Thanks!

internal/server/recon/recognize_test.go
Spaceenter: SG - could do in a follow up CL.

PR #1196, opened 2023-07-20
Fix a bug of combineContainedIn for RecognizePlaces.

Spaceenter:


internal/server/recon/golden/recognize_places/result.json
```diff
@@ -69,6 +69,38 @@
...
+      "items": [
+        {
+          "span": "I went to"
+        },
+        {
+          "span": "Half Moon Bay California",
+          "places": [
+            {
+              "dcid": "geoId/06"
+            },
```
pradh, line 82: Why does this match california?

internal/server/recon/golden/recognize_places/result.json
```diff
@@ -69,6 +69,38 @@
...
+        },
+        {
+          "span": "and"
+        },
+        {
+          "span": "California , Washington County",
+          "places": [
+            {
+              "dcid": "geoId/42125"
+            },
```
pradh, line 96: ditto, why does it match county?

If this is a combined span, it should be only geoId/4210768, right?

internal/server/recon/golden/recognize_places/result.json
Spaceenter: Changed

internal/server/recon/golden/recognize_places/result.json
Spaceenter: Done

pradh (approved):
Thanks Wei!

PR #1207, opened 2023-07-23
Directly write raw csv files into sqlite instead of creating intermediate files.

shifucun:


sqlite/writer/writer.go
```diff
@@ -261,49 +235,88 @@ func resolvePlaces(ctx *context,
 	return placeToDCID, nil
 }
 
+func prepareDatabase(fileDir string) error {
+	dbPath := path.Join(fileDir, "datacommons.db")
```
Spaceenter, line 239: https://stackoverflow.com/questions/39181790/whats-the-difference-between-path-path-filepath-packages-in-go

Better to use filepath.Join?

Spaceenter (approved):

sqlite/writer/writer.go
shifucun: good to know. Fixed

PR #1209, opened 2023-07-24
Add dominant place type to coordinate recon result

Spaceenter:
The change involves both V1 and V2 recon APIs.

For V2, all recon/resolve APIs share the same proto interface (especially the response message), therefore, we also need to update the response for V2 ID and Description resolve APIs.


pradh (approved):
Thanks!

pradh:
@keyurva - FYI, this should help import tool changes to know the type

PR #1220, opened 2023-08-02
Fix mixer bigtable read context cancel issue

shifucun:
This issue is uncovered from observation series BT cache read where O(100) variable and O(10) places are queried. In this case, the context used to read Bigtable is canceled after certain amount of read call. This won't be an issue when the variable and place number is small.

In the previous implementation, each Bigtable table has its own client to the instance, thus there are 10+ client from one binary connected to the instance. With this change, there is only one shared client for all the base bigtable (one for each import group). This apparently solves the issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,4 +61,4 @@ scripts/tmp*.sh
 http_memprof_out
 
 # custom dc data
-data/
+/data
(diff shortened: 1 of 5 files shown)


pradh (approved):
Wow!  Nice!

PR #1222, opened 2023-08-04
Allow filtering by multiple domains in the  V2 observations API

juliawu:
Convert the domain filter for the v2/observations API to take an array of strings, instead of just a single string.

This will be helpful for implementing filtering by multiple sources in the website's chart tiles and web components (e.g. multivariable charts, where each variable has a specified source).

This PR:
* Converts `FacetFilter.Domain` to `[]string` type, and is renamed to the pluralized `FacetFilter.Domains`
* Expand filtering logic to allow provenances from any of the given domains, instead of just one domain.
* Add tests and goldens for the multi-domain case.


shifucun (approved):
thanks for adding this!

PR #1223, opened 2023-08-05
Sort V2 coordinate recon result

Spaceenter:


keyurva (approved):

internal/sqlite/writer/writer.go
```diff
@@ -235,8 +235,8 @@ func resolvePlaces(
 		if _, ok := placeToDCID[entity.GetNode()]; ok {
 			continue
 		}
-		// TODO(ws): Handle the case with multiple candidates.
 		if len(entity.GetCandidates()) > 0 {
+			// The resolve API sorts candidates already, so we pick the first one.
```
shifucun, line 239: Very soon we want to read the type of place from user config. Does each item here have the corresponding type?

shifucun (approved):

internal/sqlite/writer/writer.go
Spaceenter: Yes, each item has a 'dominantType' field, so we can actually do filtering here if needed in the future.

internal/server/v2/resolve/resolve.go
```diff
@@ -176,3 +186,44 @@ func parseCoordinate(coordinateExpr string) (float64, float64, error) {
 
 	return lat, lng, nil
 }
+
+// Sort resolved place candidates by a priority list of place types.
+// If a candidate's type is not in the priority list, then sort by DCID alphabetically.
+func getSortedResolvedPlaceCandidates(
```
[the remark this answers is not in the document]
Spaceenter: Good idea - it seems like Go doesn't have an official package of treemap (or ordered map), and we're trying to reduce the dependencies on unofficial packages... 

For ordered map, a common pattern I see is to sort the map keys using 'sort' package, then iterate through the keys.

PR #1228, opened 2023-08-17
Update Golang to 1.20.7 to mitigate docker image security issue

shifucun:
The old Golang base image has security vulnerability that prevents the GCP project to give IAM access to external account (ex, stanford.edu)

Base mixer image is not used anymore, since the website-compose image (under website repo) will have Golang in it and build mixer binary as well.


keyurva (approved):
Thanks.

shifucun:
thanks for review!

PR #1230, opened 2023-08-18
Add filter observations by facet id

chejennifer:
- add a FacetFilter field to filter for facet id


proto/v2/observation.proto
```diff
@@ -73,6 +73,10 @@ message FacetFilter {
   // When specified, only observations with provenance in these domain are
   // returned.
   repeated string domains = 4;
+  // key is variable dcid and value is facet id to filter for. For all variables
+  // with a facet id to filter for, only observations with that facet id are
+  // returned.
+  map<string, string> facet_id = 5;
```
shifucun, line 79: instead of adding a map here, how about change line 98 to use repeated filter? Then in facet filter we add a repeated string of stat var dcid.

chejennifer: done

proto/v2/observation.proto
```diff
@@ -62,17 +62,13 @@ message FacetObservation {
...
-  string unit = 1;
-  // When specified, only observations with this measurement method are
-  // returned.
-  string measurement_method = 2;
-  // When specified, only observations with this observation period are
-  // returned.
-  string period = 3;
+  // List of variables that this facet filter applies to
+  repeated string variables = 1;
   // When specified, only observations with provenance in these domain are
```
shifucun, line 67: nit: this domain is... also can change the proto field to string domain = 2, ... facet_id = 3

internal/server/v2/observation/contained_in.go
```diff
@@ -190,27 +201,34 @@ func FetchContainedIn(
...
-							matchedDomain := false
-							for _, domain := range filter.Domains {
-								// To match domain or subdomain. For example, a provenance url of
-								// abc.xyz.com can match filter "xyz.com" and "abc.xyz.com".
-								if strings.HasSuffix(url.Hostname(), domain) {
-									matchedDomain = true
+						facetID := util.GetFacetID(facet)
+						// If there are facet filters, check that the cohort matches at
+						// least one filter. Otherwise, skip.
+						if facetFilters, ok := variableFacetFilters[variable]; ok {
```
shifucun, line 207: maybe make this a function that like: checkCohort(facet, filters[]) => bool? and this can be reused in the other file as well.

proto/v2/observation.proto
chejennifer: done

internal/server/v2/observation/contained_in.go
chejennifer: done

shifucun (approved):
nice!

chejennifer:
thanks for the review!

PR #1231, opened 2023-08-18
Support stat var search with remote mixer response merging

shifucun:


internal/server/handler_v1.go
```diff
@@ -482,3 +482,29 @@ func (s *Server) RecognizePlaces(
...
+// SearchStatVar implements API for Mixer.SearchStatVar.
+func (s *Server) SearchStatVar(
+	ctx context.Context, in *pb.SearchStatVarRequest,
+) (*pb.SearchStatVarResponse, error) {
+	localResp, err := statvar.SearchStatVar(ctx, in, s.store, s.cache)
+	if err != nil {
+		return nil, err
+	}
+	if len(localResp.StatVarGroups) == 0 && len(localResp.StatVars) == 0 && len(localResp.Matches) == 0 {
+		if s.metadata.RemoteMixerDomain != "" {
```
keyurva, line 495: Does it make sense to always call both in this case and return a merged response?

keyurva (approved):
Thanks for the fix!

internal/server/handler_v1.go
shifucun: Ideally yes. In current use scenarios, custom DC will almost always read from the remote mixer since it does not have custom token.

Also we would deprecate the search with embeddings very soon. So will not use the merge here.

shifucun:
thanks for review!

PR #1235, opened 2023-08-21
add PlaceFacet proto

n-h-diaz:


proto/stat.proto
```diff
@@ -47,6 +47,18 @@ message Facets {
   repeated Facet facets = 1;
 }
 
+// PlaceFacet contains the source and measurement information for a set of
+// statistical observations about a place.
+message PlaceFacet {
+  Facet facet = 1;
+  // Number of observations for given place + Facet.
```
shifucun, line 54: for a given place + variable + facet?

shifucun (approved):

proto/stat.proto
n-h-diaz: done, thanks!

PR #1247, opened 2023-08-28
Add golden test for topic entity and verify it works

shifucun:


internal/server/v2/observation/golden/existence/existence.json
```diff
@@ -20,6 +21,11 @@
         "test_entity": {}
       }
     },
+    "dc/topic/Businesses": {
```
shifucun, line 24: Note here only geoId/06 has data

PR #1248, opened 2023-08-28
add region suffix to un regions

n-h-diaz:
this is a temporary fix to avoid regressions in main dc from un regions


pradh (approved):
Thank you!

PR #1249, opened 2023-08-30
RecognizePlace: Implement adjectival place qualification

pradh:
We want to interpret China as a country in queries like [population across chinese provinces], but not in queries like [chinese speakers in california].

To do this, we qualify these "adjectival" usages only if followed immediately by place types.  In this case, we leave the place type in the original query so NL can consume it.

The adjectival words were largely from https://en.wikipedia.org/wiki/List_of_adjectival_and_demonymic_forms_for_countries_and_nations.  On that I ran place-name-resolver to map to DCID and skimmed thru the mappings.


internal/store/files/recog_place_store.go
```diff
@@ -33,9 +33,33 @@ var recogPlaceMapCSVContent []byte // Embed CSV as []byte.
...
+//go:embed "WorldGeosForPlaceRecognitionAdjectivalNames.csv"
+var recogPlaceAdjectivalNamesCSVContent []byte // Embed CSV as []byte.
 //go:embed "BogusPlaceNames.csv"
 var recogPlaceBogusPlaceNamesCSVContent []byte // Embed CSV as []byte.
 
+var (
+	// These are suffixes one of which needs to exist when dealing with adjectival place
+	// names like american, indian, african, etc.  This is is to avoid misinterpreting
+	// "african american", "chinese speakers", etc as places.
+	adjectivalSuffixes = [...]string{
```
Spaceenter, line 45: nit: doesn't need the "..."?

Spaceenter (approved):

PR #1256, opened 2023-09-10
Update the response of contained in place, time series facet API

shifucun:
Currently when fetching the facets of "contained in places for all date", the response is per child place based, which is inconsistent with the "contained in places, specific date" API call.

Updating to match the latter one, where using "" as a catch-all place.


chejennifer (approved):
thanks!

PR #1261, opened 2023-09-12
Drop region suffix for cases where there's no conflict

pradh:
This will allow answering questions about "Central Asia", "Latin America", "Eastern Europe", etc.

Since we match the longest pattern, these should not conflict with "asia", "america", "europe".


jehangiramjad (approved):
Thanks!

pradh:
Thanks!

PR #1262, opened 2023-09-12
Place Recog: Fix regions some more

pradh:
When the query comes in with "sub-saharan africa", NL passes in "sub saharan africa", so in the names replace `-` with space.  Also trim leading spaces when there's multiple names.

I've filed https://github.com/datacommonsorg/mixer/issues/1263 for the parser to be more tolerant.


jehangiramjad (approved):
Why not add both hyphenated and spaces names?

pradh:
> Why not add both hyphenated and spaces names?

Good idea!  Done.

PR #1264, opened 2023-09-13
Always prefer dcbranch over other import groups

shifucun:
This way, dcbranch could be used to block/override bad data.

### DIFF
--- a/internal/server/stat/stat_reader.go
+++ b/internal/server/stat/stat_reader.go
@@ -200,7 +200,7 @@ func ReadStatsPb(
 			result[p][sv] = &pb.ObsTimeSeries{}
 		}
 	}
-	for _, btData := range btDataList {
+	for i, btData := range btDataList {
 		for _, row := range btData {
 			place := row.Parts[0]
 			sv := row.Parts[1]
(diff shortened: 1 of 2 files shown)


internal/server/stat/stat_reader.go
```diff
@@ -211,6 +211,10 @@ func ReadStatsPb(
 			)
 			result[place][sv].PlaceName = obs.PlaceName
 		}
+		// Always prefer data from branch cache.
+		if len(btData) > 0 && btGroup.TableNames()[i] == btGroup.BranchTableName() {
```
pradh, line 215: To confirm, should we do this per SV / place?  Because what if we fix a certain SV but not others?

shifucun: Updated

internal/server/stat/stat_reader.go
```diff
@@ -192,18 +192,29 @@ func ReadStatsPb(
 		return nil, err
 	}
 	result := map[string]map[string]*pb.ObsTimeSeries{}
+	hasData := map[string]map[string]bool{}
 	for _, p := range places {
 		if _, ok := result[p]; !ok {
 			result[p] = map[string]*pb.ObsTimeSeries{}
+			hasData[p] = map[string]bool{}
```
pradh, line 199: nit: hasData -> hasBranchData

pradh (approved):
Thanks!

internal/server/stat/stat_reader.go
shifucun: Done

pradh (approved):
Nice!

PR #1268, opened 2023-09-18
Make SQL client general and add a CloudSQL connector

shifucun:
This is to prepare using SQL backend in the Cloud


internal/sqldb/sqlite/sqlite.go
```diff
@@ -0,0 +1,72 @@
...
+	);
+	`
+	_, err = sqlClient.Exec(tripleStatement)
+	if err != nil {
+		return nil, err
+	}
+
+	observationStatement := `
+	CREATE TABLE observations (
+		entity TEXT,
```
keyurva, line 60: Do we want this to be `entity` or `dcid`?

internal/server/data/import.go
```diff
@@ -32,36 +30,30 @@ func Import(
 	in *pb.ImportRequest,
 	st *store.Store,
 	metadata *resource.Metadata,
-	openSql bool,
 ) (*resource.Cache, error) {
-	var err error
-	if err = writer.Write(
+	// First clear the tables.
+	err := sqldb.ClearTables(st.SQLClient, metadata.SQLitePath)
```
keyurva, line 35: Is it worthwhile backing up vs deleting tables?

keyurva (commented):
Thanks Bo. This generalization will help a lot.

internal/server/data/import.go
shifucun: I think eventually yes. I can imagine there is an endpoint for user to backup a table and have it in a "admin" UI. This is more of a prototype workflow for now.

internal/sqldb/sqlite/sqlite.go
shifucun: should be dcid! I want to get e2e up running then clean this up.

keyurva (approved):

shifucun:
thanks!

PR #1269, opened 2023-09-19
Consolidate logic between sqlite and mysql

shifucun:
* CreateTable() func to create tables as a one time setup
* Branch some SQL syntax that are not compatible between sqlite and mysql
* Use sqlDataPath to represent raw csv data path. TODO: make this working for GCS path.

### DIFF
--- a/docs/developer_guide.md
+++ b/docs/developer_guide.md
@@ -82,14 +82,37 @@ root of this repo.
 
 ```bash
 # In repo root directory
-echo mixer_api_key=<YOUR API KEY>
+export mixer_api_key=<YOUR API KEY>
 go run cmd/main.go \
     --use_sqlite=true \
     --use_bigquery=false \
     --use_base_bigtable=false \
     --use_branch_bigtable=false \
     --use_maps_api=false \
-    --sqlite_path=$PWD/data \
+    --sql_data_path=$PWD/data \
+    --remote_mixer_domain=https://api.datacommons.org
+```
+
+## Start Mixer as a gRPC server backed by CloudSQL Database
+
+Mixer can load data stored from Google CloudSQL. This requires setting flag:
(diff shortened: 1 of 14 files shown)


docs/developer_guide.md
```diff
@@ -82,14 +82,37 @@ root of this repo.
...
-echo mixer_api_key=<YOUR API KEY>
+export mixer_api_key=<YOUR API KEY>
 go run cmd/main.go \
     --use_sqlite=true \
     --use_bigquery=false \
     --use_base_bigtable=false \
     --use_branch_bigtable=false \
     --use_maps_api=false \
-    --sqlite_path=$PWD/data \
+    --sql_data_path=$PWD/data \
```
keyurva, line 92: Wondering if this should be called something like `csv_data_path`?

internal/util/util.go
```diff
@@ -594,9 +597,23 @@ func SQLInParam(n int) string {
...
+	case *mysql.MySQLDriver:
+		result := ""
+		for i := 0; i < n-1; i++ {
+			result += "SELECT ? UNION ALL "
+		}
+		result += "SELECT ?"
+		return result
+	default:
+		// This should not happen.
+		log.Println("Invalid sql driver")
```
keyurva, line 614: Throw an exception in this case?

keyurva (approved):

pradh (approved):
Nice!

internal/util/util.go

docs/developer_guide.md
shifucun: In sqlite case, this is also the folder that contains the db files. I think sql_xx conveys the message that this is a folder for sql data?

shifucun:
Thank you both for review!

PR #1270, opened 2023-09-20
Support GCS as raw csv storage for SQL based database

shifucun:


internal/sqldb/writer.go
```diff
@@ -113,56 +130,122 @@ func Write(sqlClient *sql.DB, resourceMetadata *resource.Metadata) error {
...
 	}
 
 	// Load header.
 	header := records[0]
 	if len(header) < 3 {
 		return nil, nil, status.Errorf(codes.FailedPrecondition,
-			"Less than 3 columns in CSV file %s", csvFile)
+			"Less than 3 columns in CSV file %s", provID)
 	}
 	numColumns := len(header)
```
keyurva, line 251: Based on our earlier discussion, the resolving logic will be removed and mixer will only load pre-resolved entities, correct?

keyurva (approved):

internal/sqldb/writer.go
shifucun: I will extend mixer to take resolved entities, but it should still keep the capability to take unresolved csv.

pradh (commented):
Very cool!

PR #1275, opened 2023-09-27
Use preferred facet, not latest facet for place page population data

shifucun:


internal/server/v1/page/golden/place_page/asm.Demographics.json
```diff
@@ -2600,12 +2600,12 @@
...
-      "value": 83196078,
+      "date": "2019",
+      "value": 83019213,
       "metadata": {
-        "importName": "WorldDevelopmentIndicators",
-        "provenanceUrl": "https://datacatalog.worldbank.org/dataset/world-development-indicators/",
-        "observationPeriod": "P1Y"
+        "importName": "EurostatData",
+        "provenanceUrl": "https://ec.europa.eu/eurostat/data/database",
+        "measurementMethod": "EurostatRegionalPopulationData"
```
pradh, line 2608: nit: Maybe we should prefer WDI over Eurostat?  I can send follow-up

https://github.com/datacommonsorg/mixer/blob/d7022933628a6df248842981d3cc75cbad697b8e/internal/server/ranking/ranking.go#L66-L74

internal/server/v1/page/golden/place_page/asm.Energy.json
```diff
@@ -819,11 +819,12 @@
       }
     },
     "country/NZL": {
-      "date": "2022",
-      "value": 5151600,
+      "date": "2021",
+      "value": 5111400,
       "metadata": {
-        "importName": "New_Zealand_Census",
-        "provenanceUrl": "https://www.stats.govt.nz/"
```
pradh, line 826: We might perhaps add the per-country censuses ahead of WDI and Eurostat too

pradh (approved):
Thanks Bo!

internal/server/v1/page/golden/place_page/asm.Energy.json
shifucun: That would be good. @ajaits can suggest some imports?

PR #1277, opened 2023-10-02
Migrate ESP to v2 and fixes service management quota issue

shifucun:
More details and ration can be found in https://cloud.google.com/endpoints/docs/grpc/get-started-kubernetes-engine-espv2#deploying_the_sample_api_and_esp_to_the_cluster

Also removed deprecated flag from memdb.


chejennifer (approved):
nice

PR #1288, opened 2023-10-20
Batch write to SQL database; Increase envoy timeout to 600s

shifucun:
This greatly improves the sql write throughput.


internal/sqldb/writer.go
```diff
@@ -267,12 +270,33 @@ func writeObservations(
...
+				return err
+			}
+			log.Printf("[INSERT] observations: %d entries", batchSize)
+			valueStrings = valueStrings[:0] // reset the slices for the next batch
+			valueArgs = valueArgs[:0]
+		}
+	}
+	// Insert remaining data that didn't make a full batch
+	if len(valueStrings) > 0 {
+		sqlStmt := fmt.Sprintf(`INSERT INTO observations(entity,variable,date,value,provenance) VALUES %s`, strings.Join(valueStrings, ","))
```
keyurva, line 294: Consider extracting this into a helper function to avoid duplication of code.

internal/sqldb/writer.go
```diff
@@ -281,12 +305,31 @@ func writeTriples(
...
+				return err
+			}
+			log.Printf("[INSERT] triples: %d entries", batchSize)
+			valueStrings = valueStrings[:0] // reset the slices for the next batch
+			valueArgs = valueArgs[:0]
+		}
+	}
+	// Insert remaining data that didn't make a full batch
+	if len(valueStrings) > 0 {
+		sqlStmt := fmt.Sprintf(`INSERT INTO triples(subject_id,predicate,object_id,object_value) VALUES %s`, strings.Join(valueStrings, ","))
```
keyurva, line 327: Similar to observations, consider extracting this into a helper function to avoid duplication of code.

keyurva (approved):
This will speed things up significantly - thanks!

internal/sqldb/writer.go
shifucun: Done

internal/sqldb/writer.go
shifucun: Done

shifucun (commented):
thanks for review!

PR #1289, opened 2023-10-23
[RSI] Read triples csv (from RSI script) instead of generating them

shifucun:
Also did some minor refactor to clean up the structure.

### DIFF
--- a/internal/sqldb/csv.go
+++ b/internal/sqldb/csv.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/csv"
 	"io"
+	"log"
 	"os"
 	"path/filepath"
 	"strings"
@@ -173,6 +174,10 @@ func processObservationCSV(
 	for i := 1; i < numRecords; i++ {
 		record := records[i]
 		for j := 2; j < numColumns; j++ {
+			if record[j] == "" {
+				log.Printf("Skip empty value record: %s", record)
+				continue
+			}
 			observations = append(observations, &observation{
 				entity:     record[0],
 				date:       record[1],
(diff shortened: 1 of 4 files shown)


internal/sqldb/csv.go
```diff
@@ -0,0 +1,220 @@
...
+			"Less than 3 columns in CSV file %s", provID)
+	}
+	numColumns := len(header)
+
+	// Generate observations.
+	observations := []*observation{}
+	for i := 1; i < numRecords; i++ {
+		record := records[i]
+		for j := 2; j < numColumns; j++ {
+			observations = append(observations, &observation{
```
keyurva, line 176: Perhaps skip adding observation if value is empty?

internal/sqldb/csv.go
```diff
@@ -0,0 +1,220 @@
...
+		if err != nil {
+			return nil, nil, err
+		}
+		for _, file := range files {
+			if fName := file.Name(); strings.HasSuffix(fName, ".csv") {
+				f, err := os.Open(filepath.Join(dir, fName))
+				if err != nil {
+					return nil, nil, err
+				}
+				if strings.HasPrefix(fName, "observations") {
```
keyurva, line 120: There's some duplication that can be avoided by creating an abstraction for local and GCS files. But ok to go with what we have for now since this code will be deleted once RSI writes to sqlite natively.

keyurva (approved):
Coming together nicely!

(Reminder - merge this only after my next PR to reuse compatible column names as ids)

internal/sqldb/csv.go
shifucun: Done

PR #1290, opened 2023-10-24
Require RSI raw csv files path in /import API request; Fix SQL query for child sv

shifucun:
* Specify raw csv files path make it more flexible for client and greatly simplifies mixer logic.
* Query child sv should use "name" instead of "description".


keyurva (approved):
Thanks.

PR #1291, opened 2023-10-25
Read all entity observations from SQL by join observation with triples table on typeOf predicate

shifucun:


test/triples.csv
```diff
@@ -36,3 +36,6 @@ test_var_3,typeOf,StatisticalVariable,
 test_var_3,measuredProperty,test_var_3,
 test_var_3,statType,measuredValue,
 test_var_3,name,,"non-place entity variable"
+ein/1,typeOf,Company,
```
keyurva, line 39: How are these files used? Are there tests that use them?

keyurva (approved):

internal/server/v2/observation/contained_in.go
```diff
@@ -232,26 +223,58 @@ func FetchContainedIn(
...
+					AND o.variable IN (%s)
+				`,
+				childType,
+				variablesStr,
+			)
+			if queryDate != "" && queryDate != LATEST {
+				query += fmt.Sprintf("AND date = (%s) ", queryDate)
+			}
+			query += "ORDER BY date ASC;"
+			rows, err := store.SQLClient.Query(query)
```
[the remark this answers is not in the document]
shifucun: That's a really good point. Let me do it in a follow up PR.

test/triples.csv
shifucun: Yes, these csvs are loaded into datacommons.db for testing via "./import.sh". I will document them too.

shifucun:
thanks for review!

PR #1297, opened 2023-11-17
Data Release 2023-11-17

chejennifer:
- updates all the tables to a version built after 2023-10-19 which was when a field was added to the PlaceVariableFacet caches


ajaits (approved):
diff due to US PEP and CDC refresh are good.

pradh (approved):

PR #1300, opened 2023-11-28
Major refactor of server level Cache object and operations

shifucun:
* Re-organize the `statvar` packages into sub packages for clarity and avoid import cycle
* Use mutex to guard Cache object live update
* Change /import endpoint to /update-cache. This will be needed for the web compose solution to switch to use RSI for custom data import.


internal/server/cache/cache.go
```diff
@@ -16,30 +16,78 @@ package cache
...
+	// SVG search index
+	svgSearchIndex *resource.SearchIndex
+	// Custom provenance from SQL storage
+	customProvenances map[string]*pb.Facet
+	// Lock for updating cache
+	mu sync.RWMutex
+}
+
+func (cache *Cache) ParentSvg() map[string][]string {
+	cache.mu.RLock()
```
keyurva, line 61: One approach for avoiding read locks completely is maintaining cached data in a separate structure, say `CachedData` and simply swapping it on cache refresh. You can still maintain a single lock for when you swap but should not need it otherwise.

```
struct CachedData {
  parentSvgs map[string][]string
  ...
}

struct Cache {
  lock sync.Mutex
  data CachedData
}
```

keyurva, line 61: Wondering if this can be simplified. One approach for avoiding read locks completely is maintaining cached data in a separate structure, say `CachedData` and simply swapping it on cache or index refresh. You can still maintain a single lock for when you swap but should not need it otherwise.

```
struct CachedData {
  parentSvgs map[string][]string
  ...
}

struct Cache {
  lock sync.Mutex
  data CachedData
}
```

internal/server/handler.go
```diff
@@ -309,14 +308,18 @@ func (s *Server) BulkFindEntities(
...
-func (s *Server) Import(
-	ctx context.Context, in *pb.ImportRequest,
-) (*pb.ImportResponse, error) {
-	cache, err := data.Import(ctx, in, s.store)
-	if err != nil {
+// UpdateCache implements API for Mixer.UpdateCache
+func (s *Server) UpdateCache(
+	ctx context.Context, in *pb.UpdateCacheRequest,
+) (*pb.UpdateCacheResponse, error) {
+	if err := s.cachedata.UpdateSVGCache(ctx, s.store); err != nil {
```
keyurva, line 315: Consider encapsulating this logic in a single Cache.UpdateCache(...) call vs the API layer implementing it.

internal/server/statvar/fetcher/svg_fetcher.go
```diff
@@ -0,0 +1,176 @@
...
+					},
+				)
+			}
+		}
+	}
+	if store.SQLClient != nil {
+		// Query for all the stat var group node
+		query :=
+			`
+					SELECT t1.subject_id, t2.object_value, t3.object_id
```
keyurva, line 108: At some point we should consolidate all sql calls and BT calls into their respective modules. Currently it's spread out across multiple modules requiring them to be knowledgable of the underlying db model and making refactoring expensive.

Can be done in a separate PR but just noting it here.

keyurva (commented):
These are substantial changes and would've required lots of care. Thanks for doing it.

internal/server/handler.go
shifucun: Done

internal/server/statvar/fetcher/svg_fetcher.go
shifucun: Yes, there is a module for that. Need to move the rest code there. I can do that when touching them in a smaller PR.

internal/server/cache/cache.go
keyurva: Reading about `atomic.Pointer` (sounds similar to Java's `AtomicReference`), it should make things simpler for us. We may not need to maintain a lock at all (for writes or reads). Instead maintain an `*atomic.Pointer[CachedData]` and simply `Swap` it on mutation.

~ make all the caller site change minimal, by adding "()"

I agree that we should have methods for accessing the data vs via struct fields. With atomic pointers, the implementation of the accessors would be something like this:

```
func (c *Cache) ParentSvgs() map[string][]string {
	return c.dataPointer.Load().parentSvgs
}
```

shifucun: Gave it a try and updated server cache as atomic pointer now. Also removed the Update() function and now do a swap for a newly created cache object on update.

keyurva: Nice - no need for locks anymore! 

nit: Not a fan of exposing it as an atomic pointer in server (vs encapsulating the atomicity in the cache module). This results in requiring calling `Load()` everywhere it is used. But this is good for now. We can cleanup later.

keyurva (approved):

shifucun:
thanks for review

PR #1303, opened 2023-11-30
Cache entity variable existence from SQL database

shifucun:
* Query all <entity, variable> pair from custom DC SQL database and cache them in-memory
* Improve the count function by using this cache information.

### DIFF
--- a/internal/server/cache/cache.go
+++ b/internal/server/cache/cache.go
@@ -26,6 +26,7 @@ import (
 	"github.com/datacommonsorg/mixer/internal/server/statvar/hierarchy"
 	"github.com/datacommonsorg/mixer/internal/sqldb/sqlquery"
 	"github.com/datacommonsorg/mixer/internal/store"
+	"github.com/datacommonsorg/mixer/internal/util"
 )
 
 const (
@@ -52,7 +53,7 @@ type Cache struct {
 	// Provenance from SQL storage
 	sqlProvenances map[string]*pb.Facet
 	// SQL database entity, variable existence pairs
-	sqlEntityVariable map[string]map[string]struct{}
+	sqlExistenceMap map[util.EntityVariable]struct{}
 	// CacheOption for this Cache object
 	options CacheOptions
 }
@@ -77,8 +78,8 @@ func (c *Cache) SQLProvenances() map[string]*pb.Facet {
 	return c.sqlProvenances
 }
 
-func (c *Cache) SQLEntityVariable() map[string]map[string]struct{} {
-	return c.sqlEntityVariable
+func (c *Cache) SQLExistenceMap() map[util.EntityVariable]struct{} {
+	return c.sqlExistenceMap
 }
 
 func (c *Cache) Options() *CacheOptions {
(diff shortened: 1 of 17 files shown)


internal/sqldb/sqlquery/entity_variable.go
```diff
@@ -0,0 +1,48 @@
...
+	"database/sql"
+	"time"
+
+	"github.com/datacommonsorg/mixer/internal/util"
+)
+
+// EntityVariable returns all existent entity, variable pairs
+func EntityVariable(sqlClient *sql.DB) (map[string]map[string]struct{}, error) {
+	defer util.TimeTrack(time.Now(), "SQL: EntityVariable")
+	query := "SELECT DISTINCT entity, variable FROM observations o"
```
keyurva, line 27: A couple questions:

1. How long does this query take on Cloud SQL for the large ONE dataset?
2. Any concerns about the size of this map? If there are 200 entities and 6000 vars, there will be 1.2M entries.

internal/server/cache/cache.go
```diff
@@ -49,8 +49,10 @@ type Cache struct {
 	blocklistSvgs map[string]struct{}
 	// SVG search index
 	svgSearchIndex *resource.SearchIndex
-	// Custom provenance from SQL storage
-	customProvenances map[string]*pb.Facet
+	// Provenance from SQL storage
+	sqlProvenances map[string]*pb.Facet
+	// SQL database entity, variable existence pairs
+	sqlEntityVariable map[string]map[string]struct{}
```
keyurva, line 55: This article talks about using struct keys in such situations. Consider using it if it makes things easier. https://go.dev/blog/maps#key-types

internal/server/cache/cache.go
```diff
@@ -117,13 +123,17 @@ func NewCache(
...
+		if err != nil {
+			return nil, err
+		}
+		c.sqlProvenances = sqlProv
+		sqlExistenceMap, err := sqlquery.EntityVariable(store.SQLClient)
 		if err != nil {
 			return nil, err
 		}
-		c.customProvenances = customProv
+		c.sqlEntityVariable = sqlExistenceMap
```
keyurva, line 136: Actually `sqlExistenceMap` might be a good name for the cache entry itself.

internal/server/cache/cache.go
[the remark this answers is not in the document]
shifucun: Changed to sqlExistenceMap

shifucun: This is a very good ideal. Updated

internal/server/cache/cache.go
shifucun: Done

internal/sqldb/sqlquery/entity_variable.go
shifucun: 1. 6-7 seconds
2. Give the map is keyed by <e,v> with value of an empty struct, each entry takes O(10) bytes, so for this case it's in the order 12Mb (with a factor of 5, it's 60Mb) which is a reasonable size.

PR #1307, opened 2023-12-18
Add an /internal/import-table API to fetch import table data

juliawu:
In order to support adding an "import history" table to the /admin page of custom DCs, this PR adds a new internal API endpoint `/internal/import-table/` to fetch import history.

Testing: Added test for GetImportTableData, and accessed the new endpoint while running locally.


internal/sqldb/sqlquery/get_import_table.go
```diff
@@ -0,0 +1,79 @@
...
+	"encoding/json"
+	"time"
+
+	pb "github.com/datacommonsorg/mixer/internal/proto"
+	"github.com/datacommonsorg/mixer/internal/util"
+)
+
+// Represents "metadata" column of import table
+type ImportMetadata struct {
+	NumObs  int32
```
keyurva, line 28: Wondering if we want to allow them to be `nil` by making them pointers? We might add more metadata in the future and only a subset of fields may have values at different times.

proto/internal.proto
```diff
@@ -40,4 +41,23 @@ message UpdateCacheRequest {
...
+
+// Request to get data in the import table
+message GetImportTableDataRequest {
+}
+
+// Response of GetImportTableDataRequest
+message GetImportTableDataResponse{
+  message ImportData {
+    message ImportMetadata{
+      int32 num_obs = 5;
```
keyurva, line 54: To accommodate unset values, consider making them `optional`.

keyurva (commented):
Very cool!

internal/sqldb/sqlquery/get_import_table.go
juliawu: Good idea, made them pointers.

proto/internal.proto
juliawu: Good idea, made them `optional`.

keyurva (approved):

PR #1312, opened 2024-01-09
Create custom DC mysql tables for new database

shifucun:


pradh (approved):

internal/sqldb/createtables.go
```diff
@@ -0,0 +1,50 @@
...
+// limitations under the License.
+
+package sqldb
+
+import (
+	"database/sql"
+)
+
+// CreateTables creates tables in MySQL database if they are not present.
+func CreateTables(sqlClient *sql.DB) error {
```
keyurva, line 22: I thought mixer can deal with the scenario where the tables don't exist yet.

Let's discuss this more. This creates a tight coupling between RSI and Mixer and will require keeping both in sync all the time.

shifucun: This is a fallback safe check. It only creates table if they do not exist to prevent query errors. RSI can create tables, then this will be a no op.

keyurva: RSI creates tables only if they don't exist. Also, creating them in 2 places can create maintenance issues.

Let's discuss if there is a cleaner / simpler option.

PR #1314, opened 2024-01-17
Build stat var search index for custom DC data

shifucun:
Setting SearchNames in StatVarGroupNode_ChildSV protobuf does the job.


internal/server/statvar/fetcher/svg_fetcher.go
```diff
@@ -165,7 +165,11 @@ func FetchAllSVG(
 			}
 			result[svg].ChildStatVars = append(
 				result[svg].ChildStatVars,
-				&pb.StatVarGroupNode_ChildSV{Id: sv, DisplayName: name},
+				&pb.StatVarGroupNode_ChildSV{
+					Id:          sv,
+					DisplayName: name,
+					SearchNames: []string{name},
```
keyurva, line 171: Do you have access to the SV descriptions? If so, consider adding it to `SearchNames`.

Also, it'll be good to start adding tests for custom dc functionality.

keyurva (approved):

internal/server/statvar/fetcher/svg_fetcher.go
shifucun: added description and golden tests

internal/server/statvar/fetcher/svg_fetcher.go
```diff
@@ -165,7 +168,11 @@ func FetchAllSVG(
 			}
 			result[svg].ChildStatVars = append(
 				result[svg].ChildStatVars,
-				&pb.StatVarGroupNode_ChildSV{Id: sv, DisplayName: name},
+				&pb.StatVarGroupNode_ChildSV{
+					Id:          sv,
+					DisplayName: name,
+					SearchNames: []string{name, description},
```
keyurva, line 174: Consider checking that `description` has a value before adding it.

shifucun: done

PR #1317, opened 2024-02-07
Increase ESP->gRPC server timeout from default 15s to 60s; Make other timeout 60s as well

shifucun:


esp/envoy-config.yaml
```diff
@@ -38,7 +38,7 @@ static_resources:
                             prefix: "/"
                           route:
                             cluster: grpcserver
-                            timeout: 600s
+                            timeout: 60s
```
keyurva, line 41: Looks like we are bringing it down here and at a couple other places. Assuming this is expected?

keyurva (approved):

esp/envoy-config.yaml
pradh: Do we want to drop those?  I worry if that might cause new timeout errors?

pradh (approved):

esp/envoy-config.yaml
shifucun: As discussed offline, we should expect our APIs return in less than 1 min. Anything longer should indicate a performance issue and needs to be fixed.

PR #1318, opened 2024-02-07
Coordinate recon: support type filtering

pradh:
This is to improve the performance of coordinate-recon API.  TBD how much it improves in practice.

More context in b/321108211


shifucun (approved):
nice!

pradh:
Thanks for the review!

shifucun:
Maybe add an example in https://docs.datacommons.org/api/rest/v2/resolve Example 2

pradh:
> Maybe add an example in https://docs.datacommons.org/api/rest/v2/resolve Example 2

SG, will send a PR.

PR #1320, opened 2024-02-13
Update BLS LAUS ranking

chejennifer:
rank P1Y highest for BLS LAUS data


chejennifer:
Discussed offline that this is no longer needed because we will make the change on the website side instead

PR #1322, opened 2024-02-16
Update SSL Certificates in helm charts

juliawu:
Updates the `multi-domain-<year>` ssl certificates linked in our helm charts from the expired `multi-domain-2022` to the newly created `multi-domain-2024`. A `multi-domain-2024` ssl cert was created for each of their host projects manually in GCP.


shifucun (approved):
Thank you so much for all the fix!

PR #1325, opened 2024-02-21
Remove SVGs from search index

chejennifer:
Remove SVGs from stat var hierarchy search index and also remove it from the result.

Note: removing the svg field from the SearchStatVarRequest proto should be safe because this response is consumed by the client and the client checks if there is a svg field or not: https://github.com/datacommonsorg/website/blob/master/static/js/utils/search_utils.tsx#L43


proto/stat_var.proto
```diff
@@ -239,8 +230,6 @@ message SearchStatVarResponse {
   // A ranked list of stat vars that aren't included in any of the returned stat
   // var groups.
   repeated EntityInfo stat_vars = 1;
-  // A list of stat var group results ranked by relevance.
```
shifucun, line 242: can we still mark this field as deprecated, just in case custom DC got proto field error.

internal/server/statvar/hierarchy/statvar_hierarchy_util.go
```diff
@@ -146,6 +147,7 @@ func RemoveSvg(
 // BuildParentSvgMap gets the mapping of svg/sv id to the parent svg for that
 // svg/sv. Only gets the parent svg that have a path to the root node.
 func BuildParentSvgMap(rawSvg map[string]*pb.StatVarGroupNode) map[string][]string {
+	fmt.Println("in build parent svg map")
```
shifucun, line 150: remove logging?

internal/server/statvar/hierarchy/statvar_hierarchy_util.go
```diff
@@ -177,9 +179,11 @@ func BuildParentSvgMap(rawSvg map[string]*pb.StatVarGroupNode) map[string][]stri
 			}
 		}
 	}
+	fmt.Println("finished for loop")
```
shifucun, line 182: remove all logging

shifucun (commented):
thanks for the update!

internal/server/statvar/search/statvar_search.go
```diff
@@ -40,24 +41,26 @@ func SearchStatVar(
 ) (
 	*pb.SearchStatVarResponse, error,
 ) {
+	fmt.Println("HERE")
```
pradh, line 44: nit: drop debug msg?

proto/stat_var.proto
pradh: +1 `reserved 2;  // Deprecated "stat_var_groups"`

chejennifer: done

internal/server/statvar/search/statvar_search.go
chejennifer: done

pradh (approved):
Thanks!

I'm excited to see this change in action!

chejennifer:
thanks for the reviews!

PR #1343, opened 2024-04-01
Recon:  fix empty array access

pradh:
This suggests that the recon coordinate cache is staler than the triples cache, which can happen because the coordinate cache is built on-demand.

The likely sequence of events is that cl/597359495 modified Brazil AA2 geojsons (specifically seems to have removed some boundaries as part of fixing some buggy boundaries), but didn't rebuild the coordinate recon cache.   A fire event located within Brazil AA2 tries to look up the now removed geojson and fails.

Two part fix:
1.  Make the code more resilient
2. Rebuild recon cache

### DIFF
--- a/internal/server/recon/coordinate.go
+++ b/internal/server/recon/coordinate.go
@@ -107,8 +107,11 @@ func ResolveCoordinates(
 		return nil, err
 	}
 	s2PolygonMap := map[string]*s2.Polygon{}
-	for place, entities := range geoJSONData {
-		s2Polygon, err := parseGeoJSON(entities[0].Value)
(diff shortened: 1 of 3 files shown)


shifucun (approved):
Very nice findings. Wonder if we can have some test in g3 to alert building cache when geojson changes.

pradh:
> Very nice findings. Wonder if we can have some test in g3 to alert building cache when geojson changes.

Yes, good idea.  I'll add that presubmit alert...

keyurva:
Should we have a custom catch all error page that gets displayed for all uncaught errors (vs the default 500 internal server error page it currently shows)?

PR #1345, opened 2024-04-05
Update place recognition

chejennifer:
Update place recognition to also try to resolve entiites using the reconIdMap cache


chejennifer:
Discussed with prashanth offline, closing this & will put this logic in a new endpoint instead of adding onto place detection logic.

PR #1346, opened 2024-04-05
Add new recognize entities endpoint

chejennifer:
This new endpoint has the same request and response format as RecognizePlaces but uses the ReconIdMap cache to try to recognize entities in the query. Changes:
- add new endpoint and logic for that endpoint
- add types field to ReconEntity to reflect g3 changes
- update id recon to merge results from different tables


internal/server/recon/recognize.go
```diff
@@ -73,6 +73,138 @@ func RecognizePlaces(
...
+	for _, query := range in.GetQueries() {
+		id2spans := getId2Span(query)
+		if id2spans == nil {
+			continue
+		}
+		idsToResolve := []string{}
+		for id := range id2spans {
+			idsToResolve = append(idsToResolve, id)
+		}
+		resolvedIdEntities, err := GetResolvedIdEntities(ctx, store, "reconName", idsToResolve, "dcid")
```
pradh, line 96: use constants?

internal/server/recon/recognize.go
```diff
@@ -73,6 +73,138 @@ func RecognizePlaces(
...
+	spanTokens := strings.Split(query, " ")
+	for i := range spanTokens {
+		span := ""
+		// make n-grams from the span tokens
+		for j := i; j < len(spanTokens); j++ {
+			span = span + " " + spanTokens[j]
+			span = strings.TrimSpace(span)
+			id := strings.ReplaceAll(span, " ,", "")
+			id = strings.ReplaceAll(id, ",", "")
+			id = strings.ReplaceAll(id, "^", "")
```
pradh, line 197: Consider having a helper for "name to reconName", and commenting that this should following the corresponding logic in the flume pipeline (link to the func-name, g3 url)?

internal/server/recon/recognize.go
```diff
@@ -73,6 +73,138 @@ func RecognizePlaces(
...
+}
+
+// Takes a query and returns a map of id to use for resolution to the
+// original span (part of the query) for that id.
+func getId2Span(query string) map[string]map[string]struct{} {
+	id2spans := map[string]map[string]struct{}{}
+	spanTokens := strings.Split(query, " ")
+	for i := range spanTokens {
+		span := ""
+		// make n-grams from the span tokens
```
pradh, line 191: Should we have a limit on the N value?

internal/server/recon/recognize.go
```diff
@@ -73,6 +73,138 @@ func RecognizePlaces(
...
+	// assume a word got split in the middle and return empty list.
+	if len(newSpans[1]) > 0 && !strings.HasPrefix(newSpans[1], " ") {
+		return []string{}
+	}
+	return []string{strings.TrimSpace(newSpans[0]), splitSpan, strings.TrimSpace(newSpans[1])}
+}
+
+// Takes a query and returns a map of id to use for resolution to the
+// original span (part of the query) for that id.
+func getId2Span(query string) map[string]map[string]struct{} {
```
pradh, line 186: Could this perhaps use a unit-test?

internal/server/recon/recognize.go
```diff
@@ -73,6 +73,138 @@ func RecognizePlaces(
...
+	id2spans := map[string]map[string]struct{}{}
```
pradh, line 187: comment why single id can have >1 span?

internal/server/recon/recognize.go
```diff
@@ -73,6 +73,138 @@ func RecognizePlaces(
...
+			}
+		}
+
+		// get the list of resolved spans and sort them by longer span first
+		spans := []string{}
+		for span := range span2item {
+			spans = append(spans, span)
+		}
+		sort.Slice(spans, func(i, j int) bool {
+			return len(spans[i]) > len(spans[j])
```
pradh, line 121: When its equal, should we still sort deterministically?

internal/server/recon/recognize.go
```diff
@@ -73,6 +73,138 @@ func RecognizePlaces(
...
+	ctx context.Context,
+	in *pb.RecognizePlacesRequest,
+	store *store.Store,
+	resolveBogusName bool,
+) (*pb.RecognizePlacesResponse, error) {
+
+	resp := &pb.RecognizePlacesResponse{
+		QueryItems: map[string]*pb.RecognizePlacesResponse_Items{},
+	}
+	for _, query := range in.GetQueries() {
```
pradh, line 87: TODO: Parallelize queries if it matters

internal/server/recon/recognize.go
```diff
@@ -73,6 +73,138 @@ func RecognizePlaces(
...
+				newItems = append(newItems, queryRespItems[i+1:]...)
+				queryRespItems = append(queryRespItems[:i], newItems...)
+			}
+		}
+		resp.QueryItems[query] = &pb.RecognizePlacesResponse_Items{Items: queryRespItems}
+	}
+	return resp, nil
+}
+
+func getNewSpanList(origSpan string, splitSpan string) []string {
```
pradh, line 165: Describe function, should we have a unit test?

internal/server/recon/recognize.go
```diff
@@ -73,6 +73,138 @@ func RecognizePlaces(
...
+			spans = append(spans, span)
+		}
+		sort.Slice(spans, func(i, j int) bool {
+			return len(spans[i]) > len(spans[j])
+		})
+
+		// iterate through the resolved spans and add the resolved items to the
+		// response
+		queryRespItems := []*pb.RecognizePlacesResponse_Item{{Span: query}}
+		for _, span := range spans {
```
pradh, line 127: The logic in this loop is hard to follow.  Can you comment "why" we're doing what we're doing with some examples?

pradh (commented):
Nice!

I have trouble following the latter part of the algo, could you clarify that with more comments please?

internal/server/recon/recognize.go
chejennifer: done

internal/server/recon/recognize.go
chejennifer: done

internal/server/recon/recognize.go
chejennifer: done

internal/server/recon/recognize.go
chejennifer: done

internal/server/recon/recognize.go
chejennifer: done

internal/server/recon/recognize.go
chejennifer: done

internal/server/recon/recognize.go
chejennifer: added TODO

internal/server/recon/recognize.go
chejennifer: done

internal/server/recon/recognize.go
```diff
@@ -73,6 +77,174 @@ func RecognizePlaces(
...
+			}
+		}
+		resp.QueryItems[query] = &pb.RecognizePlacesResponse_Items{Items: queryRespItems}
+	}
+	return resp, nil
+}
+
+// Takes a span and splits it by another span into a list of 3 spans. Example:
+// origSpan: "ab cd ef g", splitSpan: "cd ef", result: ["ab", "cd ef", "g"]
+func getNewSpanList(origSpan string, splitSpan string) []string {
```
pradh, line 190: What if the same span appears multiple times in the query?

internal/server/recon/recognize.go
chejennifer: good idea, done

internal/server/recon/recognize.go
```diff
@@ -73,6 +77,174 @@ func RecognizePlaces(
...
+			if iCount > jCount {
+				return true
+			} else if jCount > iCount {
+				return false
+			} else {
+				return spans[i] < spans[j]
+			}
+		})
+
+		// iterate through the resolved spans and add the resolved items to the
```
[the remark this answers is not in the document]
chejennifer: done

internal/server/recon/recognize.go
chejennifer: realized that I did not handle this, updated the function to return a list of non-span part, span part, non-span part, etc.

internal/server/recon/recognize.go
```diff
@@ -73,6 +77,189 @@ func RecognizePlaces(
...
+}
+
+// Gets the reconName for a span.
+// The logic here should correspond to the logic for processing name into
+// reconName in flume (https://source.corp.google.com/piper///depot/google3/datacommons/prophet/flume_generator/triple_helper.cc;l=168-173)
+func getReconName(span string) string {
+	reconName := strings.ReplaceAll(span, " ,", "")
+	reconName = strings.ReplaceAll(reconName, ",", "")
+	reconName = strings.ReplaceAll(reconName, "^", "")
+	reconName = strings.ToLower(reconName)
```
pradh, line 231: We could also probably clean up consecutive spaces.  Add TODO?

internal/server/recon/recognize.go
```diff
@@ -73,6 +77,189 @@ func RecognizePlaces(
...
+		// this was not a valid split.
+		shouldSkipAdding := false
+		// if the current split is not empty and does not end in a space, assume
+		// span is part of the last word in current split which is not a valid split.
+		if len(split) > 0 && !strings.HasSuffix(split, " ") {
+			shouldSkipAdding = true
+		}
+		// if the next split is not empty and does not start with a space, assume
+		// span is part of the first word in next split which is not a valid split.
+		if len(splits[i+1]) > 0 && !strings.HasPrefix(splits[i+1], " ") {
```
pradh, line 206: Can this happen for the first split?  Example: "foo ba" is the span to search and splitting "foo bar foo ba r", we'll get:

"r ", " r"

pradh, line 206: Should we also think about `,` and `.` termination?  might be common in a list...

internal/server/recon/recognize.go
```diff
@@ -73,6 +77,189 @@ func RecognizePlaces(
...
+	splits := strings.Split(query, span)
+	// if span is not found in query, return just the query
+	if len(splits) < 2 {
+		return []string{query}
+	}
+	queryParts := []string{}
+	// prefix to add to the next split to be processed
+	nextPartPrefix := ""
+	// go through all the splits and add the span between each split if it was a
+	// valid split (we only want to split on complete words)
```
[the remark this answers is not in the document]
pradh: I wondered if this can be made more readable, and it slightly felt like split was adding complexity.  Here's a version that reads better, but I'm not sure because I wrote it :).  Anyway, see if that seems better or you'd prefer to stick with split + cleaning.

```
needle = 'foo ba'
haystack = 'foo bar foo ba r'

i = 0
parts = ['']
while i < len(haystack):
  if i == 0 or haystack[i - 1] == ' ':
    # A match starts at beginning of string or after a space!

    if haystack[i:].startswith(needle):
      eidx = i + len(needle)

      if len(haystack[i:]) == len(needle) or haystack[eidx] in [',', '.', ' ']:
        # A match ends with a space, comma, full-stop or end-of-string!
        parts.append(haystack[i:eidx])

        # Add an empty string for the next span.
        parts.append('')

        i += len(needle)
        continue

  # No luck, append to the last span
  parts[len(parts) - 1] += haystack[i]
  i += 1

# TODO: trim empty entries
```

internal/server/recon/recognize_test.go
```diff
@@ -390,3 +390,144 @@ func TestRankAndTrimCandidates(t *testing.T) {
...
+			[]string{"", "ab", "cd ef g"},
+		},
+		{
+			"ab cd ef g",
+			"ef g",
+			[]string{"ab cd", "ef g", ""},
+		},
+		{
+			"ab cd ef ab g",
+			"ab",
```
pradh, line 467: Might help to annotate the tests with some comments on what edge-case is tested?

pradh (commented):
Thanks for the edits!

internal/server/recon/recognize.go
```diff
@@ -73,6 +77,189 @@ func RecognizePlaces(
...
+
+		// get the list of resolved spans and sort them by more words first. If two
+		// spans have the same number of words, sort alphabetically.
+		spans := []string{}
+		for span := range span2item {
+			spans = append(spans, span)
+		}
+		sort.Slice(spans, func(i, j int) bool {
+			iCount := strings.Count(spans[i], " ")
+			jCount := strings.Count(spans[j], " ")
```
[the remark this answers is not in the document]
chejennifer: done

internal/server/recon/recognize.go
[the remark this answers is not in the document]
chejennifer: i like this, updated

internal/server/recon/recognize.go
```diff
@@ -73,6 +77,189 @@ func RecognizePlaces(
...
+		// if this is the last item in the list, add the query part and finish
+		if i == len(splits)-1 {
+			queryParts = append(queryParts, strings.TrimSpace(queryPart))
+			break
+		}
+		// Whether or not to skip adding anything in this round. We want to skip if
+		// this was not a valid split.
+		shouldSkipAdding := false
+		// if the current split is not empty and does not end in a space, assume
+		// span is part of the last word in current split which is not a valid split.
```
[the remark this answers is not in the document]
chejennifer: updated the algorithm

internal/server/recon/recognize.go
chejennifer: added TODO

internal/server/recon/recognize.go
chejennifer: updated the algorithm and this shouldn't happen

chejennifer: done

internal/server/recon/recognize_test.go
chejennifer: done

internal/server/recon/recognize.go
```diff
@@ -170,60 +174,67 @@ func getItemsForSpans(spans []string, query string, span2item map[string]*pb.Rec
...
+	i := 0
+	parts := []string{""}
+
+	for i < len(query) {
+		// a valid span match is one that starts at the beginning of the query or
+		// after a word separator
+		validSpanStart := true
+		if i > 0 {
+			_, validSpanStart = wordSeparators[query[i-1]]
 		}
```
pradh, line 193: Can we make this a single statement:

```
if i < len(query) && wordSeparators[query[i-1]] {
}
```

If you stored "true" value in wordSeparators, per https://stackoverflow.com/a/2051011 ?

internal/server/recon/recognize.go
```diff
@@ -73,6 +81,196 @@ func RecognizePlaces(
...
+// A single id can have multiple spans because we process a span by removing
+// certain characters to get an id:
+// E.g., "a^b" and "ab" would both have the id "ab"
+func getId2Span(query string) map[string]map[string]struct{} {
+	id2spans := map[string]map[string]struct{}{}
+	spanTokens := strings.Split(query, " ")
+	for i := range spanTokens {
+		span := ""
+		// make n-grams from the span tokens
+		for j := i; j < int(math.Min(float64(len(spanTokens)), reconNGramLimit+1)); j++ {
```
pradh, line 258: nit: Can this be a constant hoisted before the loops?

internal/server/recon/recognize_test.go
```diff
@@ -390,3 +390,173 @@ func TestRankAndTrimCandidates(t *testing.T) {
...
+
+func TestGetId2Span(t *testing.T) {
+	for _, c := range []struct {
+		query string
+		want  map[string]map[string]struct{}
+	}{
+		{
+			"entities1, Entities^2 , and entities,^2",
+			map[string]map[string]struct{}{
+				"entities1": {
```
pradh, line 402: Same as other tests, can you please add a comment each?

internal/server/recon/recognize_test.go
```diff
@@ -390,3 +390,173 @@ func TestRankAndTrimCandidates(t *testing.T) {
...
+		if diff := cmp.Diff(got, c.want); diff != "" {
+			t.Errorf("SplitQueryBySpan for query %s and span %s got diff: %s", c.query, c.span, diff)
+		}
+	}
+}
+
+func TestGetItemsForSpans(t *testing.T) {
+	cmpOpts := cmp.Options{
+		protocmp.Transform(),
+	}
```
pradh, line 535: comment what is going on here?

pradh (approved):
Thanks for all the changes.  Looks good to me, a few final nits.

Please wait for Bo to have a quick look at it too?

internal/server/recon/recognize.go
chejennifer: done

internal/server/recon/recognize.go
chejennifer: done

internal/server/recon/recognize_test.go
chejennifer: done

internal/server/recon/recognize_test.go
chejennifer: done

proto/recon.proto
```diff
@@ -143,6 +145,8 @@ message RecognizePlacesRequest {
 message RecognizePlacesResponse {
   message Place {
     string dcid = 1;
+    // All types of the entity. Ordered by dominant place type.
```
shifucun, line 148: nit: All place types?

proto/service/mixer.proto
```diff
@@ -611,6 +611,14 @@ service Mixer {
     };
   }
 
+  // Recognize non-place entities from a NL query.
+  rpc RecognizeEntities(RecognizePlacesRequest)
```
shifucun, line 615: Since these two apis share the same request and response, why not making them the same api?

proto/recon.proto
chejennifer: This will not just be place types, it may also be types like "Gene" or "Drug", etc. But if there is a "dominant place type", then that will still be at the front of the list

internal/server/recon/golden/recognize_entities/result.json
```diff
@@ -0,0 +1,101 @@
+{
+  "queryItems": {
+    "tell me about Benzodiazepine, derivatives and their use in California": {
+      "items": [
+        {
+          "span": "tell"
+        },
+        {
+          "span": "me",
+          "places": [
```
shifucun, line 10: should here use "entities" instead of "places"?

chejennifer: updated the proto message

proto/service/mixer.proto
chejennifer: Following offline discussions, keeping the apis different, but added a new request and response

shifucun (approved):
thanks for the update!

internal/server/recon/recognize.go
```diff
@@ -73,6 +81,189 @@ func RecognizePlaces(
 	return resp, nil
 }
 
+// RecognizePlaces implements API for Mixer.RecognizePlaces.
```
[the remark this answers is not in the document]
chejennifer: this was updated in the latest commit!

chejennifer:
thanks for the reviews!

PR #1348, opened 2024-04-11
Rank observations from SQL database (custom DC) over that from Bigtable (base DC)

shifucun:


internal/server/v2/observation/golden/direct/latest.json
```diff
@@ -409,6 +409,18 @@
         "dummy": {},
         "geoId/06": {
           "ordered_facets": [
+            {
+              "facet_id": "custom",
```
keyurva, line 413: A few questions:
- Without this change, would the custom facet be included after the other facets?
- With this change, since the custom facet is the first one in the list, does the UI use that one when showing charts?
- Does it make sense to support a mode where we don't return main DC data at all if we find data in custom DC for the request? This could help with UN WHO testing and possibly even for prod UN ILO when they are in the middle of moving data from custom to main DC.

keyurva:
Very cool!

internal/server/v2/observation/contained_in.go
```diff
@@ -277,7 +277,8 @@ func FetchContainedIn(
 			}
 			sqlResult = trimDirectResp(directResp)
 		}
-		result = merger.MergeObservation(result, sqlResult)
+		// Prefer SQLite data over BT data, so put sqlResult first.
```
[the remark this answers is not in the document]
shifucun: Done

internal/server/v2/observation/direct.go
```diff
@@ -103,7 +103,8 @@ func FetchDirect(
 	if err != nil {
 		return nil, err
 	}
-	return merger.MergeObservation(o1, o2), nil
+	// Prefer SQLite data over BT data, so put sqlObservation first.
```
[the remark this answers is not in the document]
shifucun: Done

internal/server/v2/observation/golden/direct/latest.json
shifucun: 1. Yes
2. Yes, for APIs that only return 1 (the top) facet, the custom data will be used
3. I think that requires more thought and I prefer source filtering than just block base DC data. The behavior is is very subtle.

keyurva: On # 2 - what about APIs that return more than 1?

keyurva (approved):

internal/server/v2/observation/golden/direct/latest.json
shifucun: the observation api either return the first facet or all facets

keyurva: I meant what's the UI behavior when all facets are returned?

shifucun:
thanks for review!

PR #1349, opened 2024-04-11
[recognize entities] handle common words

chejennifer:
There are some reconName values that are very common words like "me", "to", "a". To handle these, added a map of reconNam value to the allowed types of the entities recognized & only recognize a name if the name is preceded by its type in the query.

For example, there is a gene called "Me". The query "Tell me about the gene me" should not recognize the first "me", but should recognize the "gene me".


internal/server/recon/recon_name_to_types.json
```diff
@@ -0,0 +1,153 @@
+{
```
pradh, line 1: Could you add a README on how this is generated?

internal/server/recon/recognize.go
```diff
@@ -81,6 +86,21 @@ func RecognizePlaces(
...
+	_, filename, _, _ := runtime.Caller(0)
+	bytes, err := os.ReadFile(path.Join(path.Dir(filename), reconName2RequiredTypes))
+	if err == nil {
+		err = json.Unmarshal(bytes, &typeRequiredEntitiesMap)
+		if err != nil {
+			return map[string][]string{}
+		}
+	}
+	return typeRequiredEntitiesMap
+}
```
pradh, line 102: We don't want to read this for every request.   Instead this file should be part of: https://github.com/datacommonsorg/mixer/blob/6e695ef4deb8dcd3f15e9086ffa57195277b794f/internal/store/files/recog_place_store.go#L81

pradh (commented):
Nice.  Should we also add the script to generate this under tools? (separate PR is fine)

internal/server/recon/recon_name_to_types.json
chejennifer: done

internal/server/recon/recognize.go
chejennifer: done

chejennifer:
> Nice. Should we also add the script to generate this under tools? (separate PR is fine)

Sure, I will do that in a separate PR

internal/server/recon/recognize.go
```diff
@@ -91,8 +91,10 @@ func RecognizeEntities(
 	resp := &pb.RecognizeEntitiesResponse{
 		QueryItems: map[string]*pb.RecognizeEntitiesResponse_Items{},
 	}
+
 	// TODO: parallelize queries
 	for _, query := range in.GetQueries() {
+		query = strings.ToLower(query)
```
pradh, line 97: Why is this needed if reconName is going to match smaller case?

internal/store/files/README.md
```diff
@@ -21,3 +21,24 @@ WHERE
   predicate = 'fips52AlphaCode'
   AND subject_id LIKE 'geoId/%'
 ```
+
+## recon_name_to_types.json generation
```
pradh, line 25: nit: Intro line on what this file is, and why we need it?

internal/store/files/README.md
```diff
@@ -21,3 +21,24 @@ WHERE
...
+SELECT
+  t1.object_value as name,
+  t2.object_id
+FROM `datcom-store.dc_kg_latest.Triple` as t1
+JOIN `datcom-store.dc_kg_latest.Triple` as t2
+ON t1.subject_id = t2.subject_id 
+WHERE 
+  t1.predicate = 'name' and
+  t2.predicate = 'typeOf' and
+  t2.object_id in ("VirusGenusEnum", "VirusIsolate", "Virus", "Species", "BiologicalSpecimen", "GeneticVariant", "Gene", "Disease", "ICD10Section", "ICD10Code", "MeSHDescriptor", "Drug", "AnatomicalTherapeuticChemicalCode", "MeSHSupplementaryRecord")
```
pradh, line 39: -- This list should match the list [here](http://canonical-path-to-the-list-in-google3?)

internal/store/files/README.md
```diff
@@ -21,3 +21,24 @@ WHERE
...
+```
+
+2. Put the results of (1) through a script that takes the names and uses spacy to get its "part of sentence" classification ([e.g., noun, pronoun, etc](https://melaniewalsh.github.io/Intro-Cultural-Analytics/05-Text-Analysis/13-POS-Keywords.html#spacy-part-of-speech-tagging)). The script then flags names that do not have any punctuation, nouns, proper nouns or numbers (potentially a common word).
```
pradh, line 42: TODO: Link to script after check-in :)

internal/server/recon/recognize.go
```diff
@@ -116,7 +118,17 @@ func RecognizeEntities(
 				entities = append(entities, &pb.RecognizeEntitiesResponse_Entity{Dcid: id})
 			}
 			for span := range id2spans[entity.GetInId()] {
-				span2item[span] = &pb.RecognizeEntitiesResponse_Item{Span: span, Entities: entities}
+				if types, ok := store.RecogPlaceStore.ReconNameToRequiredTypes[entity.GetInId()]; ok {
+					// If the recognized name has required types that it can resolve for,
+					// add the type in front of the span so only spans that include the
+					// type will resolve.
+					for _, t := range types {
+						span_with_type := strings.ToLower(t) + " " + span
```
pradh, line 126: I wonder if we want alternate names (or space delimited name) for `t`?  Like BiologicalSpecimen should really be "biological specimen", rather than "biologicalspecimen".  Or "Mesh Descriptor".

Have a TODO?

internal/store/files/recog_place_store.go
```diff
@@ -75,6 +78,9 @@ type RecogPlaceStore struct {
 	DcidToNames map[string][]string
 	// Adjectival names with suffix.
 	AdjectivalNamesWithSuffix map[string]bool
+	// Map of reconName to list of possible types for reconNames that must be
```
pradh, line 81: nit: As a reader its hard to know why we have a special reconName to list of types just from the name.  Perhaps rename it to `CommonWordReconNameToTypes` or something that gives a clue?

pradh: We should probably also have comment in the code (like the stuff in the PR description) explaining this.

pradh (approved):
Thanks for the edits!

internal/server/recon/recognize.go
chejennifer: this makes things easier downstream because if there's a type, we want to match `type + " " + span` to parts in the query & capitalization of type should not matter

internal/store/files/README.md
chejennifer: done

internal/store/files/README.md
chejennifer: done

internal/store/files/README.md
chejennifer: done

internal/server/recon/recognize.go
chejennifer: added alternate names because you're right, i think the user will at least space delimit the type

internal/store/files/recog_place_store.go
chejennifer: done

internal/store/files/recog_place_store.go
```diff
@@ -63,6 +63,18 @@ var (
 		"state",
 		"states",
 	}
+	// Map of entity types to their human string.
```
pradh, line 66: TODO:  move this to a file?

Another thought:  should we just put the human strings directly in "CommonWordReconNameTo..." map?  Is there a reason the value needs to be a type?

internal/server/recon/recognize.go
pradh: Makes sense.

internal/store/files/recog_place_store.go
chejennifer: I don't think there's actually a reason the value needs to be a type, updated to just put the human strings directly in the map

pradh (approved):
Nice!  Thanks for all the edits!

chejennifer:
thanks for the reviews!

PR #1350, opened 2024-04-18
[recognize entities] cleanup

chejennifer:
- remove entities from place recognition
- update README with link to script for getting the recon names that are common words
- add more recon names that are common words


pradh (approved):
Nice!

PR #1351, opened 2024-04-19
[recognize entities] fix skipping entities beyond a certain index in the query

chejennifer:
There was a bug in limiting the n-grams used to recognize entities where beyond a certain number of words in the query, the rest of the query was not being used in recognition

### DIFF
--- a/internal/server/recon/recognize.go
+++ b/internal/server/recon/recognize.go
@@ -257,9 +257,9 @@ func getId2Span(query string) map[string]map[string]struct{} {
 	spanTokens := strings.Split(query, " ")
 	for i := range spanTokens {
 		span := ""
-		maxNGramLength := int(math.Min(float64(len(spanTokens)), reconNGramLimit+1))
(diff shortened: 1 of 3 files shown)


internal/server/recon/recognize.go
```diff
@@ -257,9 +257,9 @@ func getId2Span(query string) map[string]map[string]struct{} {
 	spanTokens := strings.Split(query, " ")
 	for i := range spanTokens {
 		span := ""
-		maxNGramLength := int(math.Min(float64(len(spanTokens)), reconNGramLimit+1))
+		maxNGramIdx := int(math.Min(float64(len(spanTokens)), float64(reconNGramLimit+i)))
```
pradh, line 260: Comment what this is doing?

pradh (approved):

internal/server/recon/recognize.go
chejennifer: done

PR #1352, opened 2024-04-20
Update v2 response merger to always rank local response higher

shifucun:
Also updated the logic for all the merger.

Before the change, the merge would not honor the "main" response but would blend "main" and "aux" response. With the change, we make sure all the merged response rank the "main" response result higher.


internal/merger/merger.go
```diff
@@ -95,205 +89,159 @@ func MergeNode(local, remote *pbv2.NodeResponse) (*pbv2.NodeResponse, error) {
...
+	auxData := aux.GetData()
+	for dcid, linkedGraph := range auxData {
+		if mainData == nil {
+			mainData = map[string]*pbv2.LinkedGraph{}
+		}
+		if _, ok := mainData[dcid]; !ok {
+			mainData[dcid] = linkedGraph
+			continue
+		}
+		for prop, nodes := range linkedGraph.GetArcs() {
```
keyurva, line 107: nit: Any chance we create internal methods for some of this logic? It will help make this method shorter and easier to follow.

keyurva (commented):
Very cool!

internal/merger/merger.go
```diff
@@ -26,65 +30,55 @@ import (
...
-
-	res := &pbv2.ResolveResponse{}
-	for node, idToCandidate := range store {
-		candidates := []*pbv2.ResolveResponse_Entity_Candidate{}
-		for _, candidate := range idToCandidate {
-			candidates = append(candidates, candidate)
+	// Add aux entities that are not in main
+	for _, e := range aux.Entities {
+		if _, ok := auxStore[e.Node]; ok {
+			main.Entities = append(main.Entities, e)
```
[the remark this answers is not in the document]
shifucun: This is to handle the node level addition from Aux source. While line 55-57 is to handle handle candidates (inside node) level merge for the same node.

internal/merger/merger.go
shifucun: Good point. Made a helper function for this.

internal/merger/merger.go
```diff
@@ -26,65 +30,98 @@ import (
...
 
-	// Sort to make result deterministic.
-	sort.Slice(res.Entities, func(i, j int) bool {
-		return res.Entities[i].Node < res.Entities[j].Node
-	})
-
-	return res
+func mergeLinkedGraph(
+	mainData, auxData map[string]*pbv2.LinkedGraph,
+) map[string]*pbv2.LinkedGraph {
```
pradh, line 73: The triples API seems involved enough that we should probably have some focussed e2e tests for local vs. remote, in different cases (node-level, prop-level, PV-level, ...) ?  Do we have those? If not, have a TODO?

pradh (approved):
Nice!

internal/merger/merger.go
shifucun: There is a unit test for this function that should be good to guard but that does need more testing cases with real data. I will leave a todo and update the unittest.

keyurva (approved):

shifucun:
thanks for review!

PR #1353, opened 2024-04-22
Fix empty data check when merging v2 node responses

shifucun:
When no data is present, the empty data is not nil but rather empty arcs and nodes.

### DIFF
--- a/internal/merger/merger_test.go
+++ b/internal/merger/merger_test.go
@@ -110,6 +110,46 @@ func TestMergeNode(t *testing.T) {
 				Data: map[string]*pbv2.LinkedGraph{
 					"dcid1": {
 						Arcs: map[string]*pbv2.Nodes{
+							"prop1": nil,
+						},
+					},
+				},
+			},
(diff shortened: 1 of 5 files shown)


internal/merger/merger_test.go
```diff
@@ -226,7 +226,15 @@ func TestMergeNode(t *testing.T) {
 			},
 		},
 		{
-			&pbv2.NodeResponse{},
+			&pbv2.NodeResponse{
+				Data: map[string]*pbv2.LinkedGraph{
+					"dcid1": {
+						Arcs: map[string]*pbv2.Nodes{
+							"prop1.2": {},
```
keyurva, line 233: Perhaps also add a test case where the value is `nil`?

```
Arcs: map[string]*pbv2.Nodes{
  "prop1.2": nil,
}
```

keyurva (approved):
Thanks for the fix.

internal/merger/merger_test.go
shifucun: Done

PR #1354, opened 2024-04-25
Parallelize FetchChildPlaces.

keyurva:


internal/server/v2/observation/contained_in.go
```diff
@@ -66,6 +67,43 @@ func trimDirectResp(resp *pbv2.ObservationResponse) *pbv2.ObservationResponse {
...
+)
+
+func fetchRemoteWrapper(
+	metadata *resource.Metadata,
+	httpClient *http.Client,
+	apiPath string,
+	remoteReq *pbv2.NodeRequest,
+) (*pbv2.NodeResponse, error) {
+	remoteResp := &pbv2.NodeResponse{}
+	return remoteResp, util.FetchRemote(metadata, httpClient, apiPath, remoteReq, remoteResp)
```
shifucun, line 83: I feel it's better to make this explicit. Will be helpful for debugging if error occurs.

err := util.FetchRemote(metadata, httpClient, apiPath, remoteReq, remoteResp)
if err != nil {
    return nil, err
}
return remoteResp, nil

internal/server/v2/observation/contained_in.go
```diff
@@ -66,6 +67,43 @@ func trimDirectResp(resp *pbv2.ObservationResponse) *pbv2.ObservationResponse {
...
+}
+
+func storeFetchChildPlaces(
+	ctx context.Context,
+	store *store.Store,
+	ancestor, childType string,
+) (map[string][]string, error) {
+	return getPlacesIn(
```
shifucun, line 91: nit: no need to wrap lines here.

internal/server/v2/observation/contained_in.go
```diff
@@ -66,6 +67,43 @@ func trimDirectResp(resp *pbv2.ObservationResponse) *pbv2.ObservationResponse {
...
+func remoteMixerFetchChildPlaces(
+	metadata *resource.Metadata,
+	httpClient *http.Client,
+	ancestor, childType string,
+) (*pbv2.NodeResponse, error) {
+	remoteReq := &pbv2.NodeRequest{
+			Nodes:    []string{ancestor},
+			Property: fmt.Sprintf("<-containedInPlace+{typeOf:%s}", childType),
+		}
+		return fetchRemote(metadata, httpClient, "/v2/node", remoteReq)
```
shifucun, line 104: The indent of line 101-104 is off?

internal/server/v2/observation/contained_in.go
```diff
@@ -74,24 +112,46 @@ func FetchChildPlaces(
...
+	if remoteMixer != "" {
+		errGroup.Go(func() error {
+		remoteMixerResponse, err := remoteMixerFetchChildPlaces(metadata, httpClient, ancestor, childType)
+		if err != nil {
+			return err
+		}
+		remoteMixerResponseChan <- remoteMixerResponse
+		return nil
+	})
+	} else {
```
shifucun, line 138: i think there is some indent issues. Maybe have a golang plugin to fix them.

internal/server/v2/observation/contained_in_test.go
```diff
@@ -0,0 +1,108 @@
...
+		childType           string
+		storeResponse       map[string][]string
+		remoteMixerResponse *pbv2.NodeResponse
+		want                []string
+	}{{
+		desc: "store only",
+		remoteMixer: "",
+		ancestor: "a1",
+		childType: "CT1",
+		storeResponse: map[string][]string{"a1": {"c1", "c2"}},
```
shifucun, line 54: this test case does not represent current behavior since remoteResponse is always empty struct instead of nil now?

internal/server/v2/observation/contained_in.go
keyurva: Sorry - my format on save was not on. All formatting issues should be fixed now.

We should add a format checker on PRs like in the other repos.

internal/server/v2/observation/contained_in.go
keyurva: Done.

internal/server/v2/observation/contained_in_test.go
keyurva: Since `remoteMixerResponse` is a pointer, it's zero value should be `nil` so I think it represents current behavior?

internal/server/v2/observation/contained_in.go
keyurva: Done.

internal/server/v2/observation/contained_in.go
keyurva: Done.

internal/server/v2/observation/contained_in_test.go
shifucun: okay, now the nil pointer looks good.

shifucun (approved):
thanks for adding the mock testing. Very nice change!

keyurva:
Thanks for the review!

PR #1355, opened 2024-04-29
Make custom dc db name configurable.

keyurva:
* DB name can be configured via a `DB_NAME` env var (defaults to `datacommons`) similar to the [loader](https://github.com/datacommonsorg/import/blob/0162324b5c72a3969bf1d226e0ad027e157c3432/simple/README.md?plain=1#L64).
* @kmoscoe - let's document this variable.

### DIFF
--- a/internal/sqldb/cloudsql/cloudsql.go
+++ b/internal/sqldb/cloudsql/cloudsql.go
@@ -27,7 +27,7 @@ import (
 )
 
 const (
-	dbName = "datacommons"
+	defaultDbName = "datacommons"
 )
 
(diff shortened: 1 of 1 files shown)


kmoscoe (commented):
Can you please add the option to custom_dc/cloudsql_env.list too? Thanks.

keyurva:
> Can you please add the option to custom_dc/cloudsql_env.list too? Thanks.

That file is in a separate repo so will need to be changed in a separate PR.

shifucun (approved):

PR #1362, opened 2024-05-16
Cache StatisticalCalculation formulas

n-h-diaz:
Tested SQL and BT (using schema_2024_05_16_12_21_31)


test/triples.csv
```diff
@@ -90,4 +90,6 @@ geoId/44,typeOf,State,
 geoId/45,typeOf,State,
 geoId/46,typeOf,State,
 geoId/47,typeOf,State,
-geoId/48,typeOf,State,
```
shifucun, line 93: why removing geoId/48,typeOf,State?

cmd/main.go
```diff
@@ -85,6 +85,8 @@ var (
 	// Serve live profiles of the process (CPU, memory, etc.) over HTTP on this port
 	httpProfilePort   = flag.Int("httpprof_port", 0, "Port to serve HTTP profiles from")
 	foldRemoteRootSvg = flag.Bool("fold_remote_root_svg", false, "Whether to fold root SVG from remote mixer")
+	// Cache SV -> inputPropertyExpressions for StatisticalCalculations
```
shifucun, line 88: a bit more detail in the comment on "->"?

cmd/main.go
```diff
@@ -85,6 +85,8 @@ var (
...
+	cacheSVFormula = flag.Bool("cache_sv_formula", true, "Whether to cache SV -> inputPropertyExpresions for StatisticalCaclulations.")
```
shifucun, line 89: can we make it default to be false? Then set it true in deployment. After PR merge, can test custom DC and make sure it works.

Add a TODO tot change this back to "true" after this is fully released?

internal/server/cache/cache.go
```diff
@@ -54,6 +55,8 @@ type Cache struct {
 	sqlProvenances map[string]*pb.Facet
 	// SQL database entity, variable existence pairs
 	sqlExistenceMap map[util.EntityVariable]struct{}
+	// SV -> inputPropertyExpressions for StatisticalCalculations.
```
shifucun, line 58: maybe use more verbose description for "->"? Also the value is a list of strings but the comment seems to be an object, clarify?

internal/server/statvar/fetcher/formula_fetcher.go
```diff
@@ -0,0 +1,149 @@
...
+	pbv2 "github.com/datacommonsorg/mixer/internal/proto/v2"
+	"github.com/datacommonsorg/mixer/internal/server/resource"
+	v1pv "github.com/datacommonsorg/mixer/internal/server/v1/propertyvalues"
+	"github.com/datacommonsorg/mixer/internal/store"
+	"github.com/datacommonsorg/mixer/internal/util"
+	"golang.org/x/sync/errgroup"
+)
+
+// V2Node for Remote Mixer.
+func FetchRemoteNode(
```
shifucun, line 31: No need to use errGroup here, since there are no other concurrent go routines. But you do want to use errGroup in FetchFormulas to fetch local and remote concurrently.

internal/server/statvar/fetcher/formula_fetcher.go
```diff
@@ -0,0 +1,149 @@
...
+			}
+		}
+		for _, nodes := range properties["inputPropertyExpression"] {
+			for _, node := range nodes {
+				calculationToFormulas[dcid] = append(calculationToFormulas[dcid], node.Value)
+			}
+		}
+	}
+	// Fetch for Remote Mixer.
+	if metadata.RemoteMixerDomain != "" {
```
shifucun, line 106: do the remote fetch concurrently with the local fetch

internal/server/statvar/fetcher/formula_fetcher.go
```diff
@@ -0,0 +1,149 @@
...
+		req := &pbv2.NodeRequest{
+			Nodes:    []string{"StatisticalCalculation"},
+			Property: "<-typeOf",
+		}
+		statisticalCalculationResp, err := FetchRemoteNode(ctx, metadata, req)
+		if err != nil {
+			return nil, err
+		}
+		statisticalCalculations := []string{}
+		for _, node := range statisticalCalculationResp.Data["StatisticalCalculation"].Arcs["typeOf"].Nodes {
```
shifucun, line 116: should have test case to check empty response and foobar input

internal/server/v2/observation/golden/contained_in_2015_test.go
```diff
@@ -120,7 +120,7 @@ func TestFetchContainIn(t *testing.T) {
 	}
 	if err := test.TestDriver(
 		"FetchContainIn",
-		&test.TestOption{UseSQLite: true},
+		&test.TestOption{UseSQLite: true, CacheSVFormula: true},
```
shifucun, line 123: you don't need to enable this if this test does not test the formula part? And same for the other tests.

I expect to see golden file changes for this PR but seems no json have changed?

internal/server/statvar/fetcher/formula_fetcher.go
```diff
@@ -0,0 +1,145 @@
...
+		return nil
+	})
+	// Fetch for Remote Mixer.
+	if metadata.RemoteMixerDomain != "" {
+		errGroup.Go(func() error {
+			req := &pbv2.NodeRequest{
+				Nodes:    []string{"StatisticalCalculation"},
+				Property: "<-typeOf",
+			}
+			statisticalCalculationResp := &pbv2.NodeResponse{}
```
shifucun, line 82: nit: golang uses short var names, also i found it easy to read :)

Given the very long lines in this function, maybe use naming like statCalResp, statCal, outputProps, inputProps, outputExp, inputExp etc?

test/triples.csv
n-h-diaz: Sorry this was my mistake, I added this back

cmd/main.go
n-h-diaz: Done

internal/server/cache/cache.go
n-h-diaz: Updated description

internal/server/statvar/fetcher/formula_fetcher.go
n-h-diaz: Oh thanks that makes sense! Done

internal/server/statvar/fetcher/formula_fetcher.go
n-h-diaz: Done

internal/server/statvar/fetcher/formula_fetcher.go
n-h-diaz: I wasn't entirely sure what you meant here, but I added a check for empty response

internal/server/v2/observation/golden/contained_in_2015_test.go
n-h-diaz: I removed the option for now (will add later when I actually use it). Currently this just created the cache, but it's not referenced anywhere, so this change shouldn't affect any existing behavior

internal/server/statvar/fetcher/formula_fetcher.go
n-h-diaz: Good idea thanks! Done

cmd/main.go
n-h-diaz: Good idea, done

shifucun (approved):
thanks for the update! very nice!

internal/server/statvar/fetcher/formula_fetcher.go
```diff
@@ -0,0 +1,151 @@
...
+	pb "github.com/datacommonsorg/mixer/internal/proto"
+	pbv2 "github.com/datacommonsorg/mixer/internal/proto/v2"
+	"github.com/datacommonsorg/mixer/internal/server/resource"
+	v1pv "github.com/datacommonsorg/mixer/internal/server/v1/propertyvalues"
+	"github.com/datacommonsorg/mixer/internal/store"
+	"github.com/datacommonsorg/mixer/internal/util"
+	"golang.org/x/sync/errgroup"
+)
+
+// FetchFormulas fetches StatisticalCalculations from storage and returns a map of SV dcids to a list of inputPropertyExpressions.
```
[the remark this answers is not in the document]
n-h-diaz: Done

n-h-diaz:
> thanks for the update! very nice!

Thanks for reviewing! :)

internal/server/statvar/fetcher/formula_fetcher.go
```diff
@@ -0,0 +1,145 @@
...
+func FetchFormulas(
```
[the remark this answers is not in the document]
n-h-diaz: That makes sense, thanks! I can add a TODO and will split up fetch and merge in a a follow up PR + add more tests

PR #1364, opened 2024-05-28
Add basic support for calculations

n-h-diaz:
This also splits up the fetching and merging for the formulas (follow up to https://github.com/datacommonsorg/mixer/pull/1362) and adds basic tests.

This currently does not support matching facets, which will be a follow up. 

Currently, this independently calculates each missing variable/entity pair, rather than combining the entities into one DerivedSeries request, and then updates the merging to accept an array of responses. This was to avoid computing the calculation if some of the entities already have data, however if it's preferred to only do one DerivedSeries per formula I can update this.


internal/merger/merger.go
```diff
@@ -213,43 +214,54 @@ func MergeEvent(main, aux *pbv2.EventResponse) *pbv2.EventResponse {
 	return main
 }
 
-// MergeObservation merges two V2 observation responses.
-func MergeObservation(main, aux *pbv2.ObservationResponse) *pbv2.ObservationResponse {
+// MergeObservation merges multiple V2 observation responses.
+func MergeObservation(main *pbv2.ObservationResponse, aux []*pbv2.ObservationResponse) *pbv2.ObservationResponse {
```
shifucun, line 218: I wonder if using https://gobyexample.com/variadic-functions is cleaner? And saying the args are ranked by preference?

internal/server/statvar/fetcher/formula_fetcher.go
```diff
@@ -116,37 +116,5 @@ func FetchFormulas(
...
-				}
-			}
-		}
-	}
-	// Sort for determinism.
-	for _, formulas := range result {
-		sort.Strings(formulas)
-	}
-	return result, nil
+	return merger.MergeFormulas(localResp, remoteResp)
```
shifucun, line 119: if you convert localResp to pbv2.NodeResponse, then can just use merger.MergeNode. Feels like the mergeFormula is just a specialized impl for MergeNode?

internal/merger/merger.go
n-h-diaz: Good idea, thanks! Added a wrapper recursive method

internal/merger/merger.go
```diff
@@ -349,3 +361,44 @@ func toStatVarSummaryMap(in []*pbv1.VariableInfoResponse) map[string]*proto.Stat
 	}
 	return out
 }
+
+// Create map of SV to inputPropertyExpressions by merging local and remote responses.
+func MergeFormulas(localResp map[string]map[string]map[string][]*pb.EntityInfo, remoteResp *pbv2.NodeResponse) (map[string][]string, error) {
```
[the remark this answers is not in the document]
n-h-diaz: I ended up removing this function based on your comments below

internal/merger/merger.go
```diff
@@ -349,3 +361,44 @@ func toStatVarSummaryMap(in []*pbv1.VariableInfoResponse) map[string]*proto.Stat
...
+	result := map[string][]string{}
+	localResult := map[string]map[string]bool{}
+	for _, props := range localResp {
+		for _, outputProps := range props["outputProperty"] {
+			for _, outputNode := range outputProps {
+				for _, inputProps := range props["inputPropertyExpression"] {
```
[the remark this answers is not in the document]
n-h-diaz: I removed this based on your comments below (using NodeResponse instead)

internal/server/statvar/fetcher/formula_fetcher.go
n-h-diaz: Thanks! This made things a lot cleaner

internal/server/handler_v2.go
```diff
@@ -242,5 +243,8 @@ func (s *Server) V2Observation(
 	localResp, remoteResp := <-localRespChan, <-remoteRespChan
 	// The order of argument matters, localResp is prefered and will be put first
 	// in the merged result.
-	return merger.MergeObservation(localResp, remoteResp), nil
+	mergedResp := merger.MergeObservation(localResp, remoteResp)
+	calculatedResps := v2observation.CalculateObservationResponses(ctx, s.store, mergedResp, s.cachedata.Load())
+	// mergedResp is preferred over any calculated response.
+	return merger.MergeMultipleObservations(append([]*pbv2.ObservationResponse{mergedResp}, calculatedResps...)...), nil
```
shifucun, line 249: Is it possible to write as `merger.MergeMultipleObservations(mergedResp, calculatedResps...)`?

internal/server/v2/observation/calculation.go
```diff
@@ -0,0 +1,66 @@
...
+func CalculateObservationResponses(
+	ctx context.Context,
+	store *store.Store,
+	inputResp *pbv2.ObservationResponse,
+	cachedata *cache.Cache,
+) []*pbv2.ObservationResponse {
+	calculatedResponses := []*pbv2.ObservationResponse{}
+	for variable, variableObservation := range inputResp.ByVariable {
+		formulas, ok := cachedata.SVFormula()[variable]
+		if ok {
```
shifucun, line 38: nit: if !ok { continue } to reduce one more indent

internal/server/v2/observation/calculation.go
```diff
@@ -0,0 +1,66 @@
...
+			for entity, entityObservation := range variableObservation.ByEntity {
+				if len(entityObservation.OrderedFacets) == 0 {
```
shifucun, line 40: save here to check !=0 and continue to remove one more indent.

internal/server/v2/observation/calculation.go
```diff
@@ -0,0 +1,66 @@
...
+							store,
+							formula,
+							[]string{entity},
+						)
+						// Missing input data.
+						if err != nil {
+							continue
+						}
+						// Successful calculation.
+						if len(derivedSeries.ByVariable[formula].ByEntity[entity].OrderedFacets) > 0 {
```
shifucun, line 54: nit: `bv := derivedSeries.ByVariable` and use this short name in these few lines.

internal/server/v2/observation/calculation.go
```diff
@@ -0,0 +1,66 @@
...
+							derivedSeries.ByVariable[variable] = derivedSeries.ByVariable[formula]
+							delete(derivedSeries.ByVariable, formula)
```
shifucun, line 56: comment on why delete here?

internal/server/v2/observation/calculation.go
```diff
@@ -0,0 +1,66 @@
+// Copyright 2023 Google LLC
```
shifucun, line 1: nit: 2023

shifucun (commented):
ver nice!!

internal/server/handler_v2.go
n-h-diaz: It says that there are too many arguments - is there another way to write this? For now I just separated out the inner append to make it more readable

internal/server/v2/observation/calculation.go
n-h-diaz: Done

internal/server/v2/observation/calculation.go
n-h-diaz: Done

internal/server/v2/observation/calculation.go
n-h-diaz: Done

internal/server/v2/observation/calculation.go
n-h-diaz: Done

internal/server/v2/observation/calculation.go
n-h-diaz: Done

PR #1367, opened 2024-05-31
Fetch SVG cache data from sql db.

keyurva:
* Fetch custom dc svgs from cache and fallback to querying sql table if cached entry is not found.


shifucun (commented):
One high level naming comment. Mixer has its own cache (a real cache) for svg, sv search index etc. Also as we move to dynamic computation, a new caching (real cache too) will be added. Given these, i would say not using "cache" here in file, function and variable names.

What do you think of "computed" or something along that line? This can contrast with the raw triples data.

internal/server/statvar/fetcher/svg_fetcher.go
```diff
@@ -100,9 +102,56 @@ func FetchAllSVG(
...
+		return nil, err
+	}
+
+	return &svgs, nil
+}
+
+func fetchSQLTableSVGs(sqlClient *sql.DB) (map[string]*pb.StatVarGroupNode, error) {
+	result := map[string]*pb.StatVarGroupNode{}
+	// Query for all the stat var group node
+	query :=
```
shifucun, line 153: nit: Remove all indent for query string

internal/sqldb/sqlquery/cache.go
```diff
@@ -0,0 +1,68 @@
...
+
+const (
+	SVGCacheKey = "c/cache/svgs"
+)
+
+// GetCacheData gets cache data for the specified key from the sql DB.
+// If not found, returns false.
+// If found, unmarshals the value into the specified proto and returns true.
+func GetCacheData(sqlClient *sql.DB, key string, out protoreflect.ProtoMessage) (bool, error) {
+	query :=
```
shifucun, line 34: ni: remove all indent for sql string

internal/sqldb/sqlquery/cache.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+const (
+	SVGCacheKey = "c/cache/svgs"
+)
+
+// GetCacheData gets cache data for the specified key from the sql DB.
+// If not found, returns false.
+// If found, unmarshals the value into the specified proto and returns true.
+func GetCacheData(sqlClient *sql.DB, key string, out protoreflect.ProtoMessage) (bool, error) {
```
shifucun, line 33: no need for the bool return value. In this pattern, "nil" error signals a success

internal/server/statvar/fetcher/svg_fetcher.go
```diff
@@ -100,9 +102,56 @@ func FetchAllSVG(
...
+	svgCache, err := fetchSQLCacheSVGs(sqlClient)
+	if err != nil {
+		return map[string]*pb.StatVarGroupNode{}, err
+	}
+	if svgCache != nil {
+		// Cached data found => return it.
+		return svgCache.StatVarGroups, nil
+	}
+
+	// Query sql table.
```
shifucun, line 135: Add a TOOD to deprecate fetching from table?

internal/sqldb/sqlquery/cache.go
keyurva: Here we need to distinguish between an error and a value not found. If the out message was just a standard return type, a `nil` there would suffice. But since it is passed as a method parameter instead, the found / not found `bool` is the way to determine existence.

One option I considered was to return the `sql.ErrNoRows` error if a value was not found. But that felt like breaking encapsulation so went with a `bool` instead.

shifucun: Is it possible to set out to nil if no data is found?

keyurva: That won't work. For out params, we can only update the value, not change the reference.

internal/server/statvar/fetcher/svg_fetcher.go
keyurva: Done. Added TODO to the method definition.

internal/server/statvar/fetcher/svg_fetcher.go
keyurva: Done.

internal/sqldb/sqlquery/cache.go
keyurva: Done.

keyurva:
Discussed offline to use a new `key_value_store` table vs using the `triples` table for KV lookups.

shifucun (approved):
thanks for the update!

keyurva:
Thanks for the review!

PR #1368, opened 2024-05-31
Add county names without "county" carefully

pradh:
Fix for b/343267908

The list was generated carefully using [this notebook](https://colab.research.google.com/github/pradh/mixer/blob/countufix2/internal/store/files/Safely_adding_names_without_county_to_NL_Place_recognition.ipynb).  Please sanity check the logic there!

This needs to be carefully done because removing "county" from "San Francisco County" for example can suppress SF city, which is not good.

Safety test I use:  if removing "county" from the name can suppress a "sizeable" place within the scope of the State or US, then  "county" cannot be safely removed!  And "sizeable" place is defined as one with >10K population.

This resulted in 2437 safe counties and 566 unsafe counties.  The list of both, and which places are suppressed by the unsafe counties are all in the notebook.


internal/server/v2/observation/golden/calculation/statistical_calculation.json
```diff
@@ -38,209 +38,209 @@
...
-              }
-            ]
-          }
+          "ordered_facets": [
+            {
+              "observations": [
+                {
+                  "date": "1975",
+                  "value": 128821
+                },
```
pradh, line 47: @n-h-diaz any idea why the diffs here?

n-h-diaz: I'm not sure - the data looks the same, so maybe it fixed the formatting?

chejennifer (approved):
thank you for the change! the logic in the notebook looks sound to me!

pradh:
Thanks for the review!

PR #1372, opened 2024-05-31
Add more tests and enable formulas on autopush

n-h-diaz:


shifucun (approved):
good to have more tests!

PR #1373, opened 2024-06-01
Add tests for calculated obsCollection

n-h-diaz:
This change also adds empty place responses for all the childPlaces, since the child places are needed by the calculator

This mirrors the response for other ObservationResponses where requested places without data are empty, but let me know if you'd like me to remove these


internal/server/v2/observation/contained_in.go
```diff
@@ -186,6 +186,11 @@ func FetchContainedIn(
 	var err error
 	if store.BtGroup != nil {
 		readCollectionCache := false
+		childPlaces, err = FetchChildPlaces(
+			ctx, store, metadata, httpClient, remoteMixer, ancestor, childType)
+		if err != nil {
+			return nil, err
+		}
```
shifucun, line 193: We don't want to always fetch this since the obsCollection collection could already exists from BT cache.

internal/server/v2/observation/contained_in.go
```diff
@@ -210,6 +215,10 @@ func FetchContainedIn(
 					obsByEntity := result.ByVariable[variable].ByEntity
 					data, ok := btData[variable]
 					if !ok || data == nil {
+						// If data is missing, attach empty child places.
+						for _, place := range childPlaces {
+							obsByEntity[place] = &pbv2.EntityObservation{}
+						}
```
shifucun, line 221: feels like this should be part of Flume logic. Right now Flume does not return place with no data. This is a behavior discrepancy. 

I actually think it makes sense to not have an entry for a child place if there is no data in all scenarios.

shifucun (commented):
How is child places needed by calculator?

n-h-diaz:
> How is child places needed by calculator?

The calculator uses a variable/entity to try and fetch the data for for the calculated response. Currently, for BT it will return an empty response with no places, so the calculator won't detect there's any missing variable/entity to compute

internal/server/v2/observation/contained_in.go
n-h-diaz: Should I add this to the flume instead?

shifucun: I still don't think we should add places with no data. Imagine fetching cities in USA for a custom DC variable which has data for 10 places. That would have thousands of empty entries.

I think we should gvc and understand why the child places are needed..

shifucun:
update on this PR? @n-h-diaz

PR #1381, opened 2024-06-06
Add new wrapper to Calculator that supports V2API and facets

n-h-diaz:
This is just a first pass to get the basic functionality working. 

I added a new wrapper because the existing wrapper (DerivedSeries) only reads from BT and doesn't support facets (which I think is by design, since I think the primary use case is to combine across imports) or ObsCollection

The new wrapper will only match facets and (like DerivedSeries) will only return one computed response. Since it uses V2API, it also supports ObsCollection without the need for child places. 

There are some pending TODOs to clean this up as a follow up, particularly relating to how facets are handled. Going forward, I can further separate the two wrappers, which will avoid precomputing the facets: 
* Define separate callbacks for Calculation (don't reuse anything from DerivedSeries)
* Define new type for Calcuation
* Make chooseItem a callback to Calculator, so Calculation and DerivedSeries can handle candidates independently
(This will simplify a lot of the logic in Calculation.computeCalculation)


shifucun:
This PR can be closed? @n-h-diaz

PR #1383, opened 2024-06-10
Add V2 Sparql endpoint (same as /v1/query endpoint)

shifucun:


proto/query.proto
```diff
@@ -49,6 +49,10 @@ message QueryRequest {
   string sparql = 1;
 }
 
+message SparqlRequest {
```
pradh, line 52: Curious why you didn't want to reuse `QueryRequest`.  If that's going to be deprecated, add a comment/TODO?

pradh (approved):
Thanks!

proto/query.proto
shifucun: That one has a different param `sparql` inside. Want to avoid confusing there.

PR #1384, opened 2024-06-10
[calculations] Add findObservationResponseHoles

n-h-diaz:
Step 1 of computing calculations: This just finds the holes from the base response which will be candidates for calculation

### DIFF
--- a/internal/server/v2/observation/calculation_util.go
+++ b/internal/server/v2/observation/calculation_util.go
@@ -1,0 +1,46 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package observation
+
+import (
+	pbv2 "github.com/datacommonsorg/mixer/internal/proto/v2"
+)
+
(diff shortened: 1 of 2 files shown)


internal/server/v2/observation/calculation_util.go
```diff
@@ -0,0 +1,46 @@
...
+// limitations under the License.
+
+package observation
+
+import (
+	pbv2 "github.com/datacommonsorg/mixer/internal/proto/v2"
+)
+
+// Given an input ObservationResponse, generate a map of variable -> entities with missing data.
+func findObservationResponseHoles(inputReq *pbv2.ObservationRequest, inputResp *pbv2.ObservationResponse) map[string]*pbv2.DcidOrExpression {
```
shifucun, line 22: nit, i would usually format the func with long args as:

func funName(
  arg1 ...,
  arg2 ...,
) retType {
}

internal/server/v2/observation/calculation_util.go
```diff
@@ -0,0 +1,46 @@
...
+	result := map[string]*pbv2.DcidOrExpression{}
+	if inputReq.Variable.GetFormula() != "" {
+		// Currently do not support nested formulas.
```
shifucun, line 25: shall we raise exception here to make this more explicit?

internal/server/v2/observation/calculation_util.go
```diff
@@ -0,0 +1,46 @@
...
+		return result
+	}
+	for variable, variableObs := range inputResp.ByVariable {
+		if len(inputReq.Entity.GetDcids()) > 0 {
+			emptyDcids := []string{}
```
shifucun, line 30: nit: holeEntities (to match function name and intent)

shifucun (commented):
nice PR! This is a really good start

internal/server/v2/observation/calculation_util.go

internal/server/v2/observation/calculation_util.go

internal/server/v2/observation/calculation_util.go

PR #1385, opened 2024-06-10
[calculations] Add function to parse AST tree and find leaf SVs

n-h-diaz:
Given a formula, returns a Calculation struct which contains 
* AST tree
* Map of leaf data (currently just SV dcid and facet) 
* List of leaf SVs

This follows a lot of logic/tests in Calculator for parsing the tree, except doesn't use a class and doesn't store candidates. (Data candidate handling will be deferred to each caller.)

### DIFF
--- a/internal/server/v2/observation/calculation_util.go
+++ b/internal/server/v2/observation/calculation_util.go
@@ -1,0 +1,168 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package observation
+
+import (
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"reflect"
+	"strings"
+
+	pb "github.com/datacommonsorg/mixer/internal/proto"
+)
+
+// The info of a node in the AST tree.
+type NodeData struct {
+	StatVar string
+	Facet   *pb.Facet
+}
+
+type Calculation struct {
+	Expr ast.Expr
+	// Key is encodeForParse(nodeName).
+	NodeDataMap map[string]*NodeData
+	StatVars    []string
+}
+
+// Golang's AST package is used for parsing the formula, so we need to avoid sensitive tokens for
+// AST. For those tokens, we swap them with insensitive tokens before the parsing, then swap them
+// back after the parsing.
+var (
+	encodeForParseTokenMap = map[string]string{
+		"dc/":          "_DC_SLASH_",
+		"dcAggregate/": "_DC_AGGREGATE_SLASH_",
+		"[":            "_LEFT_SQUARE_BRACKET_",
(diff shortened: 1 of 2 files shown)


internal/server/v2/observation/calculation_util.go
```diff
@@ -0,0 +1,168 @@
...
+	pb "github.com/datacommonsorg/mixer/internal/proto"
+)
+
+// The info of a node in the AST tree.
+type NodeData struct {
+	StatVar string
+	Facet   *pb.Facet
+}
+
+type Calculation struct {
```
shifucun, line 33: Since this does not do calculation, maybe call it "VariableFormula"?

internal/server/v2/observation/calculation_util.go
```diff
@@ -0,0 +1,168 @@
...
+	"go/ast"
+	"go/parser"
+	"reflect"
+	"strings"
+
+	pb "github.com/datacommonsorg/mixer/internal/proto"
+)
+
+// The info of a node in the AST tree.
+type NodeData struct {
```
shifucun, line 28: nit: ASTNode?

internal/server/v2/observation/calculation_util.go
```diff
@@ -0,0 +1,168 @@
...
+	StatVar string
+	Facet   *pb.Facet
+}
+
+type Calculation struct {
+	Expr ast.Expr
+	// Key is encodeForParse(nodeName).
```
shifucun, line 35: update the comment for key?

shifucun (commented):
looking good and just some minor naming question

internal/server/v2/observation/calculation_util.go

internal/server/v2/observation/calculation_util.go

internal/server/v2/observation/calculation_util.go
n-h-diaz: Done (I wasn't entirely sure what you wanted for the comment, so let me know if you'd like me to add something else)

internal/server/v2/observation/calculation_util.go
```diff
@@ -0,0 +1,168 @@
...
+
+// The info of a node in the AST tree.
+type ASTNode struct {
+	StatVar string
+	Facet   *pb.Facet
+}
+
+type VariableFormula struct {
+	Expr ast.Expr
+	// Key is encodeForParse(<name of ASTNode>).
```
shifucun, line 35: the meaning of the key is not clear just by reading the comment. Maybe explain it in plain word? With that, I feel `NodeDataMap` can also be renamed to be more precise.

internal/server/v2/observation/calculation_util.go
```diff
@@ -0,0 +1,168 @@
...
+		"=":            "_EQUAL_TO_",
+		";":            "_SEMICOLON_",
+	}
+	decodeForParseTokenMap = map[string]string{
+		"_DC_SLASH_":             "dc/",
+		"_DC_AGGREGATE_SLASH_":   "dcAggregate/",
+		"_LEFT_SQUARE_BRACKET_":  "[",
+		"_RIGHT_SQUARE_BRACKET_": "]",
+		"_EQUAL_TO_":             "=",
+		"_SEMICOLON_":            ";",
```
shifucun, line 58: instead of write the reverse map manually, how about reverting encodeForParseTokenMap?

internal/server/v2/observation/calculation_util.go
n-h-diaz: Added more details

internal/server/v2/observation/calculation_util.go
n-h-diaz: Good idea, done

shifucun (approved):
very nice!

PR #1388, opened 2024-06-11
[calculations] Add function to recursively compute a calculation

n-h-diaz:


internal/server/v2/observation/calculation_util.go
```diff
@@ -300,3 +300,36 @@ func evalBinaryExpr(
 	}
 	return result, nil
 }
+
+// Recursively iterate through the AST and perform the calculation.
+func evalExpr(
+	node ast.Node,
+	formula *VariableFormula,
+) (map[string]map[string][]*pb.PointStat, error) {
```
shifucun, line 308: Instead of define a nested function `computeChildObs`, maybe just call it out in line 321 and line 325 as it's only 4 lines. I feel that makes the switch part easier to follow.

n-h-diaz: Good idea, done

shifucun (approved):

PR #1389, opened 2024-06-12
[calculations] Detect holes in a V2ObservationResponse and attempt to fill them using calculations

n-h-diaz:


internal/server/v2/observation/calculation.go
```diff
@@ -17,56 +17,137 @@ package observation
...
-				// Successful calculation.
-				bv := derivedSeries.ByVariable
-				if len(bv[formula].ByEntity[entity].OrderedFacets) > 0 {
-					// Re-label formula response with the outputProperty and delete
-					// formula.
-					bv[variable] = bv[formula]
-					delete(bv, formula)
-					calculatedResponses = append(calculatedResponses, derivedSeries)
-					continue
+				if len(emptyDcids) < len(entity.GetDcids()) {
```
shifucun, line 141: is it possible that len(emptyDcids) > entity.GetDcids()? i wonder if you can remove the check of line 141

internal/server/v2/observation/calculation.go
```diff
@@ -17,56 +17,137 @@ package observation
...
+			if entity.GetExpression() != "" {
+				if len(variableObs.ByEntity) > 0 {
+					result = append(result, calculatedResp)
+					break
+				}
+			} else if len(entity.GetDcids()) != 0 {
+				emptyDcids := []string{}
+				for _, dcid := range entity.GetDcids() {
+					if _, ok := variableObs.ByEntity[dcid]; !ok {
+						emptyDcids = append(emptyDcids, dcid)
```
shifucun, line 138: nit: remove `emptyDcids`

internal/server/v2/observation/calculation.go
```diff
@@ -17,56 +17,137 @@ package observation
...
+	return newResp, nil
+}
+
+// Detects holes in a V2ObservationResponse and attempts to fill them using calculations.
+func MaybeCalculateHoles(
+	ctx context.Context,
+	store *store.Store,
+	cachedata *cache.Cache,
+	metadata *resource.Metadata,
+	httpClient *http.Client,
```
shifucun, line 98: Now as every function passes in these long list of args, should consolidate them into one class .. for later.

internal/server/v2/observation/calculation.go
```diff
@@ -17,56 +17,137 @@ package observation
...
 	store *store.Store,
-	inputResp *pbv2.ObservationResponse,
 	cachedata *cache.Cache,
-) []*pbv2.ObservationResponse {
-	calculatedResponses := []*pbv2.ObservationResponse{}
-	for variable, variableObservation := range inputResp.ByVariable {
+	metadata *resource.Metadata,
+	httpClient *http.Client,
+	formula string,
+	variable string,
```
shifucun, line 37: why need `variable`?

internal/server/v2/observation/calculation.go
```diff
@@ -17,56 +17,137 @@ package observation
...
+	}
+	newReq := &pbv2.ObservationRequest{
+		Variable: &pbv2.DcidOrExpression{Dcids: variableFormula.StatVars},
+		Entity:   entity,
+		Date:     inputReq.GetDate(),
+		Value:    inputReq.GetValue(),
+		Filter:   inputReq.GetFilter(),
+		Select:   inputReq.GetSelect(),
+	}
+	variableResp, err := ObservationInternal(ctx, store, cachedata, metadata, httpClient, newReq)
```
shifucun, line 53: s/variableResp/initialObs

internal/server/v2/observation/calculation.go
```diff
@@ -17,56 +17,137 @@ package observation
...
+	if err != nil {
+		return nil, err
+	}
+	computeLeafObs(variableResp, variableFormula)
```
shifucun, line 57: now the VariableFormula is not a formula anymore, but the actual tree with observation.

I wonder if it's possible to move `CandidateObs` out of `ASTNode`?

internal/server/v2/observation/calculation.go
```diff
@@ -17,56 +17,137 @@ package observation
...
+	newResp := &pbv2.ObservationResponse{
```
shifucun, line 58: It's not clear why there are both `computeLeafObs()` and `evalExpr(variableFormula.Expr, variableFormula)`

Also this is followed by using calculatedResp to construct final response.

I feel these can be combined and simplfied?

internal/server/v2/observation/calculation.go
n-h-diaz: It's not possible to be greater but it can be equal (i.e. nothing got computed from this round), so I didn't want to keep appending empty responses to result in this case

internal/server/v2/observation/calculation.go
n-h-diaz: It's just for constructing the final response (since the formula doesn't reference the target variable anywhere)

internal/server/handler_v2.go
```diff
@@ -210,11 +210,28 @@ func (s *Server) V2Event(
 func (s *Server) V2Observation(
 	ctx context.Context, in *pbv2.ObservationRequest,
 ) (*pbv2.ObservationResponse, error) {
-	mergedResp, err := v2observation.ObservationInternal(ctx, s.store, s.cachedata.Load(), s.metadata, s.httpClient, in)
+	mergedResp, err := v2observation.ObservationInternal(
```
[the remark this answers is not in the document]
n-h-diaz: Done

internal/server/v2/observation/calculation.go
n-h-diaz: Done

internal/server/v2/observation/calculation.go
n-h-diaz: Done

internal/server/v2/observation/calculation.go
n-h-diaz: evalExpr is a separate function because it's recursive. I also created a separate function for computeLeafObs to only find matching obs once (and then can use these obs throughout the recursion) rather that recomputing each time a node potentially shows up. I also structured it as entity -> facet -> obs to make it easier to match facets (vs ObservationResponse, where facetId is part of the obs), but this means I have to reconstruct the response later. I moved the candidate obs out of the formula (based on your above comment) - let me know if this makes sense, or if you'd prefer I structure it differently!

internal/server/v2/observation/calculation.go
n-h-diaz: Changed to "newEntityDcids"

n-h-diaz:
Made some updates based on discussion today, primarily:
* Changed some names to be clearer
* Updated to directly merge ObservationResponses (rather than some other structure), so no need to reconstruct the response
* (Side effect of ^) Preserve order of facets

Please take a look when you get the chance 😃

internal/server/v2/observation/calculation.go
shifucun: maybe have a class to wrap the formula and the target variable together?

internal/server/v2/observation/calculation_util.go
```diff
@@ -212,90 +211,143 @@ func compareFacet(facet1, facet2 *pb.Facet) bool {
...
+	result := &pbv2.ObservationResponse{
+		// Intermediate responses have no StatVar, so use a placeholder.
+		ByVariable: map[string]*pbv2.VariableObservation{"": {}},
+		Facets:     map[string]*pb.Facet{},
+	}
+	if len(x.ByVariable) != 1 || len(y.ByVariable) != 1 {
+		return nil, fmt.Errorf("more than one variable in intermediate response")
+	}
+	for _, xVariableObs := range x.ByVariable {
+		for _, yVariableObs := range y.ByVariable {
```
shifucun, line 308: no need to have a for loop x.ByVariable and y.ByVariable if the length is 1 (from line 304)

internal/server/v2/observation/calculation_util.go
```diff
@@ -212,90 +211,143 @@ func compareFacet(facet1, facet2 *pb.Facet) bool {
...
+			for entity, xEntityObs := range xVariableObs.ByEntity {
+				yEntityObs, ok := yVariableObs.ByEntity[entity]
+				if !ok {
+					continue
+				}
+				newOrderedFacets := []*pbv2.FacetObservation{}
+				for i := 0; i < len(xEntityObs.OrderedFacets); i++ {
```
shifucun, line 315: after line 314, can define two short-name alias and use them below

xFacets := xEntityObs.OrderedFacets
yFacets := yEntityObs.OrderedFacets

internal/server/v2/observation/calculation_util.go
```diff
@@ -212,90 +211,143 @@ func compareFacet(facet1, facet2 *pb.Facet) bool {
...
+					for j := 0; j < len(yEntityObs.OrderedFacets); j++ {
+						if xEntityObs.OrderedFacets[i].GetFacetId() == yEntityObs.OrderedFacets[j].GetFacetId() {
+							newFacetId := xEntityObs.OrderedFacets[i].GetFacetId()
```
shifucun, line 318: so merged facet id is the same as x? Need a todo to figure out full solution for new facet id?

internal/server/v2/observation/calculation_util.go

internal/server/v2/observation/calculation_util.go

internal/server/v2/observation/calculation_util.go
n-h-diaz: Right now I'm just using the same facet and keeping all the fields the same. I think we can maybe also update to use dcAggregate/ + maybe change the import name (to match what's done now in pipeline), but I need to figure out how to update the facetId accordingly -- added a TODO for this

PR #1391, opened 2024-06-13
Check formulas for computing count

n-h-diaz:
Also moves out some formula functions into a separate package to avoid import cycles

I added the check to count rather than existence, since other apis also use count, so they'll also get the calculated observations added


internal/server/count/count.go
```diff
@@ -132,3 +128,62 @@ func Count(
...
+		for _, entity := range entities {
+			if _, ok := result[dcid][entity]; ok {
+				continue
+			}
+			for _, f := range formulas {
+				variableFormula, err := formula.NewVariableFormula(f)
+				if err != nil {
+					return nil, err
+				}
+				calculatedCount, err := countInternal(
```
shifucun, line 164: can you batch the countInternal call for all entities and variables?

n-h-diaz: Done

internal/server/count/count.go
```diff
@@ -132,3 +128,69 @@ func Count(
...
+	svOrSvgs []string,
+	entities []string,
+) (map[string]map[string]int32, error) {
+	result, err := countInternal(ctx, st, cachedata, svOrSvgs, entities)
+	if err != nil {
+		return nil, err
+	}
+	// Check for count for computed observations.
+	// Use counts of formula variables as a heuristic.
+	for _, dcid := range svOrSvgs {
```
shifucun, line 150: nit: s/dcid/svOrSvg

internal/server/count/count.go
```diff
@@ -132,3 +128,69 @@ func Count(
...
+		formulas, ok := cachedata.SVFormula()[dcid]
+		if !ok {
+			continue
+		}
+		missingEntities := []string{}
+		for _, entity := range entities {
+			if _, ok := result[dcid][entity]; !ok {
+				missingEntities = append(missingEntities, entity)
+			}
+		}
```
shifucun, line 160: If the svOrSvg is a computed one, no entities would have the count data from countInternal? Should the following count happens when len(missingEntities) == len(entities)?

internal/server/count/count.go
n-h-diaz: Done

internal/server/count/count.go
n-h-diaz: Spoke offline - updated to check for all formulas in one call

shifucun (approved):
thanks for the update!

PR #1392, opened 2024-06-18
Add coal and a few more words to bogus-names

pradh:


chejennifer (approved):
nice

pradh:
Adding a few more common word counties.

PR #1393, opened 2024-06-18
Add Congo alt names

pradh:


shifucun (approved):
Are there any other similar country names?

PR #1412, opened 2024-08-29
Data release

hareesh-ms:
Data release


internal/server/v2/facet/golden/contained_in_facet/IND_AA1.json
```diff
@@ -39,11 +36,6 @@
       "measurement_method": "OECDRegionalStatistics",
       "observation_period": "P1Y"
     },
-    "1456184638": {
```
ajaits, line 42: Latest population data in KG is from 2023 July, seems like was dropped in refresh

internal/server/v2/observation/golden/contained_in_all/FRA_AA2.json
```diff
@@ -1180,18 +1192,6 @@
               "earliest_date": "1975",
               "latest_date": "2019"
             },
-            {
-              "facet_id": "1456184638",
```
ajaits, line 1184: Is there a way to see the place dcid for this deleted stat? Need to check if this was dropped in KG refresh

ajaits (approved):
The deletions are for population dropped in wiki refresh.

PR #1413, opened 2024-09-05
Add certs for staging.api-esp.datacommons.org and api-esp.datacommons.org

hqpho:


beets (approved):
thanks for the fix!

could you add a comment in the pr as to why it isn't required for autopush? in most other ways, the configs are identical.

hqpho:
This isn't required for autopush because autopush API has not yet been migrated to use Apigee.

PR #1414, opened 2024-09-10
Add script for verifying API responses won't change

hqpho:
These validate that all endpoints continue to work with routing through a new load balancer.


tools/migrationtesting/compare_responses.py
beets, line 1: super nit: rename migrationtesting to migration_testing

tools/migrationtesting/compare_responses.py
```diff
@@ -0,0 +1,502 @@
...
+    # 405
+    ("/v1/recon/resolve/id", ["GET"], {}),
+    # 415
+    ("/query?name=example.com&type=A", ["GET"], {}),
+    # 500
+    ("/v1/recon/resolve/id", ["POST"], {}),
+]
+
+
+class bcolors:
```
beets, line 394: fancy!

beets (approved):
very nice - thanks for doing the detailed diffs for the migration

tools/migrationtesting/compare_responses.py
hqpho: Done.

PR #1417, opened 2024-09-18
Improve Apigee deployment

hqpho:
- Use Secrets Manager for *.env files
- Add quota limit for trial API key
- Upgrade environment type to handle more QPS

### DIFF
--- a/deploy/apigee/policies/rewrite-quota-exceeded-message.template.xml
+++ b/deploy/apigee/policies/rewrite-quota-exceeded-message.template.xml
@@ -1,0 +1,7 @@
+<AssignMessage async="false" continueOnError="false" enabled="true"
+name="rewrite-quota-exceeded-message">
+  <DisplayName>rewrite-quota-exceeded-message</DisplayName>
+  <Set>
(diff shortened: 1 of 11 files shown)


deploy/apigee/policies/rewrite-quota-exceeded-message.template.xml
```diff
@@ -0,0 +1,7 @@
+<AssignMessage async="false" continueOnError="false" enabled="true"
+name="rewrite-quota-exceeded-message">
+  <DisplayName>rewrite-quota-exceeded-message</DisplayName>
+  <Set>
+    <Payload contentType="application/json">{"message": "Quota limit exceeded. If you are using a trial API key, please register at REPLACE_WITH_API_PORTAL_URL to obtain a different key."}</Payload>
```
dwnoble, line 5: How hard would it be to only display this if they're actually using the trial key?

hqpho: This actually is currently only displayed if they're using the trial key, since the quota policy is only applied to requests that use that particular key.

PR #1419, opened 2024-09-23
Prep for Bard API migration to datcom-apigee

hqpho:
Right now bard.datacommons.org points to a separate Apigee org on datcom-bard. These changes prepare to point bard.datacommons.org to the Apigee org on datcom-apigee, which is used by api.datacommons.org and nl.datacommons.org as well.


tools/migration_testing/compare_responses.py
```diff
@@ -411,11 +49,26 @@ class bcolors:
...
+  if is_json(response):
+    return response.json()
+  else:
+    return response.content
+
+
+def get_formatted_response(response):
+  if is_json(response):
+    formatted = json.dumps(response.json(), indent=2).replace("\n", "\n  ")
+    if len(formatted) > 1000:
```
shifucun, line 66: nit: make 1000 a constant

tools/migration_testing/compare_responses.py
```diff
@@ -426,42 +79,47 @@ def compare_responses(endpoint, use_api_key, method="GET", params=None):
     if method == "GET":
       req_params = {} if params is None else copy.deepcopy(params)
       if use_api_key:
-        req_params['key'] = args.api_key  # Add API key to query params
+        if args.endpoints == "nl":
```
shifucun, line 82: add a comment why "nl" is special?

tools/migration_testing/compare_responses.py
```diff
@@ -488,25 +146,41 @@ def send_requests(endpoint_infos, use_api_key):
...
-    send_requests(ERROR_TESTS, use_api_key=True)
+    send_requests(error_tests, use_api_key=True)
 
 
 if __name__ == "__main__":
   parser = argparse.ArgumentParser(
       description="Compare API responses between two domains with an API key.")
+  parser.add_argument(
+      "endpoints",
+      help="Which set of endpoints to test. Values are 'mixer', 'nl', or 'new'."
```
shifucun, line 182: I feel this endpoints a bit confusing. maybe use the subdomain directly?

shifucun (commented):
Great to adding the tests!

deploy/apigee/terraform/prod/routing.tf
```diff
@@ -22,7 +22,8 @@ locals {
     },
     "matcher-api2" = {
       hostname = var.api2_hostname
-      prefix   = "/api/"
+      # prefix   = "/api/"
```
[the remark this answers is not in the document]
hqpho: I left this on purpose to make it super clear that it used to point to /api/ but now is pointing to /bard/. Will clean it up when I remove the api2 config altogether after Bard migration is done.

tools/migration_testing/compare_responses.py
hqpho: Since the subdomain for Mixer API is just "api" I think that might also be confusing. I extended the comment - wdyt?

tools/migration_testing/compare_responses.py
hqpho: done

tools/migration_testing/compare_responses.py
hqpho: done

tools/migration_testing/compare_responses.py
shifucun: Wonder if `endpoints`, `current_domain` and `new_domain` can be consolidated. But it's good as is.

shifucun (approved):

PR #1420, opened 2024-09-26
Add is_dc_aggregate to facet id

n-h-diaz:
This is so online Mixer aggregations can be tagged with "is_dc_aggregate" so that the output facet will be different than the input facets / the response indicates that this has been aggregated by us

This will modify all the existing facet ids, so the goldens need to be updated too


keyurva (approved):
Thanks Natalie!

Can you add more context in the PR description for future reference?

Also, since this will change all observation responses, we should try to get this to staging sooner than usual and monitor it closely around prod rollout.

n-h-diaz:
> Thanks Natalie!
> 
> Can you add more context in the PR description for future reference?
> 
> Also, since this will change all observation responses, we should try to get this to staging sooner than usual and monitor it closely around prod rollout.

Thanks for the review! 

Added more to the description and sounds good to test early

PR #1415, opened 2024-09-16
Add support for constants in StatisticalCalculation

n-h-diaz:
This allows the schema to specify formulas with constant values (e.g. "(Count_Person * 2.5) + 1"). This is related to a One.org request to enable currency conversions in StatisticalCalculations.

StatisticalCalculations are still flag-protected by "cacheSVFormula" and are only enabled on autopush

PR #1421, opened 2024-09-26
Set is_dc_aggregate for calculated responses

n-h-diaz:
This is so the output facet will be different than the input facets and the response indicates that the result is aggregated.

This will also modify the facet ids for custom import calculations

UPDATE: Based on offline feedback, this also modifies the facet id construction to only include fields that have been set. This is to help keep the ids stable even as fields are potentially added in the future.


internal/server/v2/observation/calculation_util.go
```diff
@@ -245,3 +246,34 @@ func evalExpr(
 		return nil, fmt.Errorf("unsupported ast type %T", t)
 	}
 }
+
+func formatCalculatedResponse(
+	resp *pbv2.ObservationResponse,
+	equation *Equation,
+) error {
+	// Replace placeholder by final variable.
+	variableObs, ok := resp.ByVariable[INTERMEDIATE_NODE]
```
keyurva, line 255: It feels like we are using placeholder nodes and manipulating them because we are using `ObservationResponse` as the intermediate object as well.

Not for this PR, but I'm wondering if we can simplify things by using a custom struct for intermediate responses and building `ObservationResponse` only at the end.

internal/server/v2/observation/calculation_util.go
```diff
@@ -245,3 +246,34 @@ func evalExpr(
...
+		return fmt.Errorf("missing intermediate variable in intermediate response")
+	}
+	resp.ByVariable[equation.variable] = variableObs
+	delete(resp.ByVariable, INTERMEDIATE_NODE)
+
+	// Update Facets with IsDcAggregate=true.
+	newFacets := map[string]*pb.Facet{}
+	facetIdMap := map[string]string{}
+	for oldFacetId, oldFacet := range resp.Facets {
+		newFacet := oldFacet
```
keyurva, line 266: Doing this will update `oldFacet` as well. Wondering if you should first clone it and then update just the clone.

```
newFacet := proto.Clone(oldFacet).(*pb.Facet)
```

n-h-diaz: Good catch, thanks! Updated

internal/server/v2/observation/calculation_util.go
n-h-diaz: Good idea, I can do this in a follow up PR and as a precursor to https://github.com/datacommonsorg/mixer/pull/1415

internal/util/util.go
```diff
@@ -481,15 +480,20 @@ func GetFacet(s *pb.SourceSeries) *pb.Facet {
...
+	// This is so the hashes stay consistent if more fields are added.
+	s := strings.Join([]string{
 		m.ImportName,
 		m.MeasurementMethod,
 		m.ObservationPeriod,
 		m.ScalingFactor,
 		m.Unit,
-		strconv.FormatBool(m.IsDcAggregate),
-	}, "-")))
+	}, "")
```
keyurva, line 491: Any reason to not continue joining them with a "-"? If you leave it that way, majority of the goldens would likely stay the same.

n-h-diaz: This is so if we add more string fields the hash won't change (if we use a connector, then there will be a connector even if the field is an empty string). 

If we want to preserve the existing hashes, I could keep the connector but for any future fields just append if it exists, rather than adding to the join (like for the IsDcAggregate)

keyurva: I think without a separator, we may run into bugs in some scenarios. e.g. imagine a facet with Unit = foo and everything else not set and another one with MeasurementMethod = foo and everything else not set. Without a separator, both would produce the same string and hence yield the same facet ID.

n-h-diaz: Makes sense, I added the "-" back 

(The ids are still changing in this PR because we're making the is_dc_aggregate optional, while in the previous PR we always added it)

keyurva (approved):
Thanks Natalie!

PR #1424, opened 2024-09-30
Add pagination to formula_fetcher

n-h-diaz:
Fixes a bug where the formula_fetcher previously did not include pagination, so only retrieved the first page of formulas. We recently added several hundred more formulas, so need to use pagination to retrieve them all. As part of this change, switches the formula_fetcher to use the triples package directly to make pagination simpler. 

This fixes some goldens that had been empty due to the bug before.

### DIFF
--- a/internal/server/statvar/fetcher/formula_fetcher.go
+++ b/internal/server/statvar/fetcher/formula_fetcher.go
@@ -41,9 +41,9 @@ func FetchFormulas(
 	remoteRespChan := make(chan *pbv2.NodeResponse, 1)
 	// Fetch for BT and SQL.
 	errGroup.Go(func() error {
-		statCal := []string{}
+		statCalDcids := []string{}
 		nextToken := ""
-		for ok := true; ok; ok = (nextToken != "") {
+		for {
 			statCalReq := &pbv1.TriplesRequest{
 				Node:      "StatisticalCalculation",
 				Direction: "in",
@@ -61,19 +61,23 @@ func FetchFormulas(
 			typeOf, ok := statCalResp.Triples["typeOf"]
 			if ok {
 				for _, node := range typeOf.Nodes {
-					statCal = append(statCal, node.Dcid)
+					statCalDcids = append(statCalDcids, node.Dcid)
 				}
 			}
 			nextToken = statCalResp.GetNextToken()
+			if nextToken == "" {
+				break
+			}
 		}
-		if len(statCal) == 0 {
+		if len(statCalDcids) == 0 {
 			return nil
 		}
 		localResp := &pbv2.NodeResponse{Data: map[string]*pbv2.LinkedGraph{}}
-		for ok := true; ok; ok = (nextToken != "") {
+		for {
 			bulkReq := &pbv1.BulkTriplesRequest{
-				Nodes:     statCal,
+				Nodes:     statCalDcids,
 				Direction: "out",
+				NextToken: nextToken,
 			}
 			bulkResp, err := triples.BulkTriples(
 				errCtx,
@@ -99,6 +103,9 @@ func FetchFormulas(
 				}
 			}
 			nextToken = bulkResp.GetNextToken()
+			if nextToken == "" {
+				break
+			}
 		}
 		localRespChan <- localResp
 		return nil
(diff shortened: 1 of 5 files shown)


internal/server/statvar/fetcher/formula_fetcher.go
```diff
@@ -39,83 +41,108 @@ func FetchFormulas(
...
-		)
-		if err != nil {
-			return err
-		}
 		statCal := []string{}
-		for _, nodes := range data["StatisticalCalculation"]["typeOf"] {
-			for _, node := range nodes {
-				statCal = append(statCal, node.Dcid)
+		nextToken := ""
+		for ok := true; ok; ok = (nextToken != "") {
```
keyurva, line 46: nit: would it be cleaner to do something like:

```
for {
  ...
  if nextToken == "" {
    break
  }
}
```

internal/server/statvar/fetcher/formula_fetcher.go
```diff
@@ -39,83 +41,108 @@ func FetchFormulas(
...
+			statCalReq := &pbv1.TriplesRequest{
+				Node:      "StatisticalCalculation",
+				Direction: "in",
+				NextToken: nextToken,
+			}
+			statCalResp, err := triples.Triples(
```
keyurva, line 52: Not for this PR but something to think about for the future.

The next token approach necessitates sequential behavior. Could we come up with a design that allows us to fetch multiple pages concurrently?

internal/server/statvar/fetcher/formula_fetcher.go
```diff
@@ -39,83 +41,108 @@ func FetchFormulas(
...
-			[]string{"StatisticalCalculation"},
-			[]string{"typeOf"},
-			0,
-			"",
-			"in",
-		)
-		if err != nil {
-			return err
-		}
 		statCal := []string{}
```
keyurva, line 44: nit: these are calculation dcids, correct? Might be worthwhile to name it so and / or add a comment.

internal/server/statvar/fetcher/formula_fetcher.go
```diff
@@ -39,83 +41,108 @@ func FetchFormulas(
...
-			return err
-		}
-		// Wrap result in pbv2.NodeResponse.
 		localResp := &pbv2.NodeResponse{Data: map[string]*pbv2.LinkedGraph{}}
-		for dcid, data := range resp {
-			localResp.Data[dcid] = &pbv2.LinkedGraph{
-				Arcs: map[string]*pbv2.Nodes{},
+		for ok := true; ok; ok = (nextToken != "") {
+			bulkReq := &pbv1.BulkTriplesRequest{
+				Nodes:     statCal,
```
keyurva, line 75: Would this scale if we have millions or even 10s of thousands of stat cals?

Actually, can you help me understand the overall context of where and how the formula fetcher is used? e.g. is it used by external clients, internally, is it called once and cached, etc.

internal/server/statvar/fetcher/formula_fetcher.go
n-h-diaz: Done

internal/server/statvar/fetcher/formula_fetcher.go
n-h-diaz: Spoke offline: we decided that since this is just loaded once on Mixer start, it's okay to do the multiple requests

internal/server/statvar/fetcher/formula_fetcher.go
n-h-diaz: Good idea, I added "dcid" to the name

internal/server/statvar/fetcher/formula_fetcher.go
n-h-diaz: Spoke offline: this is loaded once and stored in Mixer cache

internal/server/statvar/fetcher/formula_fetcher.go
```diff
@@ -39,51 +41,70 @@ func FetchFormulas(
...
+				return err
 			}
-			for prop, nodes := range data {
-				if len(nodes) > 0 {
-					localResp.Data[dcid].Arcs[prop] = &pbv2.Nodes{
-						Nodes: v1pv.MergeTypedNodes(nodes),
-					}
+			// Wrap result in pbv2.NodeResponse.
+			for _, nodeTriples := range bulkResp.Data {
+				out, outOk := nodeTriples.Triples["outputProperty"]
```
keyurva, line 93: nit: Consider making the string literals that are used in multiple places constants.

keyurva (approved):
Thanks for the updates!

internal/server/statvar/fetcher/formula_fetcher.go
n-h-diaz: Good idea, done

n-h-diaz:
Thanks for the review!

PR #1427, opened 2024-10-04
Remove intermediate node for calculations

n-h-diaz:
Follow up to comment in https://github.com/datacommonsorg/mixer/pull/1421 to help simplify the calculation construction logic


keyurva (approved):
Thanks for the simplification!

PR #1428, opened 2024-10-04
Return an empty list of groups if no input groups are specified.

keyurva:
* Bug: https://issuetracker.google.com/370724577
* For statvar search calls, the input groups are empty which results in a bad sql query causing this issue.
* The fix checks for this condition and returns an empty list which is the correct behavior.


n-h-diaz (approved):
Thanks!

PR #1429, opened 2024-10-07
Joins responses from Remote and Local mixer for SearchStatVar.

gmechali:
Includes responses from both local (custom DC) and remote (base DC) mixers in stat var search. This PR aims to resolve https://b.corp.google.com/issues/370730321

Local testing verified the behavior.
See the two custom DC Stat vars: https://screenshot.googleplex.com/7a3F7W44xGyL5pR
Here is the stat var search: https://screenshot.googleplex.com/BE35RoUiEZhCcUX

### DIFF
--- a/internal/merger/merger.go
+++ b/internal/merger/merger.go
@@ -367,7 +367,7 @@ func toStatVarSummaryMap(in []*pbv1.VariableInfoResponse) map[string]*proto.Stat
 	return out
 }
 
-func processSearchStatVarResponse(resp *proto.SearchStatVarResponse, mergedStatVars []*proto.EntityInfo, matchesMap map[string]struct{}, dedupedMatches []string) ([]*proto.EntityInfo, map[string]struct{}, []string) {
+func processSearchStatVarResponse(resp *proto.SearchStatVarResponse, mergedStatVars []*proto.EntityInfo, matchesMap map[string]struct{}, dedupedMatches []string) ([]*proto.EntityInfo, []string) {
 	if resp != nil {
 		mergedStatVars = append(mergedStatVars, resp.StatVars...)
 
@@ -378,7 +378,7 @@ func processSearchStatVarResponse(resp *proto.SearchStatVarResponse, mergedStatV
 			}
 		}
 	}
-	return mergedStatVars, matchesMap, dedupedMatches
+	return mergedStatVars, dedupedMatches
 }
 
 // MergeSearchStatVarResponse merges two SearchStatVarResponse.
@@ -387,8 +387,8 @@ func MergeSearchStatVarResponse(primary, secondary *proto.SearchStatVarResponse)
 	dedupedMatches := []string{}
 	matchesMap  := map[string]struct{}{}
 	
-	mergedStatVars, matchesMap, dedupedMatches = processSearchStatVarResponse(primary, mergedStatVars, matchesMap, dedupedMatches)
-	mergedStatVars, matchesMap, dedupedMatches = processSearchStatVarResponse(secondary, mergedStatVars, matchesMap, dedupedMatches)
+	mergedStatVars, dedupedMatches = processSearchStatVarResponse(primary, mergedStatVars, matchesMap, dedupedMatches)
+	mergedStatVars, dedupedMatches = processSearchStatVarResponse(secondary, mergedStatVars, matchesMap, dedupedMatches)
 	
 	merged := &proto.SearchStatVarResponse{
 		StatVars: mergedStatVars,
--- a/internal/merger/merger.go
+++ b/internal/merger/merger.go
@@ -374,9 +374,7 @@ func MergeSearchStatVarResponse(primary, secondary *proto.SearchStatVarResponse)
 	matchesMap  := map[string]bool{}
 
 	if primary != nil {
-		for _, sv := range primary.StatVars {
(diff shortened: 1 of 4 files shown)


internal/server/handler_v1.go
```diff
@@ -532,20 +532,23 @@ func (s *Server) SearchStatVar(
...
+			remoteResp,
+		); err != nil {
+			return nil, err
 		}
 	}
-	return localResp, nil
+
+	resp := &pb.SearchStatVarResponse{
+		StatVars: append(localResp.StatVars[:], remoteResp.StatVars...),
+		Matches:  append(localResp.Matches[:], remoteResp.Matches...),
```
keyurva, line 551: The proto definition [mentions](https://github.com/gmechali/mixer/blob/8a5d4ee1a4cbe4b3a757aaf37040082f31a4e1e7/proto/stat_var.proto#L237) that matches should be unique strings. Can we ensure there are no duplicates in the merged result?

internal/server/handler_v1.go
```diff
@@ -532,20 +532,23 @@ func (s *Server) SearchStatVar(
...
+			"/v1/variable/search",
+			in,
+			remoteResp,
+		); err != nil {
+			return nil, err
 		}
 	}
-	return localResp, nil
+
+	resp := &pb.SearchStatVarResponse{
```
keyurva, line 549: Couple of suggestions for best practices. OK to implement them in a separate PR:

1. Consider moving the merge logic to `merger.go` ([example](https://github.com/datacommonsorg/mixer/blob/08a8e50c6c5c935159ea8e4a6986408d1cb4c66f/internal/merger/merger.go#L332))
2. Consider adding tests in `merger_test.go` ([example](https://github.com/datacommonsorg/mixer/blob/08a8e50c6c5c935159ea8e4a6986408d1cb4c66f/internal/merger/merger_test.go#L798)) and / or `handler_v1_test.go` ([example](https://github.com/datacommonsorg/mixer/blob/08a8e50c6c5c935159ea8e4a6986408d1cb4c66f/internal/server/handler_v1_test.go#L30))

gmechali: Done.

internal/server/handler_v1.go
gmechali: Done.

internal/merger/merger.go
```diff
@@ -366,3 +366,43 @@ func toStatVarSummaryMap(in []*pbv1.VariableInfoResponse) map[string]*proto.Stat
 	}
 	return out
 }
+
+// MergeSearchStatVarResponse merges two SearchStatVarResponse.
+func MergeSearchStatVarResponse(primary, secondary *proto.SearchStatVarResponse) *proto.SearchStatVarResponse {
+	mergedStatVars := []*proto.EntityInfo{}
+	dedupedMatches := []string{}
+	matchesMap  := map[string]bool{}
```
keyurva, line 374: The go convention for a set-like map is `map[string]struct{}` ([example](https://github.com/datacommonsorg/mixer/blob/08a8e50c6c5c935159ea8e4a6986408d1cb4c66f/internal/util/util.go#L720)). Consider using that.

internal/merger/merger.go
```diff
@@ -366,3 +366,43 @@ func toStatVarSummaryMap(in []*pbv1.VariableInfoResponse) map[string]*proto.Stat
...
+
+	if primary != nil {
+		mergedStatVars = append(mergedStatVars, primary.StatVars...)
+
+		for _, m := range primary.Matches {
+			if _, ok := matchesMap[m]; ok {
```
keyurva, line 380: Can check for `!ok` instead and eliminate the `else` block.

internal/merger/merger.go
```diff
@@ -366,3 +366,43 @@ func toStatVarSummaryMap(in []*pbv1.VariableInfoResponse) map[string]*proto.Stat
 	}
 	return out
 }
+
+// MergeSearchStatVarResponse merges two SearchStatVarResponse.
+func MergeSearchStatVarResponse(primary, secondary *proto.SearchStatVarResponse) *proto.SearchStatVarResponse {
+	mergedStatVars := []*proto.EntityInfo{}
+	dedupedMatches := []string{}
```
keyurva, line 373: Appreciate the expressive var names.

internal/merger/merger.go
```diff
@@ -366,3 +366,43 @@ func toStatVarSummaryMap(in []*pbv1.VariableInfoResponse) map[string]*proto.Stat
...
+				// skip
+			} else {
+				matchesMap[m] = true
+				dedupedMatches = append(dedupedMatches, m)
+			}
+		}
+	}
+	
+	if secondary != nil {
+		mergedStatVars = append(mergedStatVars, secondary.StatVars...)
```
keyurva, line 390: optional: Since logic to merge both primary and secondary is the same, considering extracting it out into an internal function to avoid duplicating logic.

keyurva (approved):
Very nice!

internal/merger/merger.go
gmechali: Done.

internal/merger/merger.go
gmechali: Done.

internal/merger/merger.go
gmechali: Done.

PR #1431, opened 2024-10-08
Create simple utility for logging with tags and log all remote mixer calls.

keyurva:
* Log tags can be defined as `DCLogTag` enums.
* `DCLog` function can be called to log tagged messages.
* This PR uses the above to log all calls to remote mixer.
   + Example: `[DC][RemoteMixerCall] url=https://api.datacommons.org/v2/node`
   + Currently it only logs the URL being called.
   + We could consider supporting a debug mode and log requests / responses in the future.


n-h-diaz (approved):
Cool!

PR #1432, opened 2024-10-08
Ignore key parameter instead of failing request.

keyurva:
* This PR ignores the `key` parameter in the request vs failing requests with that parameter.
* This makes it convenient for users that sometimes copy main dc api urls and use them for their custom dc without updating any parameters. 
  + Currently if the original URL includes a `key` parameter it fails on custom dcs with an obscure 415 error even though the key is not used by custom DCs.
  + With this fix, it will ignore the parameter and proceed with handling the request as usual.
* Fixit task: http://shortn/_9STKYyVCoR

#### Before

```
~ curl -I -X GET "https://dc-autopush-kqb7thiuka-uc.a.run.app/core/api/v2/node?nodes=Class&property=<-typeOf&key=foo"
HTTP/2 415 
content-type: application/grpc
x-envoy-upstream-service-time: 0
x-cloud-trace-context: e94d41d4dab5d53dba1664ef053c33aa;o=1
date: Tue, 08 Oct 2024 17:55:57 GMT
server: Google Frontend
content-length: 0
alt-svc: h3=":443"; ma=2592000,h3-29=":443"; ma=2592000
```

#### After

```
~ curl -I -X GET "http://localhost:8081/v2/node?nodes=Class&property=<-typeOf&key=foo"
HTTP/1.1 200 OK
content-type: application/json
x-envoy-upstream-service-time: 149
grpc-status: 0
grpc-message: 
content-length: 56329
date: Tue, 08 Oct 2024 17:56:10 GMT
server: envoy
```

### DIFF
--- a/esp/envoy-config.yaml
+++ b/esp/envoy-config.yaml
@@ -49,6 +49,7 @@ static_resources:
                       print_options:
                         always_print_enums_as_ints: false
                         preserve_proto_field_names: false
+                      ignored_query_parameters: ["key"]
                   - name: envoy.filters.http.router
                     typed_config:
                       "@type": type.googleapis.com/envoy.extensions.filters.http.router.v3.Router


hqpho:
- Is envoy-config.yaml used only in a custom DC context?
- Is this fix compatible with custom DCs eventually having the option to require API keys?

keyurva:
> Is envoy-config.yaml used only in a custom DC context?

It's used for both. The key is however never processed by envoy. For base dc it was originally handled by apikeys and now apigee.

> Is this fix compatible with custom DCs eventually having the option to require API keys?

Yes - keys would be processed by custom dcs fronted by apigee (similar to base dc today).

hqpho:
Base DC actually still does need to handle a key at the ESP level to make sure that only requests from Apigee are allowed. If custom DC uses the same setup (Apigee in front of ESP), the same will be true for custom DC. Will this fix be compatible with that? Or has the `key` param already been verified ESP by this point?

keyurva:
The `key` param has already been verified. Our use of envoy is limited to grpc transcoding (something we have talked about getting rid of at some point). This PR simply configures it to ignore the key parameter when transcoding.

hqpho (approved):

PR #1433, opened 2024-10-09
Add support for recognizing US zip codes.

keyurva:
### Before

```
~ curl --location 'https://dc-autopush-kqb7thiuka-uc.a.run.app/core/api/v1/recognize/places' \
--header 'Content-Type: application/json' \
--data '{
  "queries": ["population in 94043"]
}'

{"queryItems":{"population in 94043":{"items":[{"span":"population in 94043"}]}}}
```

### After

```
~ curl --location 'localhost:8081/v1/recognize/places' \
--header 'Content-Type: application/json' \
--data '{
  "queries": ["population in 94043"]
}'

{"queryItems":{"population in 94043":{"items":[{"span":"population in"},{"span":"94043","places":[{"dcid":"zip/94043"}]}]}}}
```


internal/store/files/recog_place_store.go
```diff
@@ -80,6 +82,25 @@ type RecogPlaceStore struct {
...
+
+// RecogUSZipCode recognizes US zipcodes and returns a RecogPlaces object if it recognizes one.
+// If detected, the RecogPlaces includes a list of RecogPlace of len 1.
+// If not detected, it returns nil.
+func (store *RecogPlaceStore) RecogUSZipCode(token string) *pb.RecogPlaces {
+	if _, ok := store.USZipCodes[token]; !ok {
+		return nil
+	} else {
+		dcid := fmt.Sprintf("zip/%s", token)
+		fmt.Printf("%s, %s", token, dcid)
```
chejennifer, line 97: did you mean to keep this print statement here?

internal/store/files/recog_place_store.go
```diff
@@ -80,6 +82,25 @@ type RecogPlaceStore struct {
 	// recognized with their type preceding their name in the query. This is used
 	// to handle reconNames that are also common words like "to", "me", "a".
 	CommonWordReconNameToTypes map[string][]string
+	// US Zip Codes
+	USZipCodes map[string]struct{}
```
chejennifer, line 86: curious why not do this the same way as abbreviatedNameToPlaces where the value is the RecogPlaces object & instead of having a function, can just read from this map when trying to recognize a zipcode? Is it because of memory considerations?

chejennifer (commented):
nice! maybe also add a golden place recognition test case with both a zipcode and non-zipcode place in the sentence: https://github.com/datacommonsorg/mixer/blob/master/internal/server/recon/golden/recognize_places_test.go#L4

internal/store/files/recog_place_store.go
keyurva: No - removed.

internal/store/files/recog_place_store.go
keyurva: Yes - memory considerations. Since the function was trivial in this case, so did it this way. If there's more logic to it in the future, we can do it the other way.

keyurva:
> maybe also add a golden place recognition test case with both a zipcode and non-zipcode place in the sentence

That's a great idea. Done.

chejennifer (approved):
thanks for updating!

keyurva:
Thanks for the review!

PR #1434, opened 2024-10-10
Allow Combining a 5-token span to account for the formatting of the Google Maps predictions API.

gmechali:
Today, a 3 part place does not work appropriately, the 3rd containing part is considered its own location. Example https://screenshot.googleplex.com/3qcq4dB3zKp6Qte

The Google Maps prediction API being added for autocomplete uses a 3-part format for cities in countries that use states, US, Brazil, Argentina and more. 

With this change, we can support a query autocompleted such as:
https://screenshot.googleplex.com/4jnCABAFVTHNTUs

We intentionally did NOT account for "San Diego, California USA". The Google Maps API includes the commas so figured it would not be worth the code clutter to account for a comma ommission.

Tested locally + added a unit test.


internal/server/recon/recognize.go
```diff
@@ -446,24 +455,67 @@ func getNumSpansForContainedIn(spans []*pb.TokenSpans_Span, startIdx int) int {
...
 	dcidToRecogPlaces := map[string]*pb.RecogPlace{}
+	nextContainingPlaces := []string{}
 
 	for _, p1 := range startSpan.GetPlaces() {
 		for _, containingPlace := range p1.GetContainingPlaces() {
-			for _, p2 := range endSpan.GetPlaces() {
+			for _, p2 := range nextSpan.GetPlaces() {
 				if containingPlace == p2.GetDcid() {
 					dcidToRecogPlaces[p1.GetDcid()] = p1
+					nextContainingPlaces = p2.GetContainingPlaces()
```
chejennifer, line 487: since we can get into this if statement multiple times, should we be appending to the nextContainingPlaces instead of replacing it? or have it be a set?

internal/server/recon/recognize.go
```diff
@@ -446,24 +455,67 @@ func getNumSpansForContainedIn(spans []*pb.TokenSpans_Span, startIdx int) int {
...
+func getNextPlaceTokenSpan(spans []*pb.TokenSpans_Span, startIdx int) (*pb.TokenSpans_Span, int) {
+	for i := startIdx; i < len(spans); i++ {
+		if len(spans[i].GetPlaces()) > 0 {
+			return spans[i], i
+		}
+	}
+	return nil, -1
+}
+
+func findRecogPlaces(startSpan, nextSpan *pb.TokenSpans_Span) (map[string]*pb.RecogPlace, []string) {
```
chejennifer, line 478: nit: can we add a comment explaining what this function is doing since the name doesn't exactly explain it?

internal/server/recon/recognize.go
```diff
@@ -446,24 +455,67 @@ func getNumSpansForContainedIn(spans []*pb.TokenSpans_Span, startIdx int) int {
...
+	dcidToRecogPlaces, nextContainingPlaces := findRecogPlaces(startSpan, nextSpan)
+	
+	if numSpans == 5 {
+		nextNextSpan, _ := getNextPlaceTokenSpan(spans, nextSpanIndex+1)			
+		for _, possiblePlace := range nextContainingPlaces {
+			for _, actualPlace := range nextNextSpan.GetPlaces() {
+				if actualPlace.GetDcid() == possiblePlace {
+					for _, dcid := range startingDcids {
+						if val, ok := dcidToRecogPlaces[dcid]; ok {
+							val.ContainingPlaces = append(val.ContainingPlaces, possiblePlace)
```
chejennifer, line 516: I'm not sure if this will happen in practice, but could there be a case where there are 2 different starting places each with a different containing place that matched the next span and so in this case, we wouldn't want to add this last containing place to every starting dcid right?

maybe instead of having nextContainingPlaces be a list, it could be a map of startingPlace to nextContainingPlaces for that starting place.

chejennifer (commented):
very cool! can we also add a golden place recognition test case with maybe a 5 token containing place, 3 token containing place, and regular place all in one sentence: https://github.com/datacommonsorg/mixer/blob/master/internal/server/recon/golden/recognize_places_test.go#L4

internal/server/recon/recognize.go
```diff
@@ -429,6 +429,15 @@ func (p *placeRecognition) replaceTokensWithCandidates(tokens []string) *pb.Toke
 func getNumSpansForContainedIn(spans []*pb.TokenSpans_Span, startIdx int) int {
 	size := len(spans)
 
+	// Case: "place1, place2, place3". This case is required as Google Maps Prediction API meets this format.
+	if size > startIdx+4 && len(spans[startIdx+2].GetPlaces()) > 0 && len(spans[startIdx+4].GetPlaces()) > 0 {
+		nextSpanTokens := spans[startIdx+1].GetTokens()
+		nextNextSpanTokens := spans[startIdx+3].GetTokens()
+		if len(nextSpanTokens) == 1 && nextSpanTokens[0] == "," && len(nextNextSpanTokens) == 1 && nextNextSpanTokens[0] == "," {
```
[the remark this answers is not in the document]
gmechali: Done. I did it just for this method even though the "," is checked throughout this file to split the words into tokens too. I can make the change everywhere in the file if preferred.

internal/server/recon/recognize.go
gmechali: Yes i think you're right. I just changed to appending.

internal/server/recon/recognize.go
gmechali: Done.

internal/server/recon/recognize.go
gmechali: Yeah you're definitely right. Very very unlikely to happen, but also not impossible. I changed it to a map but actually it ended up doing so much iterating through loops. I ended up switching the map to be Containing Place DCID to StartingDCID and I think it's a bit cleaner that way.
So now I ll generate the map as: {country/USA: [geoId/1234, geoId/2345], Earth: [geoId/1234, geoId/2345],...}
So then we can check the nextNextSpan and if it has an entry in the map, we append the key in dcidToRecogPlace for all the starting DCIDs it has.

internal/server/recon/recognize.go
```diff
@@ -451,24 +461,78 @@ func getNumSpansForContainedIn(spans []*pb.TokenSpans_Span, startIdx int) int {
...
+func getNextPlaceTokenSpan(spans []*pb.TokenSpans_Span, startIdx int) (*pb.TokenSpans_Span, int) {
+	for i := startIdx; i < len(spans); i++ {
+		if len(spans[i].GetPlaces()) > 0 {
+			return spans[i], i
+		}
+	}
+	return nil, -1
+}
+
+// Generates a map of dcid to the corresponding RecogPlace objects found, and a second map containing the nextContainingPlaces to the starting DCIDs.
```
chejennifer, line 484: nit: can we also explain that nextContainingPlaces is not the containing place of the starting dcids but the containing place of the containing place of the starting dcids

internal/server/recon/recognize.go
```diff
@@ -451,24 +461,78 @@ func getNumSpansForContainedIn(spans []*pb.TokenSpans_Span, startIdx int) int {
...
+	spans []*pb.TokenSpans_Span, startIdx, numSpans int) *pb.TokenSpans_Span {
+	startSpan := spans[startIdx]
+	nextSpan, nextSpanIndex :=  getNextPlaceTokenSpan(spans, startIdx+1)
+
+	startingDcids := []string{}
+	for _, v := range startSpan.GetPlaces() {
+		startingDcids = append(startingDcids[:], v.GetDcid())
+	}
+
+	// This map is used to collect all the places for the combined span, with dedup.
```
chejennifer, line 521: nit: update comment to explain the two maps

internal/server/recon/recognize.go
```diff
@@ -451,24 +461,78 @@ func getNumSpansForContainedIn(spans []*pb.TokenSpans_Span, startIdx int) int {
...
+	dcidToRecogPlaces, nextContainingPlaceToStartDcid := findRecogPlaces(startSpan, nextSpan)
+	
+	if numSpans == 5 {
```
chejennifer, line 524: nit: add comment about the 5

internal/server/recon/recognize.go
```diff
@@ -451,24 +461,78 @@ func getNumSpansForContainedIn(spans []*pb.TokenSpans_Span, startIdx int) int {
...
+		nextNextSpan, _ := getNextPlaceTokenSpan(spans, nextSpanIndex+1)
+		for _, actualPlace := range nextNextSpan.GetPlaces() {
+			// Iterate through all possible places in the 3rd place span.
+			for possibleNextContaining, startDcids := range nextContainingPlaceToStartDcid {
```
chejennifer, line 528: since nextContainingPlaceToStartDcid is a map of nextContainingPlace to list of startDcids, can't we use actualPlace.GetDcid() to get the list of startDcids like:

```
startDcids, ok := nextContainingPlaceToStartDcid[actualPlace.getDcid()]
if !ok {
continue
}
for _, dcid := range startDcids {
...
}
```

internal/server/recon/recognize.go
```diff
@@ -451,24 +461,78 @@ func getNumSpansForContainedIn(spans []*pb.TokenSpans_Span, startIdx int) int {
...
+				// For all possible next containing places, we want to see if the starting DCID matches nextNextSpan's places.
```
chejennifer, line 529: confused about this comment, why are we trying to matching starting DCID to nextNextSpan?

chejennifer (commented):
thanks for updating!

internal/server/recon/recognize.go
gmechali: Think I made it a bit more clear now

internal/server/recon/recognize.go
gmechali: Done.

internal/server/recon/recognize.go
gmechali: Done.

internal/server/recon/recognize.go
gmechali: So I just got myself mixed up with all the loops. Your comment above allowed me to clean it up and now Im no longer mixed up :)
I was basically re-iterating through the places to find which one was the place we need to add, even though it's just actualPlace.getDcid(). Fixed that now. Comment reflects it.

internal/server/recon/recognize.go
gmechali: This comment helped a lot! This is what I meant to do when I reversed the key-val ordering in the map but then got mixed up and did it wrong. I was able to remove the two most inner loops and the equality constraint.
Should be good now, thanks!

internal/server/recon/recognize.go
```diff
@@ -451,24 +461,86 @@ func getNumSpansForContainedIn(spans []*pb.TokenSpans_Span, startIdx int) int {
...
+	for i := startIdx; i < len(spans); i++ {
+		if len(spans[i].GetPlaces()) > 0 {
+			return spans[i], i
+		}
+	}
+	return nil, -1
+}
+
+// Generates and returns two maps from the startSpan and nextSpan:
+// - StartSpan's DCID to the corresponding RecogPlace object.
```
chejennifer, line 485: nit: mention that startspan can correspond to multiple places
nit: DCIDs of nextSpan's ...

chejennifer (approved):
thanks for all the updates, jsut small comment nits

PR #1435, opened 2024-10-10
[fixit] Remove resolvedIds from /resolve v2 API

n-h-diaz:
The field has already been deprecated and should be removed, according to docs https://docs.datacommons.org/api/rest/v2/resolve


keyurva (commented):
Found some [usage](https://github.com/search?type=code&q=org%3Adatacommonsorg+%28resolvedIds+OR+resolved_ids%29+repo%3Adatacommonsorg%2Fimport) of it in the import repo.

Can we remove that as well?

PR #1437, opened 2024-10-14
Add another entry to recon_name_to_types.json

chejennifer:
the word "genetic" is a common word to use in queries and should not be recognized as is as an entity


beets (approved):
out of curiosity, how did we generate this list? some of the other keys are very generic.

chejennifer:
> out of curiosity, how did we generate this list? some of the other keys are very generic.

this list was generated by taking all the names of biomedical entities and putting them through a sentence classifier to check if they are generic parts of a sentence

PR #1440, opened 2024-10-22
Add docsite link in schema check error message.

hqpho:
Docsite was updated in https://github.com/datacommonsorg/docsite/pull/530.


keyurva (approved):
Thanks!

PR #1442, opened 2024-10-26
Add constants to StatisticalCalculations

n-h-diaz:
[Updated version of stale https://github.com/datacommonsorg/mixer/pull/1415]

This allows the schema to specify formulas with constant values (e.g. "(Count_Person * 2.5) + 1"). This is related to a One.org request to enable currency conversions in StatisticalCalculations.

StatisticalCalculations are still flag-protected by "cacheSVFormula" and are only enabled on autopush


internal/server/v2/observation/calculation_util.go
```diff
@@ -183,29 +201,106 @@ func evalBinaryExpr(
...
-	return result, nil
+	return &intermediateResponse{variableObs: result}, nil
+}
+
+// Combine two intermediateResponses using an operator token.
+func evalBinaryExpr(
+	x, y *intermediateResponse,
+	op token.Token,
+) (*intermediateResponse, error) {
+	if (x.variableObs != nil) && (y.variableObs != nil) {
```
keyurva, line 267: Does this assume that `x` and `y` will either include a `variableObs` or a `constantObs` but never both? And what happens if it does?

internal/server/v2/observation/calculation_util.go
```diff
@@ -183,29 +201,106 @@ func evalBinaryExpr(
...
+		return evalBinaryVariableObsExpr(x.variableObs, y.variableObs, op)
+	}
+	if (x.variableObs != nil) && (y.constantObs != nil) {
+		return evalBinaryVariableConstantNodeExpr(x.variableObs, y.constantObs, true /*iFirst*/, op)
+	}
+	if (x.constantObs != nil) && (y.variableObs != nil) {
+		return evalBinaryVariableConstantNodeExpr(y.variableObs, x.constantObs, false /*iFirst*/, op)
+	}
+	if (x.constantObs != nil) && (y.constantObs != nil) {
```
keyurva, line 276: I'm wondering if these 4 conditions can be abstracted into an enum - e.g. 

```
type BinaryExprType int

const (
	VarVar BinaryExprType = iota
	VarConst
	ConstVar
	ConstConst
)
```

It could potentially simplify things by not requiring 2 separate eval methods. We'd just need a trivial method to compute the enum value and then eval the intermediate responses based on it.

If it does simplify things then consider adopting that approach else ok to stay with this one.

keyurva (commented):
Very cool!

internal/server/v2/observation/calculation_util.go
```diff
@@ -26,6 +27,12 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+// An intermediateResponse can either have a VariableObservation or a constant.
+type intermediateResponse struct {
```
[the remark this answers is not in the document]
n-h-diaz: Good idea, updated to `intermediateObsResponse`

internal/server/v2/observation/calculation_util.go
n-h-diaz: They both shouldn't be set (since we only ever set one at a time), but we prefer `variableObs` over any `constantObs`. Added a comment clarifying this

internal/server/v2/observation/calculation_util.go
n-h-diaz: Sorry I was a bit confused what you mean to evaluate the intermediate response based on the enum value. I could try to combine `evalBinaryVariableObsExpr` and `evalBinaryVariableConstantNodeExpr` into one function (but they currently have different behavior). I updated some of the variable names/comments though to hopefully be more clear, but let me know if there's something in particular you'd like for this!

keyurva: Ah got it - the expressive names helped clarify things! Let's go with what you have.

keyurva (approved):

PR #1444, opened 2024-11-01
Create a basic spanner graph client.

keyurva:
* Creates a basic spanner client that connects to a cloud spanner graph.
* Implements a simple query that gets nodes by their IDs (does not fetch its edges yet).
* Setup a framework for spanner golden tests.
* Tests are currently no-op by default and require an `ENABLE_SPANNER_GRAPH` variable to be set to `true` for them to be executed.
  + This is to ensure it is only run on workstations of people actively developing against it and does not affect existing tests.


n-h-diaz (approved):
Thanks!

PR #1446, opened 2024-11-06
[Spanner Graph] Get node edges.

keyurva:
* Fetching edges instead of nodes inline with the new schema changes.
* Currently the edges are duplicated and that is reflected in the goldens. 
  + We should regenerate the goldens once that is fixed.


n-h-diaz (approved):
Thanks! I'll try to remove the duplicates

PR #1448, opened 2024-11-07
Add v3 wrapper of v2node

n-h-diaz:
This currently just calls V2Node and has a wrapper around the request/response message.

I also forked the tests from V2. 

Once we add the implementation, we can update the request/response and tests as needed


keyurva (commented):
Thanks! Any reason to fork tests vs create them as and when we add the implementations?

n-h-diaz:
> Thanks! Any reason to fork tests vs create them as and when we add the implementations?

This was mostly just to test that the wrapper was working properly. But I can remove the tests if you'd prefer!

keyurva:
> This was mostly just to test that the wrapper was working properly. But I can remove the tests if you'd prefer!

I'm assuming we'll replace the wrapper with calling a new merger that merges datasources soon, correct? If yes, then perhaps can we go with a no-op handler in this PR and start wiring with datasources in the next one?

Thoughts?

n-h-diaz:
> > This was mostly just to test that the wrapper was working properly. But I can remove the tests if you'd prefer!
> 
> I'm assuming we'll replace the wrapper with calling a new merger that merges datasources soon, correct? If yes, then perhaps can we go with a no-op handler in this PR and start wiring with datasources in the next one?
> 
> Thoughts?

Spoke offline: I removed the tests and can start filling in the API functionality in a follow up using the data sources (since that was submitted)

keyurva (approved):

PR #1449, opened 2024-11-07
Define a DataSource interface and a basic spanner impl.

keyurva:
* The `NodeResponse` does not include name and types currently.
  + Those require graph traversal and will be implemented later.

### DIFF
--- a/deploy/storage/spanner_graph_info.yaml
+++ b/deploy/storage/spanner_graph_info.yaml
@@ -1,3 +1,3 @@
 project: datcom-store
 instance: dc-kg-test
-database: dc_graph_2
+database: dc_graph_3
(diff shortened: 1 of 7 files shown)


n-h-diaz (approved):
Cool! 😃 

Just for my own understanding: is the eventual goal to have the API response constructed in this datasource layer? (Like filtering by property, pagination, etc) Or is this just to fetch the full nodes and then they'll be processed later?

keyurva:
I think the final response will need to be generated in a merger since it may need to merge responses from multiple sources.

Also, while `DataSource` is currently defined using the final request and response types, it does not need to be that way. If an intermediate type(s) makes more sense, we can go with that.

Basically, the `DataSource` hierarchy is completely internal and we can define methods and types that work best for us.

PR #1451, opened 2024-11-08
Use spanner data source in v3 node

n-h-diaz:
Uses spanner data source Node implementation for V3 node

This primarily links the V3 node implementation to use the DataSource interface (though currently only spanner is supported) and adds a sample test

* Currently this just reuses the merging logic from V2 (but allows more than two responses). I forked the merging code in case we want to modify it later, independently of V2 (for example, to handle pagination)
* This PR doesn't modify the spanner Node implementation, so the response is still incomplete (doesn't include properties, pagination, or the full EntityInfo). I'll start updating the functionality as a followup

### DIFF
--- a/internal/server/v3/node/golden/node_test.go
+++ b/internal/server/v3/node/golden/node_test.go
@@ -1,0 +1,92 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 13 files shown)


n-h-diaz:
Updated based on discussion today to make node part of data sources

internal/server/v3/node/golden/node_test.go
```diff
@@ -0,0 +1,94 @@
...
+	if !test.EnableSpannerGraph {
+		return
+	}
+	t.Parallel()
+	ctx := context.Background()
+
+	_, filename, _, _ := runtime.Caller(0)
+	goldenPath := path.Dir(filename)
+
+	testSuite := func(mixer pbs.MixerClient, latencyTest bool) {
```
keyurva, line 41: nit: Consider using just datasources vs the full mixer client for tests at this level.

keyurva (approved):
Thanks for the updates!

internal/server/v3/node/golden/node_test.go
n-h-diaz: I added this just to test the full e2e for mixer v3, but I'll add all other node tests on the datasource level (I'll put them in datasources + client specific datasource)

PR #1453, opened 2024-11-13
Fetch observations from spanner graph.

keyurva:
* Will wire it to the datasource in a follow-up PR.


internal/server/spanner/query.go
```diff
@@ -18,25 +18,86 @@ package spanner
...
+		FROM StatVarObservation
+		WHERE
+			variable_measured IN UNNEST(@variables) AND
+			observation_about IN UNNEST(@entities) AND
+			observation_date = @date AND
+			value != ''
+		ORDER BY observation_date ASC
+	`,
+		getSelectColumns(ObsColumns, "")),
+	getLatestObsByVariableAndEntity: fmt.Sprintf(`
```
n-h-diaz, line 80: I think this will return the latest observation overall, rather than per facet

keyurva: hmm - true. So in that case, we probably fetch them all from the DB and then process in mixer to return the latest per facet?

keyurva:
> How is the performance?

We need to measure it under load but it was acceptable when generating the goldens.

internal/server/spanner/query.go
n-h-diaz: That could work, or we could group by facet in the query

keyurva: Let me try the latter. Will fetch less data.

keyurva: Done. Let me know if this looks ok.

n-h-diaz (approved):
Thanks!

keyurva:
Thanks for the review!

PR #1455, opened 2024-11-14
Add support for arc direction to v3 node

n-h-diaz:
Adds base query for in-arcs

Still TODO for Node:
- SingleProp filtering
- Decorators
- BracketProps
- Filters
- Multiple hops

### DIFF
--- a/internal/server/spanner/query.go
+++ b/internal/server/spanner/query.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	"cloud.google.com/go/spanner"
+	v2 "github.com/datacommonsorg/mixer/internal/server/v2"
 	"github.com/datacommonsorg/mixer/internal/server/v2/shared"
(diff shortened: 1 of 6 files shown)


internal/server/spanner/query.go
```diff
@@ -58,17 +60,29 @@ var statements = struct {
...
 			n.types) object
 	ON
 		edge.object_id = object.subject_id
 	WHERE
 		edge.subject_id IN UNNEST(@ids)
 	`,
+	getEdgesByObjectID: `
+	GRAPH DCGraph MATCH (n:Node)-[e:Edge
+	WHERE
+		e.object_id IN UNNEST(@ids)
```
keyurva, line 77: Have we created an index by object_id? Any difference in performance between the 2 queries?

keyurva (approved):
Nice!

PR #1458, opened 2024-11-18
Fetch observations modeled as a time series.

keyurva:


internal/server/spanner/model.go
```diff
@@ -32,17 +41,77 @@ type Edge struct {
...
+	MeasurementMethod string     `spanner:"measurement_method"`
+	Unit              string     `spanner:"unit"`
+	ScalingFactor     string     `spanner:"scaling_factor"`
+	ImportName        string     `spanner:"import_name"`
+	ProvenanceURL     string     `spanner:"provenance_url"`
+}
+
+type DateValue struct {
+	Date  string
+	Value float64
```
n-h-diaz, line 60: We also have non-numeric values for statType=measurementResult stat vars, so maybe we can keep this as a string?

internal/server/spanner/model.go
```diff
@@ -32,17 +41,77 @@ type Edge struct {
...
+				Date:  date,
+				Value: floatVal,
+			})
+		}
+	}
+	return nil
+}
+
+// FilterByDate filters observations based on the given date.
+func (ts *TimeSeries) FilterByDate(date string) {
```
n-h-diaz, line 97: Just to clarify, are you still planning on caching the entire series? If so, maybe the date filtering be done separately from fetching the full series in GetObservations. Or if not, maybe we can only fetch the requested dates. Wdyt?

n-h-diaz (commented):
Thanks! It seems more compact now :)

internal/server/spanner/model.go
keyurva: I think the api currently only supports floats: https://github.com/datacommonsorg/mixer/blob/3518530571e37f902117e5380f55fdf4930e93c1/proto/stat.proto#L69

So ok to go with this (eventually it gets converted to float anyway) or should we still keep it as string?

internal/server/spanner/model.go
keyurva: Good point! Yes, caching eventually. Removed filtering logic from this layer. Will apply in the datasource layer in the next PR.

keyurva: For clarity, I've kept this method but removed calling it from query.go. Will instead call it from the datasource in the next one.

internal/server/spanner/model.go
n-h-diaz: Oh I see! Yep I think it's fine to keep as a float for now, and then maybe we can revisit later to support non-numeric (I think Ajai mentioned some partner request for this)

internal/server/spanner/model.go
n-h-diaz: Sounds good! Just make sure the datasource caller can handle the case when there's no date match (so shouldn't return anything)

n-h-diaz (approved):
Thanks!

internal/server/spanner/model.go
keyurva: Ack.

PR #1466, opened 2024-11-22
Remove translate endpoint.

keyurva:


dwnoble:
Could you make this return a HTTP 410 "GONE" instead of removing it enitrely?

keyurva:
> Could you make this return a HTTP 410 "GONE" instead of removing it enitrely?

Any reason to do that? It's not receiving any traffic. And be removing it, it will still return a 4XX like with any other non-existent endpoint.

dwnoble (approved):

keyurva:
Thanks for the comments!

PR #1469, opened 2024-11-23
Data release 20241123

hareesh-ms:


internal/server/v1/page/golden/bio_page/chlamydia.json
```diff
@@ -2,14978 +2,6 @@
   "nodes": [
     {
       "neighbors": [
-        {
```
hareesh-ms, line 5: @spiekos : Is these deletions expected ? Many of the [DiseaseGeneAssociation](https://datacommons.org/browser/DiseaseGeneAssociation) nodes seems to be deleted in the autopush version.

ajaits (approved):

PR #1470, opened 2024-11-25
Adds parameterization to the BigQuery SQL Execution

gmechali:
Modifies the GetSql method to return the parameters in a map, changes the execution to add the params.
This affects both the Node and Query endpoints.
The Translate endpoint is being removed as part of this PR since we will no longer support that page.

Also updates all the tests to validate the Parameter Map along with the parameterized SQL.


dwnoble (approved):
Thanks Gabe!

PR #1475, opened 2024-12-11
Add support for observations contained in place.

keyurva:
* I think the expression support can be extended / generalized to support more types of expressions but sticking with V2 API equivalency for now so only supporting contained in place.


internal/server/spanner/datasource.go
```diff
@@ -68,15 +68,31 @@ func (sds *SpannerDataSource) Node(ctx context.Context, req *v3.NodeRequest) (*v
...
-	if err != nil {
-		return nil, fmt.Errorf("error getting observations: %v", err)
+	var observations []*Observation
+
+	if entityExpr != "" {
+		containedInPlace, err := v2.ParseContainedInPlace(entityExpr)
+		if err != nil {
+			return nil, fmt.Errorf("error getting observations (contained in): %v", err)
+		}
+		observations, err = sds.client.GetObservationsContainedInPlace(ctx, variables, containedInPlace)
```
n-h-diaz, line 81: Were there any performance concerns with fetching the full series for all the children places?

n-h-diaz (approved):
Thanks for adding!

internal/server/spanner/datasource.go
keyurva: For the golden tests it was fine. For something like "Count_Person" and "country/USA<-containedInPlace+{typeOf:State}" it was noticeable.

keyurva:
Thanks for the review!

PR #1479, opened 2024-12-17
Implement node search api.

keyurva:


n-h-diaz (approved):
Thanks for adding! Is the order of results deterministic?

keyurva:
> Thanks for adding! Is the order of results deterministic?

Good question - probably something we can ask Mingtian? My sense is that since we are ordering them by score, for a given snapshot of the schema, the results should be deterministic.

PR #1480, opened 2024-12-18
Update spanner queries

n-h-diaz:
* Updates node queries to use only GQL. This helps improve performance for querying with filters
* Removes extra join in observation with contained in place query. After speaking with Mingtian, he recommends adding observation to the graph to remove additional joins. This will require updating the graph schema, and there are some implications for write throughput, but we can potentially explore this (can discuss more later)
* Updates some search tests (it's unclear if this is just flaky, so if it starts to fail again, we can consider updating the test to support potentially inconsistent ordering)

### DIFF
--- a/internal/server/spanner/golden/query/search_nodes_with_type.json
+++ b/internal/server/spanner/golden/query/search_nodes_with_type.json
@@ -350,6 +350,13 @@
     ]
   },
   {
+    "SubjectID": "oecd/IDD_INCACTOTAL_TOT_CURRENT_METH2011",
+    "Name": "Income Distribution Database: All age groups: mean disposable income (current prices), Total population, Current definition, Income definition until 2011",
+    "Types": [
+      "StatisticalVariable"
+    ]
+  },
+  {
     "SubjectID": "oecd/IDD_INCCTOTAL_WA_INCOMPARABLE_METH2012",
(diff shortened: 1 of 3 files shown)


internal/server/spanner/query.go
```diff
@@ -58,193 +58,144 @@ var statements = struct {
...
-		DISTINCT subject_id,
-		predicate
-	FROM 
-		Edge
-	WHERE
-		subject_id IN UNNEST(@ids)
-	ORDER BY
-		subject_id,
-		predicate
+		GRAPH DCGraph MATCH -[e:Edge
```
keyurva, line 61: Do we have any numbers (or a sense for) on the performance improvements with GQL?

keyurva (approved):
Thanks - excited about the perf improvements!

Let's chat about observations in the graph.

internal/server/spanner/golden/query/search_nodes_with_type.json
```diff
@@ -349,6 +349,13 @@
       "StatisticalVariable"
     ]
   },
+  {
+    "SubjectID": "oecd/IDD_INCACTOTAL_TOT_CURRENT_METH2011",
```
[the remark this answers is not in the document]
n-h-diaz: He said he thinks it's consistent but will confirm and get back to us. He also suggested doing a hybrid search (see the question notes doc)

internal/server/spanner/query.go
n-h-diaz: It should speed up queries with filters significantly

PR #1481, opened 2024-12-20
Return names and actual types of properties.

keyurva:
* The current query that fetches triples is executed first, followed by a query to get names and types of all dcids in the triples response.
* Those are then used to populate the names and types of properties in the response.
* Note that if properties reference entities in base DC, those will not be fetched and will continue to be returned as they currently are. 
   + If we need their names / types to also be fetched, that will have to be done via a separate remote mixer call.
   + We can implement this in a separate PR if necessary.

### DIFF
--- a/internal/server/v1/propertyvalues/core.go
+++ b/internal/server/v1/propertyvalues/core.go
@@ -33,6 +33,14 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+const (
+	namePredicate   = "name"
+	typeOfPredicate = "typeOf"
+	subjectIdColumn = "subject_id"
+	objectIdColumn  = "object_id"
+	defaultType     = "Thing"
+)
+
 // Fetch is the generic handler to fetch property values for multiple
 // properties and nodes.
 //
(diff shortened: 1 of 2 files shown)


internal/server/v1/propertyvalues/core.go
```diff
@@ -107,11 +115,76 @@ func fetchSQL(
...
+	// Get triples for the specified nodes and properties.
+	triples, err := executeTriplesSQL(sqlClient, nodes, properties, matchColumn)
+	if err != nil {
+		return nil, err
+	}
+
+	// Collect all dcids from the triples and get their entity infos (name and type).
+	// NOTE: This will only fetch info on entities that are in the SQL database.
+	// If any dcids reference entities in base DC - those will not be fetched.
+	// If we want them as well, we'll need to make a remote mixer call to fetch them.
```
dwnoble, line 132: I think this will be important to add, but at the same time i believe the logic from this PR should resolve UN's issue

dwnoble: Maybe this can happen in bulk before the existing mixer merge stage?

dwnoble (approved):

internal/server/v1/propertyvalues/core.go
keyurva: Ack. Yeah, doing it at that stage is possible but could get hairy.

PR #1482, opened 2024-12-30
Define /v3/resolve

n-h-diaz:
Some differences between v3 and v2:
* Doesn't include resolved_ids property, which looks to be deprecated (https://docs.datacommons.org/api/rest/v2/resolve.html). I can add this later if it's needed 
* Returns the full EntityInfo for a candidate. This is just for consistency across responses, and it should be easy to retrieve the full EntityInfo from the Node table

Let me know if this seems okay!


proto/v3/resolve.proto
```diff
@@ -0,0 +1,47 @@
...
+//   property = "<-description->dcid" or "<-description{typeOf:City}->dcid"
+message ResolveRequest{
+    repeated string nodes = 1;
+    string property = 2;
+}
+
+message ResolveResponse{
+    message Entity {
+        string node = 1;
+        repeated datacommons.EntityInfo candidates = 2;
```
keyurva, line 44: I like the consistency that `EntityInfo` brings. One concern though is that it will break code like [this one](https://github.com/datacommonsorg/import/blob/43dbfac4e0a59d3498e201c4574cff9fa2b97f05/simple/util/dc_client.py#L143) that looks for the current fields.

Any thoughts on how we can address this? One approach could be to deprecate the current fields / messages and add new ones that use `EntityInfo`. Or another approach?

n-h-diaz: Good point - discussed offline and will use the v2 proto for now (and can maybe migrate later if needed)

keyurva (approved):

PR #1483, opened 2024-12-30
Only check top 20 matches for v3 node search tests

n-h-diaz:
The order of lower matches can be nondeterministic, so this is just to help make tests less flaky


internal/server/v3/nodesearch/golden/node_search_test.go
```diff
@@ -63,19 +63,25 @@ func TestV3NodeSearch(t *testing.T) {
 				t.Errorf("Could not run V3NodeSearch: %s", err)
 				continue
 			}
+
+			// Filter resp to top 20 matches to avoid flaky low matches.
+			topResp := &pbv3.NodeSearchResponse{
+				Nodes: resp.Nodes[:20],
```
keyurva, line 69: nit: Consider declaring the limit as a constant.

keyurva (approved):

internal/server/v3/nodesearch/golden/node_search_test.go
n-h-diaz: Good idea, done

PR #1485, opened 2025-01-03
Fix nil references to sql DB.

keyurva:
* Maintains SQLClient in the store by value instead of by reference. This avoids NPEs with the underlying DB handle.
* We'll switch back to pointer reference once the DB handle in the client is made internal.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -202,23 +202,25 @@ func main() {
 	}
 
 	// SQL client
-	var sqlClient *sqldb.SQLClient
+	var sqlClient sqldb.SQLClient
 	if *useSQLite {
-		sqlClient, err = sqldb.NewSQLiteClient(*sqlitePath)
+		client, err := sqldb.NewSQLiteClient(*sqlitePath)
 		if err != nil {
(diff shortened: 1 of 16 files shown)


clincoln8 (approved):
Thanks Keyur! 

Were you able to run the test to ensure this change allows them to pass? If so, how did you do it?

keyurva:
> Were you able to run the tests to ensure this change allows them to pass? If so, can you add a note on how you ran them?

Unfortunately not. It would require some setup - start a local base DC mixer, start a local custom DC instance by configuring it to point to the first mixer and then run the tests by modifying the [custom dc url](https://github.com/datacommonsorg/website/blob/e1f1400ca9a138b40f1e2dfec90d199cf1cc5bcd/server/webdriver/cdc_tests/base.py#L16) used by tests.

In this case, I ran a base DC local mixer and tested some of the calls that were failing in the autopush mixer. Would be more definitive if we can simplify running the CDC tests locally fore sure. With the current approach, it'll be a wait and watch until the CDC tests run on autopush :(

PR #1486, opened 2025-01-03
Encapsulate observations query in sql client.

keyurva:
* Encapsulates the observations query in sql client.
* It also brings in the sqlx library that allows us to do parameterized named queries and model the sql client in line with the spanner client.


n-h-diaz (approved):
Thanks!

PR #1488, opened 2025-01-06
Encapsulate sv summary query in sql client.

keyurva:


internal/server/v2/resolve/golden/resolve_description_test.go
```diff
@@ -59,7 +59,6 @@ func TestResolveDescription(t *testing.T) {
 					"Illes Balears,Spain",
 					"Ciudad Autónoma de Ceuta,Spain",
 					"Ciudad Autónoma de Melilla,Spain",
-					"Hauts-de-France,France",
```
keyurva, line 62: @n-h-diaz - Removed this location that was causing a golden test failure in the cloud. Let me know if there's a better way to address this.

https://screenshot.googleplex.com/7F8j2t8V3VWYLff

n-h-diaz: I think this is okay for now - this might be due to a placeId change, but I'll need to look more into the specific failure to confirm

n-h-diaz (approved):
Thanks!

PR #1491, opened 2025-01-08
Use the same protoc version (3.21.12) everywhere

hqpho:
Re-generate proto files with version 3.21.12 and recommend installing that version. This way generating protos doesn't result in local diffs.

I've verified this is the same version used by the Docker images for mixer push as well as custom DC services:
 - `docker run -it --entrypoint /bin/bash -e DEBUG=true gcr.io/datcom-ci/go-protoc:latest` -> `protoc --version`
 - `docker run -it  --entrypoint /bin/bash -e DEBUG=true gcr.io/datcom-ci/datacommons-services:latest` -> `apt show protobuf-compiler`

The website repo dev guide will also need to be updated to recommend this version (it currently recommends v3.21.9).


docs/developer_guide.md
```diff
@@ -8,7 +8,8 @@
 
   - [`gcloud`](https://cloud.google.com/sdk/docs/install)
   - [`Golang`](https://golang.org/doc/install)
-  - [`protoc`](http://google.github.io/proto-lens/installing-protoc.html)
+  - [`protoc`](http://google.github.io/proto-lens/installing-protoc.html) at version 3.21.12
+    - If using Homebrew, run `brew install protobuf@21` and be sure to update your path as described in the output.
```
keyurva, line 12: nit: Perhaps add a hint for what to look for in the output. For me it was:

```
If you need to have protobuf@21 first in your PATH, run:
  echo 'export PATH="/opt/homebrew/opt/protobuf@21/bin:$PATH"' >> ~/.zshrc
```

keyurva (approved):

PR #1492, opened 2025-01-08
Update order of V3Node response

n-h-diaz:
Update: this mostly just sorts the Spanner response, so property requests will be in alphabetical order


internal/server/spanner/dsutil.go
```diff
@@ -52,33 +54,56 @@ func nodePropsToNodeResponse(propsBySubjectID map[string][]*Property) *v3.NodeRe
 	return nodeResponse
 }
 
-// nodeEdgesToNodeResponse converts a map from subject id to its edges to a NodeResponse proto.
-func nodeEdgesToNodeResponse(edgesBySubjectID map[string][]*Edge) *v3.NodeResponse {
-	nodeResponse := &v3.NodeResponse{
-		Data: make(map[string]*v2.LinkedGraph),
+// nodeEdgesToNodeResponse converts a map of subject->predicate->Edges to a NodeResponse proto.
+func nodeEdgesToNodeResponse(edges map[string]map[string][]*Edge, nodes []string, bracketProps []string) *pbv3.NodeResponse {
```
keyurva, line 58: Is there a way to avoid multilevel maps here? Does a `SubjectPredicate` struct work as the key with a single level `map[SubjectPredicate][]*Edge`?

n-h-diaz: Good point!

After playing with some example though, I realized that the order mostly doesn't matter (proto maps are unordered, and so with enough tries I was able to get v2 also to change order), so was able to revert a lot of these changes.

The main "order" comes from repeated fields, which in this case is just the properties, which can be set directly in GQL

internal/server/spanner/query.go
```diff
@@ -104,6 +107,11 @@ var statements = struct {
 			e.provenance,
 			'' AS name,
 			ARRAY<STRING>[] AS types
+		ORDER BY
```
keyurva, line 110: This is a simple smart solution - thanks!

keyurva (approved):

PR #1493, opened 2025-01-08
Encapsulate entity count, pv and svg check queries in sql client.

keyurva:


internal/sqldb/statements.go
```diff
@@ -131,4 +135,37 @@ var statements = struct {
...
+		GROUP BY
+				variable,
+				date,
+				provenance
+		ORDER BY
+				variable,
+				date,
+				provenance;
+	`,
+	getSubjectPredicates: `
```
n-h-diaz, line 158: Do these need to be distinct?

keyurva: Good call. I was using the statements from before but distinct makes sense here. Updated.

n-h-diaz (approved):

PR #1494, opened 2025-01-09
Add select to v3 observation

n-h-diaz:
This enables selecting by just entity/variable (ie. existence checks) and by facet

Also adds facetId to response since it was missing before

Follow up: fix the order of observation response to match v2


internal/server/spanner/datasource.go
```diff
@@ -67,11 +76,29 @@ func (sds *SpannerDataSource) Node(ctx context.Context, req *v3.NodeRequest) (*v
 }
 
 // Observation retrieves observation data from Spanner.
-func (sds *SpannerDataSource) Observation(ctx context.Context, req *v3.ObservationRequest) (*v3.ObservationResponse, error) {
+func (sds *SpannerDataSource) Observation(ctx context.Context, req *pbv3.ObservationRequest) (*pbv3.ObservationResponse, error) {
+	var queryDate, queryValue, queryVariable, queryEntity, queryFacet bool
```
keyurva, line 80: nit and optional: As a good practice, consider encapsulating multiple attributes like these into a struct. For ex.

```
type queryOptions struct {
  queryDate bool
 ...
}
```

This allows us to have named utility methods like `func selectFieldsToQueryOptions(selectFields []string) queryOptions`, `func (qo *queryOptions) queryObs() bool`, etc.

internal/server/spanner/datasource.go
```diff
@@ -83,23 +110,20 @@ func (sds *SpannerDataSource) Observation(ctx context.Context, req *v3.Observati
...
 		if err != nil {
 			return nil, fmt.Errorf("error getting observations: %v", err)
 		}
-		observations = obs
 	}
 
 	observations = filterObservationsByDate(observations, date)
 
-	return observationsToObservationResponse(variables, observations), nil
+	queryObs := queryDate && queryValue
```
keyurva, line 121: Curious to know what's the use case for not selecting dates and what do we return in that case.

keyurva (approved):
Thanks!

internal/server/spanner/datasource.go
n-h-diaz: Good idea, thanks! Done

internal/server/spanner/datasource.go
n-h-diaz: Spoke offline - it's a bit confusing in v2, but both date and value are required (if one's missing it'll be treated as an existence check) 

Also, we decided to remove the checks for variable and entity in v3 to give more flexibility to user (since it's always considered true)

PR #1495, opened 2025-01-09
Fix error due to value being returned as byte array by mysql.

keyurva:
* GROUP_CONCAT values are returned as strings by sqlite but as byte arrays by mysql (cloud sql).
* This led to errors (see image below).
* This was not caught by golden tests since they use sqlite.

image

### DIFF
--- a/internal/sqldb/model.go
+++ b/internal/sqldb/model.go
@@ -54,10 +54,17 @@ func (s *StringSlice) Scan(src interface{}) error {
 		return nil
 	}
 
-	val, ok := src.(string)
-	if !ok {
-		return fmt.Errorf("failed to decode []string: (%v)", src)
+	var val string
+
+	switch v := src.(type) {
+	case []byte:
+		val = string(v)
+	case string:
+		val = v
+	default:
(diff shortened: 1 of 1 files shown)


hqpho:
Which tests should have caught this (do you have a link handy) and is it possible to also run them against CloudSQL?

hqpho (approved):

keyurva:
> Which tests should have caught this (do you have a link handy) and is it possible to also run them against CloudSQL?

I did it in a roundabout way. (Unfortunately it's not straightforward today.):

- looked at the autopush test logs. It was a weird 404 error with a browser SV hierarchy test.
- then looked at the dc-autopush instance logs and that showed the error in the screenshot above.
- then looked at the [website code](https://github.com/datacommonsorg/website/blob/bd458fa4c634f1756cfb1c99f206e150ba1c14d0/server/services/datacommons.py#L298C26-L298C48) to figure out which mixer call it made.
- then made the fix and tested it by connecting to the cloud sql that autopush connects to from my local mixer.

The Cloud SQL part makes it tricky but we should look into whether our golden tests can connect to a cloud sql instance in addition to local sqlite db files.

PR #1498, opened 2025-01-13
Add support for FacetFilter in V3 Observation

n-h-diaz:


internal/util/util.go
```diff
@@ -521,6 +523,42 @@ func GetFacetID(m *pb.Facet) string {
...
 
+func ShouldIncludeFacet(filter *pbv2.FacetFilter, facet *pb.Facet) bool {
+	if filter == nil {
+		return true
+	}
+
+	facetID := GetFacetID(facet)
+	if filter.FacetIds != nil {
+		matchedFacetId := false
+		for _, facetId := range filter.FacetIds {
```
keyurva, line 534: Since facet IDs are ephemeral, curious to know if users actually make requests with facet IDs in the request?

keyurva (approved):

internal/util/util.go
n-h-diaz: I think this likely would only be an internal use case. The ids should be deterministic though, so if a user does one request and gets a facet id in the response, they should be able to reuse it in later requests. 

(Maybe for later, but I think it could be helpful to eventually make facet ids into proper dcids) 

Thanks for reviewing!

PR #1499, opened 2025-01-14
Check presence of tables instead of creating them.

keyurva:
* Checks presence of tables at startup and fails if they are not.
* Folds this functionality into the sql client as part of the larger sql client refactor.
* For sqlite, does not create any files or directories if the db file is not found, instead returns an error.
* For cloud sql, if credentials were not valid, it used to fail on the first query to the db instead of when creating the client itself. It now fails when creating the client.


internal/sqldb/query.go
```diff
@@ -424,6 +427,93 @@ func (sc *SQLClient) queryAndCollect(
...
+		columnNames, err := sc.getTableColumns(tableName)
+		// We use an error or empty column names as a signal that tables are not found.
+		// We are not querying the db schema itself because those queries are different for different dbs.
+		// This keeps it simple and agnostic of the underlying db.
+		if err != nil || len(columnNames) == 0 {
+			if err != nil {
+				log.Printf("Error checking table %s: %v", tableName, err)
+			}
+
+			errMsg := fmt.Sprintf(`The SQL database does not have the required tables.
```
keyurva, line 459: @kmoscoe @hqpho - this is the log message when required tables are not found. Let me know if a different log or link works better.

kmoscoe: This seems fine, but what is the suggested remedy?  Does the schemaupdate mode work for it?

keyurva: Not sure if just schemaupdate would work here (@hqpho?). They would need to load their data using the data container.

internal/sqldb/tables.go
```diff
@@ -0,0 +1,50 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sqldb
+
+// SQL table and column names.
```
hqpho, line 17: Does the order of these matter? (Does it have to match the order they were created in somehow?)

internal/sqldb/query.go
```diff
@@ -424,6 +427,93 @@ func (sc *SQLClient) queryAndCollect(
...
+}
+
+func (sc *SQLClient) getTableColumns(tableName string) ([]string, error) {
+	query := fmt.Sprintf(statements.getTableColumnsFormat, tableName)
+
+	rows, err := sc.dbx.Query(query)
+	if err != nil {
+		return nil, fmt.Errorf("table not found: %s (error: %w)", tableName, err)
+	}
+	defer rows.Close()
```
hqpho, line 506: Super nit: Capitalize error messages?

internal/sqldb/client.go
```diff
@@ -99,30 +98,17 @@ func newSQLClient(db *sql.DB, driver string) *SQLClient {
...
-		}
-		return sqlClient, nil
-	}
-	if !os.IsNotExist(err) {
-		return nil, err
-	}
-	_, err = os.Create(dbPath)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("error accessing sqlite db file: %s (%w)", dbPath, err)
```
hqpho, line 103: Same super nit re capitalization

keyurva: Same golang quirk.

internal/sqldb/query.go
keyurva: Good to know either will work. We can document the approach that's most intuitive for the user.

internal/sqldb/query.go
keyurva: This is a golang quirk (which I don't like) that [error strings should not be capitalized](https://google.github.io/styleguide/go/decisions.html#error-strings). The IDE flags a style warning if you do.

internal/sqldb/sqlquery/statvar_summary_test.go
```diff
@@ -29,9 +29,9 @@ func TestGetStatVarSummaries(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Could not open testing database: %v", err)
 	}
-	err = sqldb.CheckSchema(sqlClient.DB)
+	err = sqlClient.CheckTablesAndSchema()
```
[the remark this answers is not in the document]
keyurva: I love that name and message - changed.

internal/sqldb/statements.go
```diff
@@ -223,4 +224,7 @@ var statements = struct {
 		WHERE entity in (:entities)
 		GROUP BY entity;
 	`,
+	getTableColumnsFormat: `
```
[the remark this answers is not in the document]
keyurva: Done.

internal/sqldb/tables.go
keyurva: No, order does not matter.

internal/sqldb/tables.go
```diff
@@ -0,0 +1,50 @@
...
+	ColumnMeasurementMethod = "measurement_method"
+	ColumnObservationPeriod = "observation_period"
+	ColumnProperties        = "properties"
+)
+
+// allTables is an array of all expected tables in the SQL database.
+var allTables = []string{TableObservations, TableTriples, TableKeyValueStore}
+
+// allObservationsTableColumns is a set of all column names in the observations table.
+var allObservationsTableColumns = map[string]struct{}{
```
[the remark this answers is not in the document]
keyurva: A few things:

* The original intent was to use a set. But golang does not have one and using a `map[type]struct{}` is the conventional substitute.
* This set is used for finding missing columns using the [util.GetMissingStrings](https://github.com/datacommonsorg/mixer/blob/2eab7f6e2bc3e8b13f58d443f1317cf41d4f8af4/internal/util/util.go#L735) method which expects a set.

Having said all that, I changed it to an array after all! Reason: I realized that `GetMissingStrings` is not doing a set lookup but just iterating over it. And an array is more logical in that case.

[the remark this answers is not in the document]
hqpho: Thank you for the explanation!

hqpho (approved):

PR #1500, opened 2025-01-15
Add remote datasource

n-h-diaz:
Right now I only added it when "useSpannerGraph" is also true, to avoid affecting other instances while still in development.

I wasn't sure the best way to test this. One options is to create the TypeMock DataSource and use that in place of TypeRemote in tests, but this wouldn't really test the remote mixer (but instead could help test merging behavior). Let me know if you have any ideas or suggestions! 

This PR also adds the wrapper for V3 resolve, but this currently doesn't return anything.


cmd/main.go
```diff
@@ -265,6 +266,17 @@ func main() {
 		}
 	}
 
+	// Remote Mixer.
+	// TODO: remove flag for Spanner Graph after validating and adding merging.
+	if *useSpannerGraph && *remoteMixerDomain != "" {
```
keyurva, line 271: What do you think about an `enableV3` flag to test with different combinations of data sources?

internal/server/remote/client.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"fmt"
+	"net/http"
+
+	pbv3 "github.com/datacommonsorg/mixer/internal/proto/v3"
+	"github.com/datacommonsorg/mixer/internal/server/resource"
+	"github.com/datacommonsorg/mixer/internal/util"
+)
+
+// RemoteClient encapsulates the Remote client.
```
keyurva, line 27: nit: consider making a mention of a remote mixer for clarity.

internal/server/remote/client.go
```diff
@@ -0,0 +1,75 @@
...
+type RemoteClient struct {
+	metadata   *resource.Metadata
+	httpClient *http.Client
+}
+
+// NewRemoteClient creates a new RemoteClient.
+func NewRemoteClient(metadata *resource.Metadata, httpClient *http.Client) (*RemoteClient, error) {
+	if metadata.RemoteMixerDomain == "" || metadata.RemoteMixerAPIKey == "" {
+		return nil, fmt.Errorf("error creating remote client")
```
keyurva, line 36: Consider adding a reason that can help users fix it - something about remote url and key must be specified.

internal/server/remote/client.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	return &RemoteClient{metadata, httpClient}, nil
+}
+
+func (rc *RemoteClient) Node(req *pbv3.NodeRequest) (*pbv3.NodeResponse, error) {
+	resp := &pbv3.NodeResponse{}
+	err := util.FetchRemote(rc.metadata, rc.httpClient, "/v3/node", req, resp)
```
keyurva, line 43: Wondering if these (except the new end points) should be to `/v2`?

keyurva (commented):
This is very well done!

> I wasn't sure the best way to test this

Could using the `enableV3` approach I mentioned in a comment help with this?

cmd/main.go
n-h-diaz: Added this flag for each datasource

internal/server/remote/client.go
n-h-diaz: Done

internal/server/remote/client.go
n-h-diaz: Done

internal/server/remote/client.go
n-h-diaz: I switched these to V2 (so had to add some conversion between V2 and V3 protos). If we do this though, we should just be careful that V3 response aligns with expected V2 response so they can be merged

n-h-diaz:
> This is very well done!
> 
> > I wasn't sure the best way to test this
> 
> Could using the `enableV3` approach I mentioned in a comment help with this?

I added the flag and enabled in autopush. For testing, it looks like remote mixer currently is tested either locally or by mocking results (since it requires an API key). I added a test for merging node and can continue to add more as we add more merging functionality. Let me know if you have any suggestions for how I should better test this though!

internal/server/remote/client.go
```diff
@@ -0,0 +1,114 @@
...
+// NewRemoteClient creates a new RemoteClient.
+func NewRemoteClient(metadata *resource.Metadata) (*RemoteClient, error) {
+	if metadata.RemoteMixerDomain == "" || metadata.RemoteMixerAPIKey == "" {
+		return nil, fmt.Errorf("error creating remote client: please ensure that a remote mixer domain and API key are set")
+	}
+	return &RemoteClient{metadata, &http.Client{}}, nil
+}
+
+func (rc *RemoteClient) Node(req *pbv3.NodeRequest) (*pbv3.NodeResponse, error) {
+	v2Req := &pbv2.NodeRequest{
```
keyurva, line 43: nit and optional: Consider using helper methods for doing conversions.

keyurva:
> > This is very well done!
> > > I wasn't sure the best way to test this
> > 
> > 
> > Could using the `enableV3` approach I mentioned in a comment help with this?
> 
> I added the flag and enabled in autopush. For testing, it looks like remote mixer currently is tested either locally or by mocking results (since it requires an API key). I added a test for merging node and can continue to add more as we add more merging functionality. Let me know if you have any suggestions for how I should better test this though!

Hmm. Wondering if a cloudbuild that runs at a regular cadence is something to consider here. We can expand the scope as an overall V3 sanity test. 

Adding @hqpho - if they have any suggestions.

[All of this is of course outside the scope of this PR and we can take the discussion offline if needed.]

keyurva (approved):

internal/server/remote/client.go
n-h-diaz: Good idea, done

PR #1501, opened 2025-01-16
Data release 20250116

ajaits:


internal/server/v0/triple/golden/get_triples/observation.json
```diff
@@ -67,39 +67,5 @@
       "objectName": "Total Population",
       "provenanceId": "dc/base/CensusPEP"
     }
-  ],
-  "dc/o/kyv7dxe4s18eh": [
```
ajaits, line 71: This has changed to dc/o/48rnjqg9212e1 with the same value.
https://autopush.datacommons.org/browser/dc/o/48rnjqg9212e1

hareesh-ms (approved):

PR #1503, opened 2025-01-17
Updates to V3 observation

n-h-diaz:
This includes: 
- Ranking facets in the "preferred" order
- Updates to the response structure. This is primarily to match V2 to help with merging and backward compatibility. V2 is a bit inconsistent with the responses for different request patterns, particularly for missing data and containedInPlace (likely due to using different BT caches for each), so for this PR I similarly handled the observation, facet, and existence response constructions differently. 

Going forward, I think it would be nice to unify these responses more, which will simplify the V3 implementation (and is possible with Spanner!) but I did see some website usage that looks like it could depend on the current structure, so didn't want to change anything now.

As a follow up, I'll add datasource merging and series limits, and then (I think?) observation should be complete. (I'll test this more though)


internal/server/spanner/dsutil.go
```diff
@@ -132,47 +137,30 @@ func queryObs(qo *queryOptions) bool {
...
-
-		response.ByVariable[variable].ByEntity[entity].OrderedFacets = append(
-			response.ByVariable[variable].ByEntity[entity].OrderedFacets,
-			facetObservation,
-		)
-		response.Facets[facetId] = facet
+func observationsToObservationResponse(req *pbv3.ObservationRequest, observations []*Observation) *pbv3.ObservationResponse {
+	// The select options are handled separately since each has a different behavior in V2.
+	// For now, V3 will match the behavior of V2 to preserve backward compatibility and allow datasource merging.
+	// TODO: Unify these responses more.
```
keyurva, line 155: Can you capture the differences in a bug and mention it here?

internal/server/spanner/dsutil.go
```diff
@@ -188,8 +176,166 @@ func newObservationResponse(variables []string) *pbv3.ObservationResponse {
 	return result
 }
 
-func observationToFacetObservation(observation *Observation, queryObs bool) (string, *pb.Facet, *pbv2.FacetObservation) {
-	facetId, facet := observationToFacet(observation)
+func groupObservationsByVariableAndEntity(observations []*Observation) map[string]map[string][]*Observation {
+	result := map[string]map[string][]*Observation{}
```
keyurva, line 180: nit: if a single level map works for the consuming code, consider defining a `variableEntity` struct using a `map[variableEntity][]*Observation`

internal/server/spanner/dsutil.go
```diff
@@ -188,8 +176,166 @@ func newObservationResponse(variables []string) *pbv3.ObservationResponse {
...
+				response.Facets[facetId] = facet
+			}
+		}
+	}
+
+	return response
+}
+
+// The rows are sorted when returned from Spanner,  so the child places will be in order.
+func getChildPlaces(observations []*Observation) []string {
```
keyurva, line 216: nit: is this method by itself about child places or just about returning distinct places in order from the specified observations? If so, consider renaming to `getDistinctPlaces` or similar.

internal/server/spanner/dsutil.go
```diff
@@ -188,8 +176,166 @@ func newObservationResponse(variables []string) *pbv3.ObservationResponse {
...
+		if ok {
+			variableObs.ByEntity[ENTITY_PLACEHOLDER].OrderedFacets = mergeEntityOrderedFacets(initialVariableObs.ByEntity, childPlaces)
+		}
+	}
+	return mergedResponse
+}
+
+func obsToExistenceResponse(req *pbv3.ObservationRequest, observations []*Observation) *pbv3.ObservationResponse {
+	// This is likely a bug in V2, but will be kept for now to not break existing behavior.
+	// TODO: Return actual response.
```
keyurva, line 303: Thanks for thinking of backward compatibility.

In this case since not sending a response is a bug, any reason not to "fix" it here and subsequently in V2 as well?

internal/server/spanner/dsutil.go
```diff
@@ -132,47 +137,30 @@ func queryObs(qo *queryOptions) bool {
...
-			facetObservation,
-		)
-		response.Facets[facetId] = facet
+func observationsToObservationResponse(req *pbv3.ObservationRequest, observations []*Observation) *pbv3.ObservationResponse {
+	// The select options are handled separately since each has a different behavior in V2.
+	// For now, V3 will match the behavior of V2 to preserve backward compatibility and allow datasource merging.
+	// TODO: Unify these responses more.
+	qo := selectFieldsToQueryOptions(req.Select)
+	if queryObs(&qo) {
+		return obsToObsResponse(req, observations)
```
keyurva, line 158: Can you add brief comments here or elsewhere on what the 3 types of responses are.

keyurva (approved):
This PR required a lot of attention to detail. Thanks for working through it and thinking deeply about backward compatibility!

internal/server/spanner/dsutil.go
n-h-diaz: Added a bug to the hotlist and added more description to the comment. I also wrote a short doc (linked the bug) that has some examples of the differences

internal/server/spanner/dsutil.go
n-h-diaz: Good idea, thanks! Done

internal/server/spanner/dsutil.go
n-h-diaz: Makes sense, updated

internal/server/spanner/dsutil.go
n-h-diaz: I think I'd have to look more in website code to determine if this is actually a bug (facet is mostly an internal use case). I updated the comment, but agree that if this is an issue we can fix in both v2 and v3

internal/server/spanner/dsutil.go
n-h-diaz: Done

PR #1504, opened 2025-01-18
Encapsulate all remaining queries in sql client.

keyurva:
* All queries are now encapsulated!
* Removed the public DB variable from the client since it's no longer needed.
* One remaining thing is to use a SQLClient pointer (vs by value) in the store - will do that in the next PR.

### DIFF
--- a/internal/sqldb/model.go
+++ b/internal/sqldb/model.go
@@ -133,7 +133,7 @@ type EntityVariable struct {
 	Variable string `db:"variable"`
 }
 
-// EntityVariable represents a row that includes an entity and a variable.
+// ProvenanceInfo represents a row that includes provenance info (id, name, URL).
 type ProvenanceInfo struct {
 	ID   string `db:"provenance_id"`
 	Name string `db:"provenance_name"`
--- a/internal/sqldb/model.go
+++ b/internal/sqldb/model.go
@@ -16,6 +16,7 @@
 package sqldb
 
 import (
+	"encoding/json"
 	"fmt"
 	"strings"
 )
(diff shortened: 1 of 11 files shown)


internal/sqldb/model.go
```diff
@@ -125,3 +126,49 @@ type ObservationCount struct {
...
 	Count    int    `db:"num_obs"`
 }
+
+// EntityVariable represents a row that includes an entity and a variable.
+type EntityVariable struct {
+	Entity   string `db:"entity"`
+	Variable string `db:"variable"`
+}
+
+// EntityVariable represents a row that includes an entity and a variable.
```
n-h-diaz, line 136: nit: update comment

internal/sqldb/statements.go
```diff
@@ -277,4 +284,22 @@ var statements = struct {
...
+			t1.predicate = "typeOf"
+			AND t1.object_id = "Provenance"
+			AND t2.predicate = "name"
+			AND t3.predicate = "url"
+	`,
+	getAllImports: `
+		SELECT imported_at, status, metadata
+		FROM imports
+		ORDER BY imported_at DESC
+		LIMIT 100;
```
n-h-diaz, line 303: What is the limit for?

n-h-diaz (approved):
Thanks!

internal/sqldb/model.go
keyurva: Thanks for catching it - done.

internal/sqldb/statements.go
keyurva: To keep the results bounded to the most recent 100 which is sufficient for this endpoint. Added a comment in the code to clarify this.

PR #1505, opened 2025-01-21
Move V3 protos to V2

n-h-diaz:
We've decided to keep the protos unchanged, so can use the same ones for simplicity. (If we need to modify the V2 protos in the future, we'll only add new fields and mark them as specifically for V3)

I added NodeSearch proto to V2 as well for consistency


keyurva (approved):
Thanks!

PR #1507, opened 2025-01-23
Rank exact matches higher.

keyurva:
* Made a couple of changes to node search:
   + Rank exact matches higher.
   + When scores are the same, sort by name to keep results consistent. Currently the results in this scenario are non-deterministic which affects golden tests.
* Also, separated sql / gql statements in a separate file to keep file lengths in check.


internal/server/v3/nodesearch/golden/node_search_test.go
```diff
@@ -33,6 +33,7 @@ const (
 )
 
 func TestV3NodeSearch(t *testing.T) {
+	test.EnableSpannerGraph = true
```
n-h-diaz, line 36: Are the tests stable enough to be enabled by default? I wonder if this might cause failures for other folks as we load more data

n-h-diaz (approved):
Thanks for separating out the queries!

internal/server/v3/nodesearch/golden/node_search_test.go
keyurva: No, don't want it enabled by default right now - thanks for catching! I'd only enabled it to test it locally. Undid this.

keyurva (commented):
Thanks for the review!

PR #1508, opened 2025-01-24
Create a dispatcher framework for managing request lifecycle.

keyurva:
* Introduced a `Dispatcher` for managing a request's lifecycle - pre-processing, handling, post-processing.
  + Could have done this in the handler itself but a standalone struct allows us to test it independent of a server.
* Introduced a `Processor` interface that can be implemented for pre and post processing.
* Created a no-op `CalculationProcessor` as an example.
   + To be replaced with a real implementation subsequently.
* Made a pragmatic call to use casting to get to the concrete request and response types.
   + See code comments for details.


n-h-diaz (approved):
Thanks! This makes a lot more sense now, and I think will really help generalize things! I can update the calculation pr once this is checked in

PR #1509, opened 2025-01-28
Add support for neighbors to V3 Node

n-h-diaz:
This is currently only used for one hardcoded case in V2: https://github.com/datacommonsorg/mixer/blob/bbfacfd7e542df897658f94e44b9b2dcfafcdc2f/internal/server/v2/propertyvalues/api.go#L201-L234 , so I added a similar special case in V3. 

Eventually, I think we should migrate website to use arcs for this request too, since now we don't need to cache this and can use Spanner directly. 

This can use Spanner instead of the Mixer cache, since all Spanner custom DCs will use the new Custom DC architecture so SVGs will no longer be in BT (so no dc/g/Custom_), and so neighbors for Custom groups should be handled by sql datasource separately

### DIFF
--- a/internal/server/spanner/datasource.go
+++ b/internal/server/spanner/datasource.go
@@ -51,6 +51,18 @@ func (sds *SpannerDataSource) Node(ctx context.Context, req *pbv2.NodeRequest) (
 	}
 	arc := arcs[0]
 
+	// Certain requests utilize neighbors in the response so are handled differently.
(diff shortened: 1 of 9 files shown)


n-h-diaz:
Closing this PR since the V2 functionality is no longer being used.

As a followup, I'll migrate the V2 use case to use arcs instead.

PR #1510, opened 2025-01-28
Deprecate neighbors in V2 Node

n-h-diaz:
Updates ->specializationOf+{typeOf:StatVarGroup} requests to use arcs.

(Note that the response doesn't contain node names since these aren't in the cache.)


internal/server/v2/propertyvalues/api.go
```diff
@@ -198,35 +198,19 @@ func LinkedPropertyValues(
...
-						Neighbor: map[string]*pbv2.LinkedGraph{},
-					}
-					g = g.Neighbor[curr]
-				} else {
-					break
+			res.Data[node] = &pbv2.LinkedGraph{}
+			ancestors := hierarchy.GetSVGAncestors(node, parentSvgs)
+			if len(ancestors) > 0 {
+				res.Data[node].Arcs = map[string]*pbv2.Nodes{
+					hierarchy.SpecializationOf + "+": {
```
keyurva, line 211: nit: Consider declaring "+" as a const.

internal/server/statvar/hierarchy/statvar_hierarchy_util.go
```diff
@@ -315,3 +317,44 @@ func MergeSVGNodes(node1, node2 *pb.StatVarGroupNode) {
 		}
 	}
 }
+
+// GetSVGAncestors returns EntityInfo of SVG ancestors for an SVG from the Mixer cache.
+// The EntityInfo does not contain names, which are not stored in the cache.
+func GetSVGAncestors(node string, parentSvgs map[string][]string) []*pb.EntityInfo {
+	ancestors := []string{}
```
keyurva, line 324: nit: You can probably declare `ancestors` as an `[]*pb.EntityInfo` itself to avoid the second loop. To sort it, you can use `sort.Slice`

```
sort.Slice(ancestors, func(i, j int) bool {
        return ancestors[i].Dcid < ancestors[j].Dcid
})
```

proto/v2/node.proto
```diff
@@ -27,10 +27,10 @@ message Nodes {
 message LinkedGraph {
   // Keyed by property, with values being a list of nodes
   map<string, Nodes> arcs = 1;
-  // Keyed by property, with values being another LinkedGraph
-  map<string, LinkedGraph> neighbor = 2;
   // Only set when querying for properties only, with "[nodes]->" or "[nodes]<-".
   repeated string properties = 3;
+
+  reserved 2;
```
keyurva, line 33: nit: Perhaps add a comment that this was the deprecated neighbor field for future reference if needed.

internal/server/v2/propertyvalues/api.go
keyurva: For my understanding: Are there other instances where we return "+" nodes?

keyurva (approved):
Thanks!

internal/server/statvar/hierarchy/statvar_hierarchy_util.go
n-h-diaz: Thanks this is helpful! Done

proto/v2/node.proto
n-h-diaz: Good idea, done

PR #1511, opened 2025-01-30
Update chain query

n-h-diaz:
Use "match any" instead of "distinct" which has slightly better performance.

(Also one nice side effect is that now we can return types! I can update V2 to match as a follow up)


keyurva (approved):
Nice improvements!

PR #1513, opened 2025-01-31
Fix non deterministic facet ranking

chejennifer:
Update facet ranking to take import name into account because some facets differ by only the import name. This non-deterministic ranking of facets causes flaky NL tests in the website repository because NL chooses the top ranked facet.


keyurva (approved):
Thanks for the fix!

PR #1516, opened 2025-02-07
Add ability to node search across all predicates.

keyurva:
* V3 node search now supports querying by `predicates` in addition to `types`.
* If no predicates are specified, it searches on `name` and `description` by default.
* The response includes the matched property and value for each result.

---

Example request with all supported parameters (only `query` is required):

```
{
  "query": "ether",
  "types": ["ChemicalCompound"],
  "predicates": ["alternateName"]
}
```


proto/v2/node_search.proto
```diff
@@ -28,10 +28,26 @@ message NodeSearchRequest {
...
-  // List of nodes resulting from the search.
-  repeated datacommons.EntityInfo nodes = 1;
+  // List of results.
+  repeated NodeSearchResult results = 2;
+
+  // Deprecated nodes field
+  reserved 1;
+}
+
+message NodeSearchResult {
```
keyurva, line 47: @n-h-diaz - let me know if this works for results or if a different definition works better.

n-h-diaz: Discussed offline - this is fine for the experimental API, but we can get more detailed review on request/response when we finally merge with resolve

n-h-diaz (approved):
Thanks for the update! :)

PR #1518, opened 2025-02-13
Add merging to V3 resolve and node search

n-h-diaz:


internal/merger/merger.go
```diff
@@ -395,18 +414,65 @@ func processSearchStatVarResponse(resp *proto.SearchStatVarResponse, mergedStatV
...
+			// Check if all results from resp have been added.
+			if counter >= len(allResp[i].Results) {
+				complete[i] = true
+				continue
+			}
+
+			result := allResp[i].Results[counter]
+
+			// Check if result was already added.
+			key, err := proto.Marshal(result)
```
keyurva, line 463: I'm wondering how expensive this check is and whether we need to check all? If we do, can we base it on `result.Node.Dcid`?

internal/merger/merger.go
```diff
@@ -395,18 +414,65 @@ func processSearchStatVarResponse(resp *proto.SearchStatVarResponse, mergedStatV
...
+	if len(allResp) == 0 {
+		return &pbv2.NodeSearchResponse{}, nil
+	}
+
+	merged := &pbv2.NodeSearchResponse{}
+	results := map[string]bool{}
+	complete := map[int]bool{}
+	counter := 0
+loop:
+	for {
```
keyurva, line 445: nit and optional: I prompted gemini on different ways of merging arrays based on our requirements and it came up with the following. Feel free to use it and perfectly fine to ignore (the logic in the PR works):

```
func mergeArraysClean(arrays [][]int, maxNumResults int) []int {
        merged := make([]int, 0, maxNumResults)

        for len(merged) < maxNumResults && len(arrays) > 0 {
                newArrays := make([][]int, 0, len(arrays)) // Create a new slice for remaining arrays
                added := false

                for _, arr := range arrays {
                        if len(arr) > 0 {
                                merged = append(merged, arr[0])
                                added = true
                                if len(merged) >= maxNumResults {
                                        return merged
                                }
                                if len(arr) > 1 {
                                        newArrays = append(newArrays, arr[1:]) // Add remaining elements to the new slice
                                }
                        }
                }
                if !added {
                        break;
                }

                arrays = newArrays // Update arrays with the remaining sub-arrays
        }

        return merged
}
```

keyurva (commented):
Thanks!

internal/merger/merger.go
n-h-diaz: Discussed offline and we'll only return the first match for a dcid, so yep we can just use dcid - updated

internal/merger/merger.go
n-h-diaz: Thanks! I should use gemini more 😮  I updated to use this (slightly modified to match the existing syntax)

internal/merger/merger.go
```diff
@@ -395,18 +413,57 @@ func processSearchStatVarResponse(resp *proto.SearchStatVarResponse, mergedStatV
...
+func MergeMultiNodeSearch(allResp []*pbv2.NodeSearchResponse) (*pbv2.NodeSearchResponse, error) {
+	if len(allResp) == 0 {
+		return &pbv2.NodeSearchResponse{}, nil
+	}
+
+	merged := &pbv2.NodeSearchResponse{}
+	results := map[string]bool{}
+
+	for len(merged.Results) < MAX_SEARCH_RESULTS && len(allResp) > 0 {
+		newResp := []*pbv2.NodeSearchResponse{} // Create a new slice for remaining arrays
```
keyurva, line 443: super nit: Consider adding comments above the statement instead of inline with the statement. Not sure why gemini does the latter.

keyurva (approved):
Thanks for the updates!

internal/merger/merger.go
n-h-diaz: Good idea, done.

Thanks for the review!

PR #1523, opened 2025-02-25
Add observation response limits to spanner

n-h-diaz:
This is to match the behavior of V2 which returns the same error for large requests (e.g. Count_Person for country/USA<-containedInPlace+{typeOf:City}) to prevent timeouts/memory limits: https://github.com/datacommonsorg/mixer/blob/master/internal/server/v2/observation/contained_in.go#L160

Currently the limit is set to the same as V2 (5000 series per response)

### DIFF
--- a/internal/server/spanner/datasource.go
+++ b/internal/server/spanner/datasource.go
@@ -21,6 +21,9 @@ import (
 	pbv2 "github.com/datacommonsorg/mixer/internal/proto/v2"
 	"github.com/datacommonsorg/mixer/internal/server/datasource"
 	v2 "github.com/datacommonsorg/mixer/internal/server/v2"
+	"github.com/datacommonsorg/mixer/internal/server/v2/shared"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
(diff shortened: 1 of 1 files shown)


internal/server/spanner/datasource.go
```diff
@@ -91,6 +94,14 @@ func (sds *SpannerDataSource) Observation(ctx context.Context, req *pbv2.Observa
 
 	observations = filterObservationsByDateAndFacet(observations, date, req.Filter)
 
+	if len(observations) > shared.MaxSeries {
+		return nil, status.Errorf(
+			codes.Internal,
+			"Stop processing large number of concurrent observation series: %d",
```
keyurva, line 100: Should we return an error message that helps the user take an alternative action? Something along the lines of:

"Your request exceeds the allowed limit on the number of time series processed: x > y. Please try reducing the number of variables or entities in your request."

keyurva (approved):
Thanks!

internal/server/spanner/datasource.go
n-h-diaz: Good idea, thanks! Updated

PR #1526, opened 2025-02-27
Remove non numeric values from time series

n-h-diaz:


internal/server/spanner/dsutil.go
```diff
@@ -349,7 +358,15 @@ func observationToFacetObservation(observation *Observation, includeObs bool) (*
 	var observations []*pb.PointStat
 
 	for _, dateValue := range observation.Observations.Observations {
-		observations = append(observations, dateValueToPointStat(dateValue))
+		pointStat, err := dateValueToPointStat(dateValue)
+
+		// Skip observations with non-numeric values.
+		if err != nil {
+			log.Printf("Error decoding PointStat: %v", err)
```
keyurva, line 365: nit: Consider adding more info about the row (entity, var, date, val) to the log to help with debugging.

internal/server/spanner/model.go
```diff
@@ -57,7 +56,7 @@ type Observation struct {
 
 type DateValue struct {
 	Date  string
-	Value float64
+	Value string
```
keyurva, line 59: nit: Perhaps add a comment on why string and not float.

internal/server/v3/observation/golden/observation_test.go
```diff
@@ -187,6 +187,18 @@ func TestV3Observation(t *testing.T) {
...
+				req: &pbv2.ObservationRequest{
+					Variable: &pbv2.DcidOrExpression{
+						Dcids: []string{"Median_Age_Person"},
+					},
+					Entity: &pbv2.DcidOrExpression{
+						Dcids: []string{"geoId/4822892"},
+					},
+					Select: []string{"entity", "variable", "date", "value"},
+				},
+				goldenFile: "observations_str_val.json",
```
keyurva, line 200: nit: A comment on a pointer / screenshot to the string value in the DB would be helpful.

keyurva (approved):
Thanks!

internal/server/spanner/dsutil.go
n-h-diaz: Done

internal/server/spanner/model.go
n-h-diaz: Done

internal/server/v3/observation/golden/observation_test.go
n-h-diaz: Done

PR #1527, opened 2025-02-27
Integrate redis cache with the v3 api.

keyurva:
* Reads redis config from a new yaml file.
* Redis instances are configured per region similar to website.
* Instantiates a `CacheProcessor` on startup if redis is configured.
* Verified by running a local mixer against a local redis instance:
```
go run cmd/main.go \
    ...
    --use_redis=true \
    --redis_info="
instances:
  - region: dummy
    host: "localhost"
    port: "6379"
    " &
```


deploy/helm_charts/mixer/templates/deployment.yaml
```diff
@@ -103,6 +103,8 @@ spec:
             {{- end }}
             - --use_spanner_graph={{ $.Values.mixer.useSpannerGraph }}
             - --spanner_graph_info=$(SPANNER_GRAPH_INFO)
+            - --use_redis={{ $.Values.mixer.useRedis }}
```
keyurva, line 106: @n-h-diaz - I'm not intimately familiar with the deployment configs. Can you take a closer look at them? If needed, I can create a separate PR for the deployment files.

n-h-diaz: You might need to add the corresponding config map entry for redis info: https://github.com/datacommonsorg/mixer/blob/4ff6d6845147e911f1623920c3a3866740d79e7e/deploy/helm_charts/mixer/templates/config_maps.yaml#L35-L44 , which you can then point to the yaml file (example: https://github.com/datacommonsorg/mixer/blob/4ff6d6845147e911f1623920c3a3866740d79e7e/scripts/deploy_gke.sh#L126-L128)

n-h-diaz (approved):
Thanks! Excited to try this out :)

deploy/helm_charts/mixer/templates/deployment.yaml
n-h-diaz: I think this looks okay, thanks! There are a few other docs that reference deployment that we might eventually want to update as well, but that can be done later after more testing

keyurva: Got it. Thanks for the review!

PR #1531, opened 2025-03-03
Add id field to pagination proto

n-h-diaz:
Originally was trying to reuse the V1 pagination proto, but it was very specific to BT/import group implementation so was difficult to use for Spanner. So this will hopefully be more generic, and each Data Source can have it's own implementation.

As a side effect, I needed to modify the remote datasource to use V3 endpoint, since V2 relies on v1Pagination.remotePaginationInfo (but now we won't be distinguishing remote vs other data source) 

Edit: now this just adds a new field for the spanner implementation


proto/v2/pagination.proto
```diff
@@ -0,0 +1,31 @@
...
+message Cursor {
+  // The id of the cursor.
+  string id = 1;
+}
+
+// Represents the cursor information of one pagination request.
+// Each Cursor corresponds to one DataSource.
+message PaginationInfo {
+  // Keyed by DataSource type.
+  map<string, Cursor> cursors = 1;
```
keyurva, line 30: I have a few questions here:

- the datasource type is currently internal, is the plan here to make it external?
- we could have multiple datasources of the same type (e.g. multiple remote mixers). How would this work in that case?
- Is the plan to add a new field of type `PaginationInfo` to the response? Or repurpose `nextToken` to be a string representation (e.g. marshal proto, then base64 encode) of this proto?

n-h-diaz: For the first two questions: I updated the comment to be any unique identifier (so could be type, or could be index, especially if there are multiple of the same type, etc). The main thing is that each datasource can identify it's own unique portion of the pagination info to use in subsequent requests. 

For the second, yep I was thinking of using nextToken. This is how it's done currently, where it's just an encoded proto: https://github.com/datacommonsorg/mixer/blob/fe51f865cb23c9d7cb372016922c368dac79243f/internal/server/pagination/pagination.go#L26-L41

n-h-diaz: Spoke offline and we'll reuse the previous proto and just have CursorGroups for each data source. So I added the field instead

proto/v1/pagination.proto
```diff
@@ -26,6 +26,8 @@ message Cursor {
   int32 page = 2;
   // The position of the next read item in the current page, starts from 0.
   int32 item = 3;
+  // The id of the cursor.
```
keyurva, line 29: nit: add a comment indicating that this will be used by spanner. If other data sources use it in the future, they should update the comment.

keyurva (approved):
Thanks for the updates!

proto/v1/pagination.proto
n-h-diaz: Done

PR #1532, opened 2025-03-04
Add Id() to datasource

n-h-diaz:
This will be used to identify between multiple datasources of the same type (particularly for pagination)

### DIFF
--- a/internal/server/spanner/client.go
+++ b/internal/server/spanner/client.go
@@ -26,11 +26,12 @@ import (
 // SpannerClient encapsulates the Spanner client.
 type SpannerClient struct {
 	client *spanner.Client
+	id     string
 }
 
 // newSpannerClient creates a new SpannerClient.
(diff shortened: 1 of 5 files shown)


internal/server/spanner/client.go
```diff
@@ -60,6 +62,11 @@ func createSpannerClient(ctx context.Context, cfg *SpannerConfig) (*spanner.Clie
 	return client, nil
 }
 
+// createSpannerId creates the id for the Spanner client.
+func createSpannerId(cfg *SpannerConfig) string {
+	return fmt.Sprintf("%s.%s.%s", cfg.Project, cfg.Instance, cfg.Database)
```
keyurva, line 67: Looks like there is a [DatabaseName() method](https://pkg.go.dev/cloud.google.com/go/spanner#Client.DatabaseName) on `spanner.Client` that we can use for this purpose.

PR #1539, opened 2025-03-05
Add pagination for spanner data source

n-h-diaz:
The spanner pagination info is the number of the next page. For the corresponding CursorGroup in PaginationInfo, it's stored in Cursors[0].Page.

Follow up: will update merger to use pagination info


internal/server/spanner/dsutil.go
```diff
@@ -82,17 +85,96 @@ func nodePropsToNodeResponse(propsBySubjectID map[string][]*Property) *pbv2.Node
...
 		nodeResponse.Data[subjectID] = nodeEdgesToLinkedGraph(edges)
+		rows += len(edges)
+		if len(edges) > 0 {
+			cursor = edges[len(edges)-1]
+		}
 	}
 
-	return nodeResponse
+	// If the number of rows equals the page size, there's likely more rows.
+	if rows == PAGE_SIZE {
```
keyurva, line 169: One edge case is if the number of rows is equal to the page size. Any thoughts on how that can be addressed?

One approach that comes to mind is to fetch 1 more than page size from the DB and generate a next token based on that.

internal/server/spanner/statements.go
```diff
@@ -227,6 +231,15 @@ var statements = struct {
 			name,
 			types			
 	`,
+	cursorNodeFilter: `
+		AND CONCAT(m.subject_id, n.subject_id) > '%s'
```
keyurva, line 235: Can this be parameterized to avoid sql injection issues since the string here is based on values in the user request? (Similar for the edge filter)

keyurva, line 235: nit: Perhaps add a comment that this approach is predicated on the results being ordered a certain way.

keyurva (commented):
Very cool use of order by to implement pagination!

internal/server/spanner/dsutil.go
```diff
@@ -82,17 +85,96 @@ func nodePropsToNodeResponse(propsBySubjectID map[string][]*Property) *pbv2.Node
...
+		return nil, err
+	}
+
+	for _, cursorGroup := range pi.GetCursorGroups() {
+		for _, key := range cursorGroup.GetKeys() {
+			if key == id {
+				if len(cursorGroup.GetCursors()) < 1 {
+					return nil, fmt.Errorf("pagination info missing cursor group for data source: %s", id)
+				}
+				cursorId := cursorGroup.GetCursors()[0].GetId()
```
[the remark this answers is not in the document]
n-h-diaz: This is a good idea! I can do this as a follow up to clean this up more

internal/server/spanner/dsutil.go
```diff
@@ -82,17 +85,96 @@ func nodePropsToNodeResponse(propsBySubjectID map[string][]*Property) *pbv2.Node
...
+	pi, err := pagination.Decode(nextToken)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, cursorGroup := range pi.GetCursorGroups() {
+		for _, key := range cursorGroup.GetKeys() {
+			if key == id {
+				if len(cursorGroup.GetCursors()) < 1 {
+					return nil, fmt.Errorf("pagination info missing cursor group for data source: %s", id)
```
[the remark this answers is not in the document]
n-h-diaz: I was thinking that if the cursor group is found for a datasource, there should be cursor information for it (so no empty cursor groups to try and reduce the size of the token). So if it's missing, the pagination info was created incorrectly. I updated the error message to be more clear. Lmk what you think!

internal/server/spanner/dsutil.go
```diff
@@ -82,17 +85,96 @@ func nodePropsToNodeResponse(propsBySubjectID map[string][]*Property) *pbv2.Node
...
+	if err != nil {
+		return "", err
+	}
+
+	return nextToken, nil
+}
+
 // nodeEdgesToNodeResponse converts a map from subject id to its edges to a NodeResponse proto.
-func nodeEdgesToNodeResponse(edgesBySubjectID map[string][]*Edge) *pbv2.NodeResponse {
+func nodeEdgesToNodeResponse(nodes []string, edgesBySubjectID map[string][]*Edge, id string) (*pbv2.NodeResponse, error) {
```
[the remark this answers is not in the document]
n-h-diaz: Good idea, done

internal/server/spanner/dsutil.go
n-h-diaz: This is a really good idea, thank you! I had been wondering about this case and had thought there would just have to be an empty page, but this is a cool workaround :)

internal/server/spanner/statements.go
n-h-diaz: Thanks for the catch! Updated

n-h-diaz: Done

n-h-diaz:
Thank you for the review! 

After thinking some more, I'm not sure if this approach makes the most sense, as it's more complicated (I found a few bugs I had to fix) and actually performs worse than offset pagination for small offsets (it's fairly consistent regardless of page number so eventually will be faster). I can try to see if there are any improvements, but let me know if you have any suggestions or thoughts!

keyurva:
> Thank you for the review!
> 
> After thinking some more, I'm not sure if this approach makes the most sense, as it's more complicated (I found a few bugs I had to fix) and actually performs worse than offset pagination for small offsets (it's fairly consistent regardless of page number so eventually will be faster). I can try to see if there are any improvements, but let me know if you have any suggestions or thoughts!

Does the latest version address the performance? Does the complex where clause associated with pagination cause any issues?

internal/server/spanner/dsutil.go
```diff
@@ -116,19 +116,23 @@ func getCursorEdge(nextToken, id string) (*Edge, error) {
...
 
 // getNextToken encodes the cursor Edge in a nextToken string.
-func getNextToken(edge *Edge, id string) (string, error) {
+func getNextToken(edge *Edge, dataSourceID string) (string, error) {
 	if edge == nil {
 		return "", nil
 	}
 
+	// Clear name and types to save space since they aren't used.
+	edge.Name = ""
```
keyurva, line 125: Should we create a clone of the edge and update the clone so as to not accidentally update the original edge instance if ever it is used for other purposes.

internal/server/spanner/statements.go
```diff
@@ -227,6 +231,23 @@ var statements = struct {
...
 	`,
+	// The cursor assumes the results are in order, specified by the ORDER BY clauses.
+	cursorNodeFilter: `
+		AND (m.subject_id > @cursor_subject_id
+		OR (m.subject_id = @cursor_subject_id AND n.subject_id > @cursor_object_id)
+		OR (m.subject_id = @cursor_subject_id AND n.subject_id = @cursor_object_id)) 
+	`,
+	cursorEdgeFilter: `
+		AND (m.subject_id > @cursor_subject_id
+		OR (m.subject_id = @cursor_subject_id AND e.predicate > @cursor_predicate)
```
keyurva, line 242: In the meeting last week, we chatted about trying to avoid using OR clauses for performance reasons. Checking if you tried it and if it worked or not?

n-h-diaz:
Sorry for the delay on this and thanks for your patience! This is updated with offset pagination now (only for spanner), and I'll follow up with merging

n-h-diaz:
Thanks for the reviews! Updated to store offset instead of page

proto/v1/pagination.proto
```diff
@@ -19,15 +19,16 @@ option go_package = "github.com/datacommonsorg/mixer/internal/proto/v1";
...
-  // The index of the page, starts from 0. Used for Bigtable.
+  // The index of the page, starts from 0
   int32 page = 2;
   // The position of the next read item in the current page, starts from 0. 
-  // Used for Bigtable.
   int32 item = 3;
-  // The id of the cursor. Used for Spanner.
-  string id = 4;
+  // The offset of the next read item, starts from 0.
+  int32 offset = 5;
```
keyurva, line 29: nit: Perhaps add "Used for Spanner"?

keyurva (approved):
Thanks for the updates!

PR #1540, opened 2025-03-05
Fix sorting order

n-h-diaz:
Fixes the sorting one of the spanner queries to be more stable

### DIFF
--- a/internal/server/spanner/statements.go
+++ b/internal/server/spanner/statements.go
@@ -102,11 +102,21 @@ var statements = struct {
 			e.provenance,
 			'' AS name,
 			ARRAY<STRING>[] AS types
+		NEXT
+		RETURN
+			subject_id,
+			predicate,
+			object_id,
+			object_value,
+			provenance,
+			name,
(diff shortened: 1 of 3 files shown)


internal/server/spanner/statements.go
```diff
@@ -102,11 +102,21 @@ var statements = struct {
 			e.provenance,
 			'' AS name,
 			ARRAY<STRING>[] AS types
+		NEXT
```
keyurva, line 105: Just for my understanding: what does the NEXT clause do?

keyurva (approved):
Thanks!

internal/server/spanner/statements.go
n-h-diaz: It creates a linear query statement (uses output as new input) https://cloud.google.com/spanner/docs/graph/queries-overview#graph-linear-query-statements

Thanks for the review!

keyurva: Thanks!

PR #1541, opened 2025-03-07
Support node properties request in sql datasource.

keyurva:
* Creates and adds a sql datasource if v3 is enabled and sql params are configured both for runtime and golden tests.
* Implements the node properties request.
* Fixes some response merge logic.

### DIFF
--- a/internal/server/v2/graph.go
+++ b/internal/server/v2/graph.go
@@ -15,6 +15,8 @@
 // Package v2 is the version 2 of the Data Commons REST API.
 package v2
 
+import "github.com/datacommonsorg/mixer/internal/util"
+
 // Arc represents an arc in the graph.
 type Arc struct {
 	// Whether it's out or in arc.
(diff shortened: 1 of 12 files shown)


internal/server/v3/node/golden/out_pv_all.json
```diff
@@ -10,14 +10,14 @@
                 "Property"
               ],
               "dcid": "gender",
-              "provenanceId": "dc/base/HumanReadableStatVars"
+              "provenance_id": "dc/base/HumanReadableStatVars"
```
keyurva, line 13: Not sure why it's marshaling it as snake case now. Let me know if you've run into this. I tried explicitly setting [UseProtoNames](https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson#MarshalOptions.UseProtoNames) to `false` but that had no effect.

The http response continues to be camel case so this is not an issue there.

internal/server/v2/graph.go
```diff
@@ -29,6 +31,17 @@ type Arc struct {
...
 }
 
+func (arc *Arc) Direction() string {
+	if arc.Out {
+		return util.DirectionOut
+	}
+	return util.DirectionIn
+}
+
+func (arc *Arc) IsNodePropertiesArc() bool {
```
keyurva, line 41: Let me know if this function and function name makes sense or if you have any alternate suggestions.

internal/server/v3/node/golden/out_pv_all.json
n-h-diaz: Yep I've run into this for a lot of the goldens, though haven't done anything for it yet since the test seem to be okay

n-h-diaz (approved):
Thanks!

PR #1543, opened 2025-03-11
Add support for fetching property values in sql datasource.

keyurva:


internal/sqldb/dsutil.go
```diff
@@ -17,9 +17,116 @@
...
+	namePredicate   = "name"
+	typeOfPredicate = "typeOf"
+	subjectIdColumn = "subject_id"
+	objectIdColumn  = "object_id"
+	defaultType     = "Thing"
+)
+
+type entityInfo struct {
+	Name string
+	Type string
```
n-h-diaz, line 35: Will there always only be one type?

n-h-diaz (approved):
Thanks!

internal/sqldb/dsutil.go
keyurva: With sql at this time, yes that's the case.

keyurva (commented):
Thanks for the review!

PR #1550, opened 2025-04-03
Return correct facet info in series facet responses.

keyurva:


internal/server/v2/facet/series.go
```diff
@@ -127,44 +129,49 @@ func SeriesFacet(
...
+	response = shared.TrimObservationsResponse(response)
+
+	for _, variableData := range response.ByVariable {
+
+		for _, entityData := range variableData.ByEntity {
+			for _, facetData := range entityData.OrderedFacets {
+				obsCount := len(facetData.Observations)
+				facetData.Observations = []*pb.PointStat{
+					{
+						Value: proto.Float64(float64(obsCount)),
```
keyurva, line 169: I'm continuing to return this value since it is being returned currently but I'm not sure whether it is used at all. More likely, the [obs_count](https://github.com/datacommonsorg/mixer/blob/9d86bfc62b1f5da060cc9af333f5d6d8f4c57d46/proto/v2/observation.proto#L63) field is the one used instead of this one. Also, this field is not returned for base DC facets.

@dwnoble - if there's a definitive way to find this out else I'll leave it as-is.

n-h-diaz (approved):

PR #1552, opened 2025-04-04
Update V3 node merging with pagination

n-h-diaz:
Updated to use the V2.Pagination proto

I also modified the proto here to use a repeated field instead of a map so the encoded nextTokens would be consistent (I modified the proto in place vs using reserved since this proto isn't yet used anywhere else)


n-h-diaz:
I updated the PR to use the new proto, ptal when you have the chance :)  

I also modified the proto here to use a repeated field instead of a map so the encoded nextTokens would be consistent (I modified the proto in place vs using reserved since this proto isn't yet used anywhere else)

n-h-diaz:
Updated with the stringinfo - I think after doing some local testing it looks okay with the remote pagination, thanks for the suggestion!

internal/server/remote/dsutil.go
```diff
@@ -0,0 +1,73 @@
...
+
+		req.NextToken = ""
+		for _, dataSourceInfo := range info.Info {
+			if dataSourceInfo.GetId() == id {
+				string_info, ok := dataSourceInfo.GetDataSourceInfo().(*pbv2.Pagination_DataSourceInfo_StringInfo)
+				if !ok {
+					return fmt.Errorf("found different data source info for remote data source id: %s", id)
+				}
+
+				req.NextToken = string_info.StringInfo
```
keyurva, line 43: super nit: could break out of the loop here?

proto/v2/pagination.proto
```diff
@@ -28,11 +28,15 @@ message SpannerInfo {
...
-      SpannerInfo spanner_info = 2;
+      datacommons.v1.PaginationInfo bigtable_info = 2;
+      SpannerInfo spanner_info = 3;
+      string string_info = 4;
     }
   }
-  // The key is the id of the data source.
-  map<string, DataSourceInfo> info = 1;
+
+  repeated DataSourceInfo info = 1;
```
keyurva, line 41: Can you add a comment on why repeated vs map?

keyurva (approved):
Thanks for the updates and testing!

internal/server/remote/dsutil.go
n-h-diaz: Done

proto/v2/pagination.proto
n-h-diaz: Done

PR #1553, opened 2025-04-08
Deprecate V2 ResolveResponse.Entity.resolved_ids

n-h-diaz:
It looks like this is no longer being used in website and has been marked as deprecated in docs (https://docs.datacommons.org/api/rest/v2/resolve.html)


keyurva (approved):
Thanks. Can you remove it from the import tool as well as a follow up?

https://github.com/datacommonsorg/import/blob/e0a142b372fe62c478548d76407335f63eb044b4/util/src/main/proto/resolve.proto#L52
https://github.com/datacommonsorg/import/blob/e0a142b372fe62c478548d76407335f63eb044b4/util/src/main/java/org/datacommons/util/ExternalIdResolver.java#L295

n-h-diaz:
> Thanks. Can you remove it from the import tool as well as a follow up?
> 
> https://github.com/datacommonsorg/import/blob/e0a142b372fe62c478548d76407335f63eb044b4/util/src/main/proto/resolve.proto#L52 https://github.com/datacommonsorg/import/blob/e0a142b372fe62c478548d76407335f63eb044b4/util/src/main/java/org/datacommons/util/ExternalIdResolver.java#L295

Yes will do, thanks for the review!

PR #1555, opened 2025-04-16
Create V2.PaginationInfo proto

n-h-diaz:
Currently just a map of datasource id to info, where info can be datasource specific


proto/v2/pagination.proto
```diff
@@ -0,0 +1,38 @@
...
+import "v1/pagination.proto";
+
+message SpannerInfo {
+  // The offset of the next read item, starts from 0.
+  int32 offset = 1;
+}
+
+// Represents the pagination info of one paginated request.
+// Each DataSourceInfo corresponds to the info for one data source.
+message PaginationInfo {
```
keyurva, line 29: nit: Can we use a different name to avoid confusion. `PaginationInfo2` might also be ok.

keyurva (approved):

proto/v2/pagination.proto
n-h-diaz: Updated to just be "Pagination"

Thanks for the review!

PR #1556, opened 2025-04-17
Update spanner pagination to use v2.Pagination

n-h-diaz:
Will follow up with updated merging in https://github.com/datacommonsorg/mixer/pull/1552/files


keyurva (approved):
Thanks!

PR #1557, opened 2025-04-21
Add entity existence requests to V3 observation

n-h-diaz:
This is for requests of the form /observation?entity.dcids=<> where no variable is specified, so essentially getting the variables for a place(s). 

Kara brought up that this is supported in V2, but we noticed it wasn't yet supported it in V3. In V2, this only works for existence and not with any date or facet filters (so it's a bit inconsistent). Here I added it everywhere, but can restrict it if needed as a follow up, along with some other query optimizations 

This PR also splits up some tests into different tests to avoid test timeouts

### DIFF
--- a/internal/server/spanner/datasource.go
+++ b/internal/server/spanner/datasource.go
@@ -17,6 +17,7 @@ package spanner
 import (
 	"context"
 	"fmt"
+	"log"
 
 	pbv2 "github.com/datacommonsorg/mixer/internal/proto/v2"
 	"github.com/datacommonsorg/mixer/internal/server/datasource"
@@ -90,6 +91,7 @@ func (sds *SpannerDataSource) Observation(ctx context.Context, req *pbv2.Observa
 	if req.Variable != nil {
 		// Variable expressions are not yet supported in Spanner.
 		if req.Variable.Expression != "" {
+			log.Printf("Received spanner request with variable expression: %s. Variable expressions are not yet supported in Spanner.", req.Variable.Expression)
 			return nil, nil
 		}
 		variables = req.Variable.Dcids
--- a/internal/server/spanner/datasource.go
+++ b/internal/server/spanner/datasource.go
@@ -77,7 +77,27 @@ func (sds *SpannerDataSource) Node(ctx context.Context, req *pbv2.NodeRequest) (
 
 // Observation retrieves observation data from Spanner.
 func (sds *SpannerDataSource) Observation(ctx context.Context, req *pbv2.ObservationRequest) (*pbv2.ObservationResponse, error) {
-	variables, entities, entityExpr := req.Variable.Dcids, req.Entity.Dcids, req.Entity.Expression
+	if req.Entity == nil {
+		return nil, fmt.Errorf("entity must be specified")
+	}
+
+	entities, entityExpr := req.Entity.Dcids, req.Entity.Expression
+	if len(entities) > 0 && entityExpr != "" {
+		return nil, fmt.Errorf("only one of entity.dcids and entity.expression should be specified")
+	}
+
+	variables := []string{}
+	if req.Variable != nil {
+		// Variable expressions are not yet supported in Spanner.
+		if req.Variable.Expression != "" {
+			return nil, nil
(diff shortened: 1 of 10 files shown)


internal/server/spanner/datasource.go
```diff
@@ -77,7 +77,27 @@ func (sds *SpannerDataSource) Node(ctx context.Context, req *pbv2.NodeRequest) (
...
+	entities, entityExpr := req.Entity.Dcids, req.Entity.Expression
+	if len(entities) > 0 && entityExpr != "" {
+		return nil, fmt.Errorf("only one of entity.dcids and entity.expression should be specified")
+	}
+
+	variables := []string{}
+	if req.Variable != nil {
+		// Variable expressions are not yet supported in Spanner.
+		if req.Variable.Expression != "" {
+			return nil, nil
```
keyurva, line 93: Should we return an error instead of failing silently here?

internal/server/spanner/dsutil.go
```diff
@@ -248,12 +248,16 @@ func observationsToObservationResponse(req *pbv2.ObservationRequest, observation
...
 }
 
-func newObservationResponse(variables []string) *pbv2.ObservationResponse {
+func newObservationResponse(variable *pbv2.DcidOrExpression) *pbv2.ObservationResponse {
 	result := &pbv2.ObservationResponse{
 		ByVariable: map[string]*pbv2.VariableObservation{},
 		Facets:     map[string]*pb.Facet{},
 	}
-	for _, variable := range variables {
+	if variable == nil || variable.Expression != "" {
```
keyurva, line 256: If this method only supports creating responds for `variable.Dcids`, should we check for its existence vs a non-existence of something else? i.e. `variable == nil || len(variable.Dcids) == 0`

internal/server/spanner/golden/query/get_observations_entity.json
```diff
@@ -0,0 +1,82 @@
+[
+  {
+    "VariableMeasured": "Count_HeatTemperatureEvent",
+    "ObservationAbout": "wikidataId/Q341968",
+    "Observations": {
+      "Observations": [
```
keyurva, line 6: If this is meant to be an existence check, should we be returning observations?

How does it work with V2? Also, how's the performance of queries for entities with potentially a large number of variables? e.g. `/observation?entity.dcids=country/USA` for both V2 and now V3

internal/server/spanner/datasource.go
n-h-diaz: This was so the whole request doesn't fail (for example remote data source supports variable expressions, so we could still have a valid response even if spanner data source doesn't return anything)

internal/server/spanner/golden/query/get_observations_entity.json
n-h-diaz: We currently return observations for all existence checks in spanner, even beyond this new case, since we don't distinguish between observation, facet, and existence requests, and defer the filtering to mixer. I don't think we should long-term though (this is one of the optimizations I'm hoping to look into as a follow up!)  

V2 uses the BtPlacesStatsVar cache. For country/USA, it takes a ~5 seconds with V2 and currently times out for V3, but should hopefully be faster after some refactoring (~11 seconds when I did a quick test)

internal/server/spanner/dsutil.go
n-h-diaz: Makes sense, done

n-h-diaz:
Thanks for the review!

internal/server/spanner/datasource.go
keyurva: SG. Perhaps let's log a message in that case.

internal/server/spanner/golden/query/get_observations_entity.json
keyurva: Got it - look forward to the follow up!

keyurva (approved):

n-h-diaz:
For some reason I'm not able to reply directly to your comment, but sounds good - added a log message. Thanks for the review!

PR #1559, opened 2025-04-23
fix bug in merging node responses

n-h-diaz:
Fixes an bug with merging where some data was being deleted

This was causing some custom dc node responses to be incomplete

Tested with unit tests and with local custom dc mode

### DIFF
--- a/internal/merger/merger.go
+++ b/internal/merger/merger.go
@@ -95,7 +95,9 @@ func mergeLinkedGraph(
 
 	for dcid, linkedGraph := range auxData {
 		if isEmptyLinkedGraph(linkedGraph) {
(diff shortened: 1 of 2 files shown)


internal/merger/merger.go
```diff
@@ -95,7 +95,9 @@ func mergeLinkedGraph(
 
 	for dcid, linkedGraph := range auxData {
 		if isEmptyLinkedGraph(linkedGraph) {
-			mainData[dcid] = &pbv2.LinkedGraph{}
+			if _, ok := mainData[dcid]; !ok {
```
keyurva, line 98: Very nuanced case - thanks for fixing!

keyurva (approved):

PR #1560, opened 2025-04-24
Add date filters to spanner

n-h-diaz:
This moves the date filtering into spanner, which reduces the size of the response from spanner and also means that all obs level filtering has already happened, so we can remove the time series when they're unneeded (I can do this as follow up) 

For most requests this doesn't really affect performance, but will help improve existence checks, since we'll be able to remove time series

### DIFF
--- a/internal/server/spanner/statements.go
+++ b/internal/server/spanner/statements.go
@@ -324,7 +324,7 @@ var statements = struct {
 			)result
 		INNER JOIN (%s)obs
 		ON 
-			result.object_id = obs.observation_about	
+			result.object_id = obs.observation_about
 	`,
 	searchNodesByQuery: fmt.Sprintf(`
 		GRAPH DCGraph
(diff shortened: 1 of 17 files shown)


internal/server/spanner/statements.go
```diff
@@ -251,25 +259,61 @@ var statements = struct {
...
+			import_name,
+			provenance_url
+		FROM 
+			Observation
+	`,
+	getDateObs: `
+		SELECT
+			variable_measured,
+			observation_about,
+			[obs] AS observations,
```
keyurva, line 282: Haven't seen this syntax before. For my understanding, how does this work exactly?

keyurva (approved):

internal/server/spanner/statements.go
n-h-diaz: Oh this was just to put the result back into an array. But it turns out it's not necessary for parsing the response so I just removed :) thanks for the catch!

PR #1561, opened 2025-04-25
filter observations from spanner for existence requests

n-h-diaz:
This is to make the results from spanner smaller for existence requests that don't need them 

(note: some of the tests got updated with some temperature data since I added it to the test instance, but that's unrelated to this change)

### DIFF
--- a/internal/server/spanner/golden/query/get_observations_entity.json
+++ b/internal/server/spanner/golden/query/get_observations_entity.json
@@ -3,7 +3,7 @@
     "VariableMeasured": "Count_HeatTemperatureEvent",
     "ObservationAbout": "wikidataId/Q341968",
     "Observations": {
-      "Observations": null
+      "Observations": []
     },
     "Provenance": "dc/base/TemperatureEvents_Agg",
     "ObservationPeriod": "P1M",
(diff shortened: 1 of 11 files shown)


internal/server/spanner/statements.go
```diff
@@ -291,6 +282,18 @@ var statements = struct {
...
 			UNNEST(observations) as obs
 	`,
+	allObs: `
+		observations
+	`,
+	latestObs: `
+		[observations[ARRAY_LENGTH(observations)-1]] AS observations
+	`,
+	dateObs: `
+		[obs] AS observations
```
n-h-diaz, line 292: (sorry for confusion, looks like arrays are needed after all, since some tests were failing)

internal/server/spanner/golden/query/get_observations_entity.json
```diff
@@ -3,16 +3,7 @@
...
-        {
-          "Date": "2022-09",
-          "Value": "1"
-        },
-        {
-          "Date": "2022-10",
-          "Value": "2"
-        }
-      ]
+      "Observations": null
```
keyurva, line 6: I don't think this has runtime implications, but do you know why this encodes a null value vs not encoding this field at all?

keyurva (approved):

internal/server/spanner/golden/query/get_observations_entity.json
n-h-diaz: I think this was due to how we were decoding the field https://github.com/datacommonsorg/mixer/blob/3f236ed57fd7d6e12c813478ccc73b16a7a93135/internal/server/spanner/model.go#L71 - I updated to set as [] instead

(As a follow up, I'll see if there's something we can do about the nested Observation fields) 

Thanks for the review!

PR #1564, opened 2025-05-02
Regenerate protos with protoc v3.21.12

hqpho:
This way there isn't a diff in the mixer submod when compiling protos for run_cdc_dev.sh.


n-h-diaz (approved):
Thanks!

PR #1566, opened 2025-05-19
Add defaults to node search queries

n-h-diaz:
This is to fix the broken "series" query

Also regenerates some goldens since it looks like the order was updated (unrelated to this change, possibly due to indexing)


keyurva (approved):
Thanks!

PR #1567, opened 2025-05-21
Add flag for skipping redis to v3

n-h-diaz:
Set header X-Skip-Cache: true to skip reading from redis


internal/server/redis/client.go
```diff
@@ -37,6 +37,12 @@ const (
 	protoRequestKeyPrefix = cacheKeyPrefix + "request:"
 )
 
+// Interface to allow mocking.
+type CacheClientInterface interface {
```
keyurva, line 41: nit: consider naming this `CacheClient` and the redis one `RedisCacheClient`?

internal/server/redis/processor.go
```diff
@@ -15,23 +15,33 @@
...
 	"log"
 
 	pbv2 "github.com/datacommonsorg/mixer/internal/proto/v2"
 	"github.com/datacommonsorg/mixer/internal/server/dispatcher"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/protobuf/proto"
 )
 
+const (
+	XSkipCache = "X-Skip-Cache"
```
keyurva, line 28: nit: perhaps add a comment

internal/server/redis/processor_test.go
```diff
@@ -245,3 +262,69 @@ func TestNewEmptyResponse(t *testing.T) {
...
+				CurrentResponse: test.currentResponse,
+			}
+
+			outcome, err := processor.PreProcess(rc)
+
+			if err != nil {
+				t.Errorf("Error runing TestSkipCache: %s", err)
+			}
+
+			assert.Equal(t, test.getCachedResponseCalled, m.GetCachedResponseCalled)
```
keyurva, line 323: Love these tests!

internal/server/redis/processor.go
keyurva, line 28: The documentation says the keys are lowercased: https://pkg.go.dev/google.golang.org/grpc/metadata#FromIncomingContext

It probably works as-is but something to verify.

keyurva (approved):
Thanks - this will be very handy!

internal/server/redis/client.go
n-h-diaz: Done

internal/server/redis/processor.go
n-h-diaz: Done

n-h-diaz: Gemini had said something similar! Though when I tried testing with the envoy proxy the headers weren't getting modified, so I left them as is. I can verify on autopush though, and then update if necessary

n-h-diaz:
Thanks for the review!

PR #1568, opened 2025-05-30
Switch V3 to use dc_graph_stable

n-h-diaz:
This switches Mixer to use dc_graph_stable, which has all the import groups (except dcbranch)

This also updates the queries and goldens for the Node API (including adding object_bytes) and Observation API

Note:
* The search index is still pending cherrypick release, so the NodeSearch API will not work after this change
* This doesn't yet include the isDcAggregate field, so some of the facet ids changed. As a follow up I can create a new build and then update with this field
* This build includes the nodes that are missing name/types, so some of them are dropped in the goldens. After they're properly merged/resolved I can create a new build
* Since querying within maps is till unclear, for now I just removed date filtering and am returning the full time series


internal/server/spanner/model.go
```diff
@@ -65,26 +67,27 @@ type TimeSeries []*DateValue
...
-				Value: strVal,
-			})
-		}
+	for date, value := range obs.Values {
+		*ts = append(*ts, &DateValue{
+			Date:  date,
+			Value: value,
+		})
 	}
+	sort.Slice(*ts, func(i, j int) bool {
```
keyurva, line 88: Curious why this was needed? Were the dates in the map being returned in random order? I used to think proto maps maintain insertion order but perhaps that isn't the case?

proto/storage.proto
```diff
@@ -0,0 +1,15 @@
...
+
+package org.datacommons.proto;
+
+option go_package = "github.com/datacommonsorg/mixer/internal/proto";
+
+// Includes protos that are used in spanner storage.
+// The protos are typically gzipped and stored in a bytes column.
+
+// Observations represent the observations (time series) stored in the observations column in the Observation table.
+message Observations {
```
keyurva, line 11: nit: Perhaps link to the proto in the import repo and add a note to keep them in sync.

keyurva (approved):
Thanks Natalie. This needed careful changes across the board - much appreciated!

proto/storage.proto
n-h-diaz: Good idea, done!

PR #1580, opened 2025-07-28
Upgrade golang to 1.23.11

hqpho:
v1.23 is needed to use OpenTelemetry.

After this PR is merged, I will also need to:
- Run `./scripts/push_go_protoc_image.sh latest`
- Update build/full_env/Dockerfile in the website repo
- Update build/cdc_services/Dockerfile in the website repo
- Update build/cdc/dev/Dockerfile in the website repo


internal/server/healthcheck/healthchecker.go
hqpho, line 1: A new method was added to the interface between versions.

PR #1581, opened 2025-07-30
Upgrade golangci-lint to 2.3.0 and fix or ignore existing issues

hqpho:
The upcoming upgrade of golang version causes new existing errors to surface.

While here, I updated `run_test.sh` to be similar to the version in the website repo: located in the repo root and capable of running or auto-fixing with the linter.

There is also one resolve test that fails for me even at head. Not sure how that made it through previous precommit hooks, but I updated the golden.


n-h-diaz (approved):
Thanks!

hqpho:
> Is this related to the build failure? [pantheon.corp.google.com/cloud-build/builds;region=global/54214ede-a838-405e-963b-836ec03487db?project=datcom-ci&invt=Ab4JkQ&inv=1](https://pantheon.corp.google.com/cloud-build/builds;region=global/54214ede-a838-405e-963b-836ec03487db?project=datcom-ci&invt=Ab4JkQ&inv=1) (sorry missed this before)

Resolved now! I couldn't use the "ignore before revision" strategy because when running on pull request, a Cloud Build trigger only has access to the current revision. So I added explicit ignore comments or used the auto-fixes for existing issues.

PR #1585, opened 2025-08-04
Add endpoint to filter stat vars by places

alyssasuiyiguo:
Refactors the `SearchStatvar` function to separate logic for filtering stat vars by places into its own function/endpoint. This change is required to enable Vertex AI source filtering, which will be calling the new endpoint to use the filtering logic after performing the stat var search independently. 

I tested the endpoint locally to ensure that the new endpoint works independently, and that the old `SearchStatVar` function still works as intended. From a UX standpoint, the changes are not noticeable.


proto/stat_var.proto
```diff
@@ -235,4 +235,22 @@ message SearchStatVarResponse {
...
+message StatVar {
+  string dcid = 1;
+  string name = 2;
+}
+
+message FilterStatVarsByPlaceRequest {
+  // A list of stat vars to be filtered.
+  repeated EntityInfo stat_vars = 1;
+  // repeated string stat_var_names = 2;
+  // Places to filter by.
```
gmechali, line 249: is this also going to be the source?

proto/stat_var.proto
```diff
@@ -235,4 +235,22 @@ message SearchStatVarResponse {
...
+
+message StatVar {
+  string dcid = 1;
+  string name = 2;
+}
+
+message FilterStatVarsByPlaceRequest {
+  // A list of stat vars to be filtered.
+  repeated EntityInfo stat_vars = 1;
+  // repeated string stat_var_names = 2;
```
gmechali, line 248: nit: remove this commented out line

proto/stat_var.proto
```diff
@@ -235,4 +235,22 @@ message SearchStatVarResponse {
   // A list of unique strings in the names of the results that match the search
   // tokens
   repeated string matches = 3;
+}
+
+message StatVar {
```
gmechali, line 240: is there not already an object we can use for this?

proto/service/mixer.proto
```diff
@@ -559,6 +559,14 @@ service Mixer {
     };
   }
 
+  rpc FilterStatVarsByPlace(FilterStatVarsByPlaceRequest)
+      returns (FilterStatVarsByPlaceResponse) {
+    option (google.api.http) = {
+      post : "/v1/variable/filter"
```
gmechali, line 565: lets get someone from backend to review to confirm that it s okay to add a v1 endpoint for this? V1 is kind of going away.

proto/stat_var.proto
alyssasuiyiguo: yeah - I kept the naming as "places" to be consistent with `SearchStatVar`, but it is supposed to be the sources to filter by as well. I just updated the comment, but I'm also open to renaming this entirely if you think that'd be more appropriate

internal/server/handler_v1.go
```diff
@@ -550,3 +550,29 @@ func (s *Server) SearchStatVar(
 
 	return merged, nil
 }
+
+// FilterStatVarsByPlace implements API for Mixer.FilterStatVarsByPlace.
+func (s *Server) FilterStatVarsByPlace(
+	ctx context.Context, in *pb.FilterStatVarsByPlaceRequest,
+) (*pb.FilterStatVarsByPlaceResponse, error) {
+	localResp, err := search.FilterStatVarsByPlace(ctx, in, s.store, s.cachedata.Load())
```
n-h-diaz, line 558: Can the local/remote paths be done concurrently? Something like https://github.com/datacommonsorg/mixer/blob/2214b4ee93ab9ed6a947da73d231c0b3f19cfe08/internal/server/handler_v1.go#L196

n-h-diaz (commented):
Thanks Alyssa! Excited to see the new search :) 

We're hoping to improve our existence apis to support cases like stat vars with data for a source, so this bulk filtering is a good use case that we should make sure to support! (This is fine to add for now though!)

proto/service/mixer.proto
hqpho: If we don't plan for this to be used by clients, can we exclude it from the API product on Apigee? Here is where that is configured: https://github.com/datacommonsorg/mixer/blob/2214b4ee93ab9ed6a947da73d231c0b3f19cfe08/deploy/apigee/terraform/prod/main.tf#L110

internal/merger/merger.go
```diff
@@ -452,6 +452,24 @@ func MergeSearchStatVarResponse(primary, secondary *pb.SearchStatVarResponse) *p
...
 
+// MergeFilterStatVarsByPlaceResponse merges two FilterStatVarsByPlaceResponse.
+func MergeFilterStatVarsByPlaceResponse(primary, secondary *pb.FilterStatVarsByPlaceResponse) *pb.FilterStatVarsByPlaceResponse {
+	mergedStatVars := []*pb.EntityInfo{}
+
+	if primary != nil {
+		mergedStatVars = append(mergedStatVars, primary.StatVars...)
+	}
+	if secondary != nil {
+		mergedStatVars = append(mergedStatVars, secondary.StatVars...)
```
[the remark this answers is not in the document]
alyssasuiyiguo: Yes it's base vs customDC, and I think there should not be duplicates (a similar merge function for `SearchStatVar` didn't check for stat var duplicates, so I assume the same should apply here)

proto/service/mixer.proto
alyssasuiyiguo: @hqpho what exactly does "clients" here refer to? I plan on using this endpoint in the `website` backend to filter the Vertex AI statvars, not sure if it needs to be included in the API product for that

gmechali: Lets exclude it for now because we dont yet support custom DC. Something for us to consider supporting in the future

esp/endpoints.yaml.tmpl
```diff
@@ -102,6 +102,8 @@ usage:
     allow_unregistered_calls: true
   - selector: "datacommons.Mixer.SearchStatVar"
     allow_unregistered_calls: true
+  - selector: "datacommons.Mixer.FilterStatVarsByPlace"
+    allow_unregistered_calls: true
```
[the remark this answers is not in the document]
n-h-diaz: Does it need an api key? I'm not super familiar with which endpoints should be added to this list. @hqpho do you know?

internal/merger/merger.go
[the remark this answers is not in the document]
n-h-diaz: The SearchStatVar merge does deduping https://github.com/datacommonsorg/mixer/blob/2214b4ee93ab9ed6a947da73d231c0b3f19cfe08/internal/merger/merger.go#L424-L436 - if this only ever will be used within the SearchStatVar maybe it's okay since it'll get deduped later, but since this endpoint is being exposed it might be cleaner to do it here too

proto/service/mixer.proto
```diff
@@ -559,6 +559,14 @@ service Mixer {
     };
   }
 
+  rpc FilterStatVarsByEntity(FilterStatVarsByEntityRequest)
```
n-h-diaz, line 562: nit: can you group this with the other v2 rpcs in the file

n-h-diaz (approved):
Thanks for the updates!

Since we're exposing a new api, it might be good to eventually add some e2e tests too, just to validate the expected behavior (something like https://github.com/datacommonsorg/mixer/blob/2214b4ee93ab9ed6a947da73d231c0b3f19cfe08/internal/server/statvar/golden/search_statvar_test.go), but can be follow up if easier

deploy/apigee/terraform/prod/main.tf
```diff
@@ -106,8 +106,92 @@ resource "apigee_product" "datacommons-api" {
   }
   operation {
     api_source = apigee_proxy.api[0].name
-    path       = "/"
-    methods    = [] # Accept all methods
+    path       = "/**"
```
hqpho, line 109: Alas I think this list is not comprehensive still. But no worries! An alternate approach would be to create a [policy](https://cloud.google.com/apigee/docs/api-platform/develop/policy-attachment-and-enforcement) that returns a 403 error for just this endpoint. We can use a RaiseFault policy and call it `block-internal-only-endpoints` or something.

hqpho (approved):

PR #1591, opened 2025-08-08
Switch to new spanner instance dc_graph_15

n-h-diaz:
Mostly some GQL changes due to spanner schema changes and the relevant updated goldens (the query responses have changed, but the API responses should mostly be the same) 

Also
* Sets is_dc_aggregate field
* Uses dataflow-generated facet ids instead of mixer-generated
* Switches NodeSearch to use node name_tokenlist

Note that some of the random changes to/from "Thing" type are unrelated to this change and are due to an existing bug relating to nondeterminism in writing nodes to spanner. (I commented on one example of what this looks like) I'll look into fixing this as a follow up, see the bug for more details


internal/server/spanner/golden/query/get_node_edges_out_bracket_props.json
```diff
@@ -15,22 +14,20 @@
...
-      "ObjectValue": "",
-      "ObjectBytes": null,
       "Provenance": "dc/base/BaseGeos",
-      "Name": "Franklin",
+      "Value": "geoId/51620",
+      "Bytes": null,
+      "Name": "",
       "Types": [
-        "County"
+        "Thing"
```
n-h-diaz, line 22: This is an example of the merging issue where depending on the order nodes are written, some nodes end up with default type "Thing"

internal/server/spanner/model.go
```diff
@@ -30,28 +30,29 @@ type Property struct {
...
-	Provenance  string   `spanner:"provenance"`
-	Name        string   `spanner:"name"`
-	Types       []string `spanner:"types"`
+	SubjectID  string   `spanner:"subject_id"`
+	Predicate  string   `spanner:"predicate"`
+	Provenance string   `spanner:"provenance"`
+	Value      string   `spanner:"value"`
+	Bytes      []byte   `spanner:"bytes"`
+	Name       string   `spanner:"name"`
+	Types      []string `spanner:"types"`
```
keyurva, line 41: I understand we're now using empty types as an indicator of terminal values in the DB. I'm wondering if using the same convention in code is better or being more explicit about it is.

Not suggesting that we address it in this PR but something to do in a follow-on if a different representation makes things easier in code.

internal/server/spanner/query.go
keyurva, line 1: One overall point on query testing - currently the tests test for query execution outputs. Since the queries can get quite complicated, I'm wondering introducing tests that test the final queries themselves are useful? It may help with debugging, running the final query in spanner studio and also sharing them internally or with spanner folks.

Again not for this PR but something to think about for future maintenance and troubleshooting.

keyurva (approved):
Thanks Natalie!

internal/server/spanner/query.go
```diff
@@ -100,42 +97,67 @@ func (sc *SpannerClient) GetNodeEdgesByID(ctx context.Context, ids []string, arc
...
 
+	// Generate filters.
+	returnEdges := ""
+	i := 0
+	for prop, val := range arc.Filter {
+		params["prop"+strconv.Itoa(i)] = prop
+		objectFilter := ""
+		filterVal := val
+		for _, v := range val {
+			filterVal = append(filterVal, generateValueHash(v))
```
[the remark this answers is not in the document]
n-h-diaz: It's new - previously we would search both object_id and object_value fields

internal/server/spanner/model.go
n-h-diaz: Noted - I added some more comments clarifying this convention, but if it gets confusing we can update the model to be more explicit

n-h-diaz:
Thanks for the review!

PR #1600, opened 2025-08-20
[cleanup] refactor spanner query tests

n-h-diaz:
* Moves test cases for query_test and query_builder_test into shared file (so both tests will automatically test the same cases) 
* Moves some shared test logic for each test into helper functions


internal/server/spanner/golden/query_test_cases.go
```diff
@@ -0,0 +1,284 @@
...
+
+	"github.com/datacommonsorg/mixer/internal/server/spanner"
+	v2 "github.com/datacommonsorg/mixer/internal/server/v2"
+)
+
+// The goldenTestFunc type represents a function that can be tested with the golden file pattern.
+// It returns the actual result as an interface and an error.
+type goldenTestFunc func(ctx context.Context) (interface{}, error)
+
+var nodePropsTestCases = []struct {
```
keyurva, line 29: I like this approach of assigning test cases to a variable vs inlining it in tests. Wondering if we should adopt this elsewhere also (not en masse but any new tests or tests we touch).

keyurva (approved):
Thanks!

internal/server/spanner/golden/query_test_cases.go
n-h-diaz: I think that sounds good! Some of the existing tests are quite long, so it could help with readability if we cleaned them up a bit 

Thanks for the review!

PR #1828, opened 2026-03-27
Implement V3 Coordinate-to-DCID Resolution using Spanner and S2Cells level 10

rohitkumarbhagat:
### Summary
This PR implements the `V3Resolve` path for converting latitude/longitude coordinates to Place DCIDs (via S2 level-10 cells) using the Spanner graph backend. It also introduces a centralized "dominant place type" priority scheme to break ties when multiple places are returned.

### Key Logic & Changes

1.  **S2 Cell Mapping (Spanner)**:
    *   Converts `"lat#lng"` nodes into level-10 S2 cell IDs.
    *   Queries `containedInPlace` edges for those cells from the Spanner `Edge` table.
2.  **Dominant Type Filtering**:
    *   Loads priority-based rules (e.g., Country > State > County > City) from a new `place_types.yaml` configuration.
    *   Filters and sorts candidate places by their dominant type when returned from Spanner.
    *  `place_types.yaml` - yaml format to allow comments
3.  **Regression Testing**:
    *   Expanded `V3Resolve` tests with 4 target coordinates and type filter regressions. 

### Verification Results
All tests pass locally for V3 Spanner:
```bash
ENABLE_SPANNER_GRAPH=true go test ./internal/server/v3/resolve/golden
```


internal/util/place_type.go
```diff
@@ -0,0 +1,61 @@
...
+			return
+		}
+	})
+
+	// 1. Check non-place types first (PoliticalCampaignCmte, Currency)
+	for _, t := range []string{"PoliticalCampaignCmte", "Currency"} {
+		if StringContainedIn(t, types) {
+			return t
+		}
+	}
```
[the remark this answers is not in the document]
rohitkumarbhagat: The list is small. Can be moved to file once it grows.

internal/server/spanner/datasource.go
```diff
@@ -260,6 +266,83 @@ func (sds *SpannerDataSource) Sparql(ctx context.Context, req *pb.SparqlRequest)
...
+		}
+		cellID := level10S2CellID(lat, lng)
+		coordinateNodes = append(coordinateNodes, coordinateNode{
+			node:   node,
+			cellID: cellID,
+		})
+		cellIDSet[cellID] = struct{}{}
+	}
+
+	containedInPlaceArc := &v2.Arc{SingleProp: "containedInPlace", Out: true}
```
keyurva: nit: Is there a const for `"containedInPlace"` that we can use?

internal/server/spanner/datasource.go
```diff
@@ -260,6 +266,83 @@ func (sds *SpannerDataSource) Sparql(ctx context.Context, req *pb.SparqlRequest)
...
+	if err != nil {
+		return nil, err
+	}
+
+	resp := &pbv2.ResolveResponse{}
+	for _, coordinateNode := range coordinateNodes {
+		candidateSet := map[string]struct{}{}
+		places := []*pb.ResolveCoordinatesResponse_Place{}
+		for _, edge := range cellToEdges[coordinateNode.cellID] {
+			if edge.Value == "" || strings.HasPrefix(edge.Value, "s2CellId/") {
```
keyurva: nit: can you add a comment why the prefix is skipped? Also, if you can make it a const as well.

internal/server/spanner/datasource.go
```diff
@@ -260,6 +266,83 @@ func (sds *SpannerDataSource) Sparql(ctx context.Context, req *pb.SparqlRequest)
...
+				continue
+			}
+			dominantType, err := util.GetDominantType(edge.Types)
+			if err != nil {
+				return nil, err
+			}
+			// Coordinate resolve filters by dominant type only. Secondary types on
+			// the place node do not qualify a candidate for a typeOf match.
+			if len(typeOfs) > 0 && !matchesRequestedType(dominantType, typeOfs) {
```
keyurva, line 329: Does this mean that if a typeOf filter is specified, unless it is the dominant type, it will be skipped?

internal/server/spanner/datasource.go
```diff
@@ -260,6 +266,83 @@ func (sds *SpannerDataSource) Sparql(ctx context.Context, req *pb.SparqlRequest)
...
+
+	typeOfs := inArc.Filter["typeOf"]
+	coordinateNodes := []coordinateNode{}
+	cellIDSet := map[string]struct{}{}
+	for _, node := range req.GetNodes() {
+		lat, lng, err := resolvev2.ParseCoordinate(node)
+		if err != nil {
+			return nil, err
+		}
+		cellID := level10S2CellID(lat, lng)
```
keyurva, line 290: Can you explain how this works for a large geography like say California or Europe. These potentially cover a large number of level 10 cells. Are we storing them all in the KG?

internal/util/resource/place_types.yaml
```diff
@@ -0,0 +1,102 @@
+# Priority list of place types for Data Commons
+# Top items have highest priority
+
+- Continent
```
keyurva, line 4: Does it make sense to start maintaining configs like these in the KG itself?

Not necessary right away but bringing it up if it's a possibility.

keyurva (commented):
Very nicely done! Thanks Rohit.

I'll setup a sync with you to go over some of my questions so we can get this in sooner.

internal/server/spanner/datasource.go
rohitkumarbhagat: Could not find a reusable constant. Created one in constants.go.

internal/server/spanner/datasource.go
rohitkumarbhagat: Added a comment and constants for the S2 skip logic.

  I also switched the check to prefer the destination node Types over the s2CellId/ prefix, since that’s less
  brittle and matches the actual graph semantics.

keyurva (approved):
Approving this PR but chatting offline for improvements to s2cell management.

internal/server/spanner/datasource.go
rohitkumarbhagat: Yes, adopted this for backward compatability.

internal/server/spanner/datasource.go
rohitkumarbhagat: Yes. DC stores all level 10 s2cell -> contained in place mapping. 
lat,longt -> Level 10 s2cell is deduced inprocess.

internal/util/resource/place_types.yaml
rohitkumarbhagat: Maybe eventually, if we need a shared precedence across consumers. But this does not represent a real-
  world fact, so the KG probably is not the right place for it right now. dominant_type is mainly a
  website-facing/caller-facing concept today, and ideally we should deprecate that from Mixer and let
  callers like the website define their own ranking/tie-break logic. For now, keeping it as a small
  checked-in config seems simpler.

rohitkumarbhagat:
Thanks Keyur for the review and useful suggestions.

PR #1829, opened 2026-03-27
fix: add missing OR operator to block-deprecated-endpoints condition in Apigee target endpoint template

rohitkumarbhagat:


juliawu (approved):
Thanks for the fix!

PR #1832, opened 2026-04-03
Add spanner queries for fetching descendent stat var count

n-h-diaz:
Supports initial query patterns for BulkVariableGroupInfo, including

SVG/Topic nodes
Place/Dataset/Source constrainedEntities
numEntitiesExistence


internal/server/spanner/golden/query_builder/count_descendent_stat_vars_import.sql
```diff
@@ -0,0 +1,17 @@
+		SELECT
+			e.object_id,
+			COUNT(e.subject_id) AS descendent_stat_vars
+		FROM Edge@{FORCE_INDEX=InEdge} e
```
keyurva, line 4: Not a big fan of having to provide such hints. Not a blocker for this PR but should work with the spanner team to see if this can be avoided.

internal/server/spanner/query_builder.go
```diff
@@ -390,6 +390,40 @@ func GetCacheDataQuery(typeFilter CacheDataType, keys []string) *spanner.Stateme
 	}
 }
 
+func CountDescendentStatVarsQuery(nodes []string, constrainedEntities []string, numEntitiesExistence int, filterProp string) *spanner.Statement {
```
keyurva, line 394: nit: Please add a method comment.

internal/server/spanner/query_builder.go
```diff
@@ -390,6 +390,40 @@ func GetCacheDataQuery(typeFilter CacheDataType, keys []string) *spanner.Stateme
...
+	nodeFilter, nodeVal := getParamStatement("nodes", nodes)
+	params := map[string]interface{}{
+		"nodes":                nodeVal,
+		"numEntitiesExistence": numEntitiesExistence,
+	}
+
+	var entityFilter string
+	var distinct string
+	if filterProp == "isPartOf" || filterProp == "source" {
```
keyurva: Can you add a comment on what's special about these filters? And any way to generalize this?

keyurva (approved):
Thanks!

internal/server/spanner/query_builder.go
n-h-diaz: done, thanks for the review!

internal/server/spanner/query_builder.go
n-h-diaz: added a comment and removed hardcoded filters

internal/server/spanner/golden/query_builder/count_descendent_stat_vars_import.sql
n-h-diaz: query optimizer doesn't always pick best plan, especially for parameterized query, so adding hints helps significantly improve performance in some cases (but agree that it's be ideal if optimizer could pick better plans without us forcing the plan)

keyurva: Interesting. Big picture, is hints the way to go for other parameterized queries as well until spanner has a fix for it?

PR #1837, opened 2026-04-08
Add v2 wrapper of BulkVariableGroupInfo

n-h-diaz:
This is to unblock frontend changes until the Spanner implementation is ready.

It calls the V1 implementation and modifies the response to what we expect from the Spanner implementation.

Note: I also deleted some test files related to /v1/recognize/entities. It was marked as deprecated in the proto, causing the lint tests to fail


internal/server/handler_v2.go
```diff
@@ -648,6 +648,33 @@ func (s *Server) V2BulkVariableInfo(
...
+		return nil, err
+	}
+
+	// The new response will not contain all legacy V1 fields.
+	// To ensure the new response is sufficient, clear all legacy fields until swapping over to the new backend.
+	for _, info := range v2Resp.GetData() {
+		if info.GetInfo() == nil {
+			continue
+		}
+		info.Info.ParentStatVarGroups = nil
```
keyurva: nit: Consider putting this in a helper method. Useful if we identify more such fields.

keyurva (approved):

internal/server/handler_v2.go
n-h-diaz: done, thanks for the review!

PR #1839, opened 2026-04-10
Implement normalized Spanner schema for observations

keyurva:
* Added statements, query builder, and execution logic for the new normalized Spanner schema.
* Implemented support for all observation related methods for the new normalized schema:
   + `GetObservations`
   + `CheckVariableExistence`
   + `GetObservationsContainedInPlace`
* The following methods are not currently used by API calls and are *not* implemented for the new schema. We can add them once their use is finalized:
  + `getStatVarGroupNode`
  + `getFilteredChildSVGs`
  + `getFilteredTopic`
* Iterated on queries for the normalized schema and eventually landed on using array subqueries and index hints to bring latencies down from ~5-10s to ~0.5s (on par with current schema).
* Added a request-time selector to invoke the default or normalized schema based on a `X-Use-Normalized-Schema` header.
* Added support for accessing the raw Spanner client directly to avoid breaking existing tests.
* Added unit and integration tests to verify both schema paths.


internal/server/spanner/client_normalized.go
```diff
@@ -0,0 +1,36 @@
...
+// to reuse internal helpers like queryStructs. This is a compromise to avoid
+// exporting internal implementation details while allowing tests in the golden package
+// to construct it.
+func NewNormalizedClient(client SpannerClient) *normalizedClient {
+	sc, ok := client.(*spannerDatabaseClient)
+	if !ok {
+		panic("NewNormalizedClient: expected *spannerDatabaseClient")
+	}
+	return &normalizedClient{sc: sc}
+}
```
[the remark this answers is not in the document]
keyurva: Done.

internal/server/spanner/query_normalized.go
```diff
@@ -0,0 +1,126 @@
...
+			case "observationPeriod":
+				obs.ObservationPeriod = attr.Value
+			case "measurementMethod":
+				obs.MeasurementMethod = attr.Value
+			case "unit":
+				obs.Unit = attr.Value
+			case "scalingFactor":
+				obs.ScalingFactor = attr.Value
+			case "isDcAggregate":
+				obs.IsDcAggregate, _ = strconv.ParseBool(attr.Value)
```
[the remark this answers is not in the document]
keyurva: Done.

internal/server/spanner/query_normalized.go
```diff
@@ -0,0 +1,126 @@
...
+		return nil, fmt.Errorf("entity must be specified")
+	}
+
+	rawObs, err := nc.fetchRawObservations(ctx, variables, entities)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(rawObs) == 0 {
+		return nil, nil
```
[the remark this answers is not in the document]
keyurva: Done.

internal/server/spanner/golden/query_builder/check_variable_existence_normalized.sql
```diff
@@ -0,0 +1,8 @@
+		SELECT DISTINCT
+			ts.variable_measured,
+			a.value AS entity
+		FROM 
+			TimeSeries@{FORCE_INDEX=TimeSeriesByVariableMeasured} ts
+		JOIN
+			TimeSeriesAttribute@{FORCE_INDEX=TimeSeriesAttributePropertyValue} a ON ts.id = a.id
```
n-h-diaz: since we aren't filtering by property in the query, could we index by just value? or is it more efficient this way? (just curious)

internal/server/spanner/golden/query_builder/get_normalized_obs_basic.sql
```diff
@@ -0,0 +1,12 @@
+		SELECT
```
n-h-diaz, line 1: it looks like this pattern isn't being used anymore - can these be cleaned up? (as in, fetch ids and obs, then fetch attributes)

n-h-diaz (approved):
Thanks!

not needed for this pr, but some things to think about:
- this new schema should (I think) make it much more efficient to do filtering in spanner directly (which is currently a huge limitation of the previous schema), in particular, doing the date filtering directly vs fetching the full time series or doing facet filtering, maybe something to explore 
- can these queries efficiently be extended to also retrieve observation attributes

internal/server/spanner/golden/query_builder/get_normalized_obs_basic.sql
keyurva: I actually ran into performance issues (like combinatorial explosion or N+1 queries) when I tried to split these up or use large joins. That's why settled on using array subqueries here to pull everything in a single row per time series.

So keeping it as-is for now. But happy to discuss and update if there's a more efficient approach.

internal/server/spanner/golden/query_builder/check_variable_existence_normalized.sql
keyurva: Done. Thanks for the suggestion - yes only value needed here!

keyurva:
> Thanks!
> 
> not needed for this pr, but some things to think about:
> 
> * this new schema should (I think) make it much more efficient to do filtering in spanner directly (which is currently a huge limitation of the previous schema), in particular, doing the date filtering directly vs fetching the full time series or doing facet filtering, maybe something to explore

Agree. The goal of this PR was to make the current requests work with minimal changes. But yes, we should use the new schema to make queries more efficient where we can.

> * can these queries efficiently be extended to also retrieve observation attributes

We should do this once we have some data with obs attributes. Is this something we should chat with the data team?

PR #1840, opened 2026-04-10
Data release: release-20260410

ajaits:
Rollback place import to revert place names changes in https://github.com/datacommonsorg/website/pull/6192


dhotchkiss (approved):
Thanks Ajai, lgtm!

PR #1847, opened 2026-04-15
Add query hint to observation queries

n-h-diaz:
This is to help query optimizer pick a better plan - in this case base table is better than using other indexes


nick-nlb (approved):
LGTM (and looks like the Codacy failure is a false positive and can be ignored).

PR #1858, opened 2026-04-17
Add dispatcher diversion to resolve

n-h-diaz:
This PR adds diversion to the dispatcher for V2Resolve

It also updates the dispatcher resolve with some of the helper logic in the existing V2 resolve, so this (and corresponding tests) is moved into the resolve package, including
- Using ResolveRouting to specify which sources to fetch from using target (here I'm filtering this at the datasource level, since a given datasource doesn't know about the remoteMixerDomain) 
- Using ValidateAndParseResolveInputs - updates function to return a cloned request rather than mutating the original request due to parallel access to the same request (see gemini comments) 

This also fixes a small bug in description resolution where only the first type was fetched. There are also a few automatic lint fixes to modified files

Note: For now I have not yet migrated the embeddings resolution. I'd be curious to hear thoughts on how to handle this (since it sounds like spanner might have it's own embeddings resolution soon?) One thought is I can create an EmbeddingsResolver that does the ResolveUsingEmbeddings and that I can provide to relevant datasources. For now, spanner will just return an empty response if ResolveResolverIndicator is set. I can update in a follow up PR based on feedback if needed


internal/server/spanner/datasource.go
```diff
@@ -208,47 +209,24 @@ func (sds *SpannerDataSource) NodeSearch(ctx context.Context, req *pbv2.NodeSear
 
 // Resolve searches for nodes in the graph.
 func (sds *SpannerDataSource) Resolve(ctx context.Context, req *pbv2.ResolveRequest) (*pbv2.ResolveResponse, error) {
-	arcs, err := v2.ParseProperty(req.GetProperty())
+	inProp, outProp, typeOfValues, err := resolvev2.ValidateAndParseResolveInputs(req)
```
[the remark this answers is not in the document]
n-h-diaz: Updated to clone and return the modified request

internal/server/v2/resolve/resolve.go
```diff
@@ -237,3 +265,168 @@ func GetSortedResolvedPlaceCandidates(
...
+			}
+		}
+	}
+
+	if len(validationErrors) > 0 {
+		return "", "", nil, status.Errorf(codes.InvalidArgument, "Invalid inputs in request. %s", strings.Join(validationErrors, ". "))
+	}
+
+	return inProp, outProp, typeOfValues, nil
+}
```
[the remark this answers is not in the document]
n-h-diaz: Updated to clone and return the modified request

internal/server/v2/resolve/resolve.go
```diff
@@ -237,3 +266,170 @@ func GetSortedResolvedPlaceCandidates(
...
+			typeOfValues = filter
+		}
+	}
+
+	return inArc.SingleProp, outArc.SingleProp, typeOfValues, nil
+}
+
+// resolveRouting determines whether to route to local and/or remote instances
+// based on the target parameter and the presence of a remote mixer domain.
+// Returns (shouldCallLocal, shouldCallRemote, error).
```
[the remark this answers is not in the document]
n-h-diaz: done

internal/server/spanner/datasource.go
```diff
@@ -208,47 +209,24 @@ func (sds *SpannerDataSource) NodeSearch(ctx context.Context, req *pbv2.NodeSear
...
+	normalizedReq, inProp, outProp, typeOfValues, err := resolvev2.ValidateAndParseResolveInputs(req)
 	if err != nil {
 		return nil, err
 	}
 
-	if len(arcs) != 2 {
-		return nil, status.Errorf(codes.InvalidArgument,
-			"invalid property for resolving: %s", req.GetProperty())
+	if resolver := normalizedReq.GetResolver(); resolver == resolvev2.ResolveResolverIndicator {
+		// Spanner doesn't do embeddings resolution yet.
```
keyurva, line 218: nit: Consider logging to avoid silent failures

internal/server/datasources/datasources.go
```diff
@@ -168,3 +171,29 @@ func (ds *DataSources) BulkVariableGroupInfo(ctx context.Context, in *pbv1.BulkV
...
+	for _, source := range sources {
+		if source.Type() == datasource.TypeRemote {
+			hasRemoteMixerDomain = true
+			break
+		}
+	}
+
+	callLocal, callRemote := resolve.ResolveRouting(in.GetTarget(), hasRemoteMixerDomain)
+	var filteredSources []datasource.DataSource
+	for _, source := range sources {
```
keyurva: To simplify the loop, if we maintain `ds.remoteDataSource`, we could do:
```go
for _, source := range sources {
    isRemote := source == ds.remoteDataSource
    if (isRemote && callRemote) || (!isRemote && callLocal) {
        filteredSources = append(filteredSources, source)
    }
}
```

internal/server/datasources/datasources.go
```diff
@@ -168,3 +171,29 @@ func (ds *DataSources) BulkVariableGroupInfo(ctx context.Context, in *pbv1.BulkV
 		},
 	)
 }
+
+// Resolve API specifies which sources to call in the target input params.
+// filterResolveSources filters sources accordingly.
+func filterResolveSources(sources []datasource.DataSource, in *pbv2.ResolveRequest) []datasource.DataSource {
+	hasRemoteMixerDomain := false
+	for _, source := range sources {
+		if source.Type() == datasource.TypeRemote {
```
keyurva: Can we maintain a pointer to the remote datasource in the `DataSources` struct? This loop won't be required if we do that.

internal/server/spanner/datasource.go
```diff
@@ -208,47 +209,24 @@ func (sds *SpannerDataSource) NodeSearch(ctx context.Context, req *pbv2.NodeSear
...
+		return &pbv2.ResolveResponse{}, nil
 	}
 
-	inArc := arcs[0]
-	outArc := arcs[1]
-	if inArc.Out || !outArc.Out {
-		return nil, status.Errorf(codes.InvalidArgument,
-			"invalid property for resolving: %s", req.GetProperty())
+	switch inProp {
```
keyurva: Nice - much cleaner than multiple if statements!

internal/server/v2/resolve/resolve.go
```diff
@@ -237,3 +266,170 @@ func GetSortedResolvedPlaceCandidates(
...
+//   - "place": if "inProp" is "description" or "geoCoordinate", "outProp" must be "dcid".
+//   - "indicator": "inProp" must be "description" and "outProp" must be "dcid".
+//
+// Returns:
+// - normalized request (ResolveRequest)
+// - input property (string)
+// - output property (string)
+// - typeOf filter values ([]string, from the input property filter)
+// - error if validation fails
+func ValidateAndParseResolveInputs(in *pbv2.ResolveRequest) (*pbv2.ResolveRequest, string, string, []string, error) {
```
keyurva: nit: Consider encapsulating the returned args in a struct for better readability

internal/server/v2/resolve/resolve.go
```diff
@@ -237,3 +266,170 @@ func GetSortedResolvedPlaceCandidates(
...
+	normalizedRequest := proto.Clone(in).(*pbv2.ResolveRequest)
+	var validationErrors []string
+
+	// Parse and validate `target`
```
keyurva, line 295: nit: Consider extracting validation of individual components (target, resolver, etc.) into separate helper methods.

keyurva (approved):
Thanks Natalie - I like PRs like these that simplify things!

On the `EmbeddingsResolver` question: I'd like to understand more. Let's sync offline.

internal/server/spanner/datasource.go
n-h-diaz: Done

internal/server/datasources/datasources.go
n-h-diaz: Done

internal/server/datasources/datasources.go
n-h-diaz: Nice! Added

internal/server/spanner/datasource.go
n-h-diaz: Yes, thanks Christie for adding switch logic!

internal/server/v2/resolve/resolve.go
n-h-diaz: Done

internal/server/v2/resolve/resolve.go
n-h-diaz: Done

PR #1859, opened 2026-04-17
[DCP Ingestion - No Downtime] Use the correct error check

gmechali:
Verified by deplyoing CDC image in DCP.


n-h-diaz (approved):
optional: gemini suggestion if it helps with readability

PR #1861, opened 2026-04-20
Enable stale reads for spanner

n-h-diaz:
Stale reads will be used for spanner data consistency, so this PR consolidates the flag with other spanner features - it will always be used when spanner is enabled

### DIFF
--- a/deploy/featureflags/prod.yaml
+++ b/deploy/featureflags/prod.yaml
@@ -7,4 +7,3 @@ flags:
   WriteUsageLogs: 1.0
   UseSpannerGraph: true
   SpannerGraphDatabase: dc_graph_2025_11_07
-  UseStaleReads: true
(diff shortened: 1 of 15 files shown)


deploy/featureflags/prod.yaml
```diff
@@ -7,4 +7,3 @@ flags:
   WriteUsageLogs: 1.0
   UseSpannerGraph: true
   SpannerGraphDatabase: dc_graph_2025_11_07
-  UseStaleReads: true
```
gmechali, line 10: These feature flags deploy immediately.

With feature flags, we don't want the changes to be in the same PR as the code so that PR reverting doesn't impact prod.
Can you make a separate PR for all the featureflag changes. And Production should also be in its own PR for extra safety. That way we can also test in staging before doing it in prod.

gmechali (approved):
Code LGTM! Just please put the feautre flag changes in a separate PR

gmechali:
Thanks for doing this. This will actually simplify some of the PRs I have out currently with https://github.com/datacommonsorg/website/pull/6215 and https://github.com/datacommonsorg/datacommons/pull/31/changes

deploy/featureflags/prod.yaml
n-h-diaz: Ah good point, I should wait for the code changes to roll out

reverted all flag flips and just included code changes for now 

Thanks for the review!

PR #1863, opened 2026-04-21
Block node property endpoints in Apigee configuration

rohitkumarbhagat:


n-h-diaz (approved):
Thanks! 

nit: could we try to keep this list alphabetized

PR #1864, opened 2026-04-21
Spanner resolver integration

shixiao-coder:
This PR add a new resolver type embedding and use this embedding type to resolve on nodes by node type

It invokes the configurations coded in v2/resolve
pass the cfg to the requests and invokes embedding generation statements then vector_search statement to find search result per node

And at last generate the standard resolve output from above nodes search results

The code is tested against a copied DB (https://pantheon.corp.google.com/spanner/instances/dc-kg-test/databases/dc_graph_search_experiments/) from dc_graph_2026_01_27,

whereas a local mixer is triggered with below command:
`
# In repo root directory
export MIXER_API_KEY=<YOUR API KEY>
./run_server.sh \
    --use_base_bigtable=false \
    --use_branch_bigtable=false \
    --use_bigquery=false \
    --use_spanner_graph=true \
    --spanner_graph_info="$(cat deploy/storage/spanner_graph_info.yaml)" 
`
with updated local.yaml and spanner_graph_info.yaml to point to dc_graph_search_experiments and EnableSpannerSearchEmbeddings to true in feature flag

and with API calls: http://localhost:8081/v2/resolve?nodes=population&resolver=indicator

we are able to receive below json data:

{
  "entities": [
    {
      "node": "population",
      "candidates": [
        {
          "dcid": "WHO/RS_1845",
          "metadata": {
            "sentence": "Population",
            "score": "0.9921"
          },
          "typeOf": [
            "StatisticalVariable"
          ]
        },
        {
          "dcid": "Count_Person",
          "metadata": {
            "score": "0.8950",
            "sentence": "Total population"
          },
          "typeOf": [
            "StatisticalVariable"
          ]
        },
        ...
     ]
  ]
{


internal/server/spanner/golden/query_builder/vector_search_node.sql
```diff
@@ -1,12 +1,16 @@
 		SELECT
 			subject_id,
-			embedding_content,
+			embedding_content AS name,
+			types,
```
n-h-diaz, line 4: is this the same types as Node.types? if so, is there a way to use that in the query efficiently?

internal/server/spanner/datasource.go
```diff
@@ -230,6 +232,81 @@ func (sds *SpannerDataSource) Resolve(ctx context.Context, req *pbv2.ResolveRequ
...
+				TypeOf: res.Types,
+				Metadata: map[string]string{
+					"score":    fmt.Sprintf("%.4f", res.CosineSimilarity),
+					"sentence": res.Name,
+				},
+			})
+		}
+		entity.Candidates = candidates
+		resolveResponse.Entities = append(resolveResponse.Entities, entity)
+	}
```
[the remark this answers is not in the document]
n-h-diaz: it looks like the two requests need to be sequential, but can the requests for each node be parallelized?

internal/server/spanner/coordinate_resolve_test.go
```diff
@@ -335,3 +345,78 @@ func TestResolveCoordinateSkipsS2CellCandidatesByType(t *testing.T) {
 		t.Fatalf("Resolve() diff (-want +got):\n%s", diff)
 	}
 }
+
+func TestResolveEmbeddingsEmpty(t *testing.T) {
```
n-h-diaz: nit: since this file is called coordinate_resolve_test, could we either move these tests to a different file or rename the file?

(thanks for adding the tests btw!)

internal/server/spanner/datasource.go
```diff
@@ -230,6 +231,85 @@ func (sds *SpannerDataSource) Resolve(ctx context.Context, req *pbv2.ResolveRequ
...
+	cfg := sds.searchConfig
+	if cfg == nil {
+		return nil, status.Errorf(codes.Internal, "failed to load search config")
+	}
+
+	resolveResponse := &pbv2.ResolveResponse{
+		Entities: []*pbv2.ResolveResponse_Entity{},
+	}
+	typeOfs := req.TypeOfValues
+	if len(typeOfs) == 0 {
```
n-h-diaz, line 272: it looks like now we'd also support requesting typeOf beyond just StatisticalVariable and Topic - are you planning on adding embeddings to all nodes? (and if not, maybe can check for this, or if not yet, maybe we could either log or add a message saying other types aren't supported)

internal/server/spanner/datasource.go
```diff
@@ -215,9 +218,7 @@ func (sds *SpannerDataSource) Resolve(ctx context.Context, req *pbv2.ResolveRequ
 	}
 
 	if resolver := normalizedResolveRequest.Request.GetResolver(); resolver == resolvev2.ResolveResolverIndicator {
-		// Spanner doesn't do embeddings resolution yet.
-		slog.Warn("Received unsupported ResolveResolverIndicator request to Spanner", "request", req)
-		return &pbv2.ResolveResponse{}, nil
+		return sds.resolveEmbeddings(ctx, normalizedResolveRequest)
```
n-h-diaz: when I last checked the NodeEmbeddings table doesn't exist yet on the running database, just want to double check this gets handled safely (so there's no crash since dev is now running on spanner)

n-h-diaz (commented):
Thanks! overall looks good, had a few small comments/questions

internal/server/spanner/golden/query_builder/vector_search_node.sql
shixiao-coder: do we have multiple types in a single node? If not we can just check if the types[0] is in the types filter this would avoid the unnest operation

internal/server/spanner/datasource.go
shixiao-coder: we are planning to add more nodes, as a starting point we would just includes SVs and Topics to replicate how current v2/resolve works with indicator type resolver

internal/server/spanner/datasource.go
shixiao-coder: Which DB are we running on in dev? I'm adding the ingestion code for embeddings as well as initialization logic, would be good to test and update the DB for data and embeddings ingestion

internal/server/spanner/datasource.go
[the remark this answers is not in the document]
shixiao-coder: Added parallel logic

internal/server/spanner/coordinate_resolve_test.go
shixiao-coder: moved to a new test file

internal/server/spanner/datasource.go
shixiao-coder: also added a log when we of typeOf outside SV and Topics

internal/server/spanner/golden/query_builder/vector_search_node.sql
n-h-diaz: a node could have multiple types (for example, geoId/06 is both State and AdministrativeArea1)

to clarify, I was just curious if we could efficiently use the parent table Node.types in this query instead of duplicating the types field in the child NodeEmbeddings table?

internal/server/spanner/datasource.go
n-h-diaz: dev is using 2026_01_27 (all env except prod are using this one, prod is still using 2025_11_07 to test incremental ingestion at scale)

to be safe, I think we should either create the new tables (even if empty) or flag gate this before enabling

internal/server/spanner/golden/query_builder/vector_search_node.sql
shixiao-coder: I did a scan with the above condition compared with ARRAYS_OVERLAP function, whereas the UNNEST is much faster

internal/server/spanner/datasource.go
shixiao-coder: Sure, I'll add a feature flag to gate this logic and in the meanwhile I'm building local mixer to test against a test DB to validate the function. Will update the 2026_01_27 once local test is good

shixiao-coder: feature flag added, local test could now properly return resolve result under the same format

internal/server/spanner/datasource.go
```diff
@@ -57,11 +60,15 @@ func NewSpannerDataSource(
 	client SpannerClient,
 	recogPlaceStore *files.RecogPlaceStore,
 	mapsClient internalmaps.MapsClient,
+	enableSpannerSearchEmbeddings bool,
```
clincoln8, line 70: I think this is okay for now, but does it make sense to just pass in the flags block to datasources instead of passing individual flags? @n-h-diaz curious about your thoughts on this.

internal/server/spanner/datasource.go
```diff
@@ -230,6 +239,104 @@ func (sds *SpannerDataSource) Resolve(ctx context.Context, req *pbv2.ResolveRequ
...
+func loadSpannerSearchConfig() (*SpannerSearchConfig, error) {
+	cfgPath := GetSpannerSearchConfigPath("default")
+	if cfgPath == "" {
+		return nil, fmt.Errorf("failed to get search config path")
+	}
+	return ReadSpannerSearchConfig(cfgPath)
+}
+
+// resolveEmbeddings resolves nodes using Spanner vector search.
+func (sds *SpannerDataSource) resolveEmbeddings(
```
clincoln8: nit: what about calling this method vectorSearchResolution (or something along those lines) instead of calling it resolveEmbeddings in order to be more clear what this actually does and to not get confused with the legacy embeddings server

internal/server/spanner/datasource.go
```diff
@@ -230,6 +239,104 @@ func (sds *SpannerDataSource) Resolve(ctx context.Context, req *pbv2.ResolveRequ
...
+	resolveResponse := &pbv2.ResolveResponse{
+		Entities: []*pbv2.ResolveResponse_Entity{},
+	}
+	typeOfs := req.TypeOfValues
+	if len(typeOfs) == 0 {
+		typeOfs = []string{"StatisticalVariable", "Topic"}
+	} else {
+		for _, t := range typeOfs {
+			if t != "StatisticalVariable" && t != "Topic" {
+				slog.Warn("Embeddings resolution requested for unsupported type. Current support is only for StatisticalVariable and Topic.", "type", t)
```
clincoln8, line 277: should we return early here, an empty response?

internal/featureflags/featureflags.go
```diff
@@ -46,20 +46,23 @@ type Flags struct {
 	UseStatisticalCalculation bool `yaml:"UseStatisticalCalculation"`
 	// Whether to enable the SDMX API endpoint.
 	EnableSDMXDataApi bool `yaml:"EnableSDMXDataApi"`
+	// Whether to enable Spanner search embeddings.
+	EnableSpannerSearchEmbeddings bool `yaml:"EnableSpannerSearchEmbeddings"`
```
clincoln8, line 51: nit: enable this in local.yaml for easy development?

clincoln8 (approved):
Thanks Xiao!

Can you add a section to the PR description about how you tested this before merging? Did you write embeddings to a test instance and run mixer locally to test it? Do you have example queries and responses that show that this is working to query spanner?

Overall, I think this is a good start for prototyping and iteration. As we continue to iterate, I imagine the config setup will look different.

shixiao-coder:
> Thanks Xiao!
> 
> Can you add a section to the PR description about how you tested this before merging? Did you write embeddings to a test instance and run mixer locally to test it? Do you have example queries and responses that show that this is working to query spanner?
> 
> Overall, I think this is a good start for prototyping and iteration. As we continue to iterate, I imagine the config setup will look different.



> Thanks Xiao!
> 
> Can you add a section to the PR description about how you tested this before merging? Did you write embeddings to a test instance and run mixer locally to test it? Do you have example queries and responses that show that this is working to query spanner?
> 
> Overall, I think this is a good start for prototyping and iteration. As we continue to iterate, I imagine the config setup will look different.

Yep, I added the test method and result in the description.

internal/server/spanner/datasource.go
shixiao-coder: yep, updated the name

internal/server/spanner/datasource.go
shixiao-coder: good pointed, returning it early now

internal/featureflags/featureflags.go
shixiao-coder: updated the comments

internal/server/spanner/datasource.go
n-h-diaz: sure I think we can add all flags especially if there will be more experimentation happening

n-h-diaz (approved):
Thank you! If possible, would still like to try using the parent types field to avoid duplicating in child table, but don't want to block on that

PR #1870, opened 2026-04-28
Implement filtering to exclude specific SVG nodes from BulkVariableGroupInfo response

nick-nlb:
## Issue

[b/507054271](https://b.corp.google.com/issues/507054271)

## Description

Certain SVGs may be hidden from the Stat Var Hierarchy. In `v1`, this was done upstream, outside of mixer by having certain groups given a "hidden" attribute.  With the switch to `v2`, those groups are coming through to the stat var hierarchy. 

At present only the dc/g/Hidden group is hidden. 

Mixer has been updated to now explicitly exclude this group (via the creation of an exclusion list constant that could be updated over time). It is applied to both the spanner and the `v1` wrapper paths of the `v2` endpoint for consistency, although in practice, the `v1` path will have already excluded this by virtue of getting its data from the BT cache.

## Notes

As Go does not have the equivalent of a set, I used a map for the exclusion list (to keep lookups O(1)).  I put the list in a new constants file, but if there is a more canonical place to put this type of constant in the application, please let me know.

Although the auto-reviewer's comment about the race condition was not correct, I did like an aspect of the suggested code for doing a more targeted cloning of only the elements of the response that needed cloning rather than cloning the whole response.

## Testing

[Stat Var Explorer](http://localhost:8080/tools/statvar)

* When running local mixer with Spanner on master, the "Hidden" SVG will appear.
* When running local mixer without Spanner, the "Hidden" SVG will not appear.
* When running local mixer with Spanner on this branch, the "Hidden" SVG will not appear [change].
* When running local mixer without Spanner on this branch, the "Hidden" SVG will not appear [no change].


internal/server/handler_v2.go
```diff
@@ -546,6 +555,33 @@ func (s *Server) V2BulkVariableGroupInfo(
...
+					filteredChildren = append(filteredChildren, child)
+				}
+			}
+			if item.Info.DescendentStatVarCount < 0 {
+				item.Info.DescendentStatVarCount = 0
+			}
+			item.Info.ChildStatVarGroups = filteredChildren
+		}
+		filteredData = append(filteredData, item)
+	}
```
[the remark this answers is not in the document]
nick-nlb: The auto-reviewer made two separate statements here.

The first one was about the potential for race conditions; however, I do not believe race-conditions of that nature would be an issue because we were cloning the whole response first before mutation. However, in its suggestion, it did suggest a copy-on-write approach that is more efficient for large lists. I have implemented that aspect of the suggestion.

To the second statement: in principle, this is true; however, our priority is to get the spanner path at parity with the old V1 outputs. Ideally, we would not have hard-coded exclusions, and this can be something to consider post-launch.

internal/server/handler_v2.go
```diff
@@ -516,7 +516,13 @@ func (s *Server) V2BulkVariableGroupInfo(
 	ctx context.Context, in *pbv1.BulkVariableGroupInfoRequest,
 ) (*pbv1.BulkVariableGroupInfoResponse, error) {
 	if s.shouldDivertV2(ctx) {
-		return s.dispatcher.BulkVariableGroupInfo(ctx, in)
+		resp, err := s.dispatcher.BulkVariableGroupInfo(ctx, in)
+		if err != nil {
+			return nil, err
+		}
+		return &pbv1.BulkVariableGroupInfoResponse{
+			Data: filterVariableGroupInfo(resp.GetData()),
```
n-h-diaz: could we make the filtering specific to the spanner implementation? (since each implementation could in theory handle the filtering differently, e.g. BT doesn't store results for the hidden group) (also so other calls to this function also do the filtering, including v3 https://github.com/datacommonsorg/mixer/blob/f3dd047da222cda81c03319109be3126d241e0ba/internal/server/handler_v3.go#L83)

in the new flow, the dispatcher calls the BulkVariableGroupInfo implementation for each of its sources (https://github.com/datacommonsorg/mixer/blob/f3dd047da222cda81c03319109be3126d241e0ba/internal/server/dispatcher/dispatcher.go#L233) which each independently executes the api

the spanner implementation is in https://github.com/datacommonsorg/mixer/blob/f3dd047da222cda81c03319109be3126d241e0ba/internal/server/spanner/datasource.go#L591

internal/server/handler_v2.go
```diff
@@ -530,6 +536,9 @@ func (s *Server) V2BulkVariableGroupInfo(
 
 	convertV1ToV2BulkVariableGroupInfo(v2Resp)
 
+	// Filter out excluded SVGs.
+	v2Resp.Data = filterVariableGroupInfo(v2Resp.GetData())
```
n-h-diaz: does the legacy implementation also need to be filtered?

internal/server/handler_v2.go
```diff
@@ -546,6 +555,52 @@ func (s *Server) V2BulkVariableGroupInfo(
...
 }
 
+// filterVariableGroupInfo filters out excluded SVGs from the response data.
+// We use a copy-on-write approach to perform a more efficient mutation without cloning the whole response
+func filterVariableGroupInfo(data []*pbv1.VariableGroupInfoResponse) []*pbv1.VariableGroupInfoResponse {
+	// Pre-allocate to avoid underlying array resizing
+	filteredData := make([]*pbv1.VariableGroupInfoResponse, 0, len(data))
+
+	for _, item := range data {
+		if _, ok := svgGroupInfoExclusionList[item.GetNode()]; ok {
```
n-h-diaz: rather than removing this entry after fetching it, could we filter nodes at request time? (to avoid unnecessary computation if the whole entry will  be dropped)

internal/server/v2/bulk_variable_group_info/bulk_variable_group_info_hidden_excluded.json
```diff
@@ -0,0 +1 @@
+{}
```
n-h-diaz: nit: for empty responses, we currently still return the node with empty info, for example

https://api.datacommons.org/v1/bulk/info/variable-group?key={}&nodes=dc/g/Hidden

returns {"data":[{"node":"dc/g/Hidden","info":{}}]}

n-h-diaz (commented):
Thanks! I had a few comments on when we might do the filtering (happy to discuss more in person tomorrow if helpful), but generally looks good

thank you for volunteering to take on mixer bugs! :)

internal/server/handler_v2.go
nick-nlb: I mulled this - I ended up at first wrapping the legacy endpoint with the idea that if we ever did add an excluded SVG here later, it would be excluded by both legacy and spanner regardless of whether either excluded them internally.

However with the migration of this functionality to the spanner implementation (which cleans things up substantially) this becomes moot anyway.

internal/server/handler_v2.go
nick-nlb: I've moved the filtering now to the spanner implementation in datasource.go

internal/server/spanner/datasource.go
```diff
@@ -38,8 +38,13 @@ import (
 	"github.com/golang/geo/s2"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/proto"
 )
 
+var svgGroupInfoExclusionList = map[string]struct{}{
```
n-h-diaz, line 46: nit: could you add a comment on what this is used for

internal/server/v2/bulk_variable_group_info/bulk_variable_group_info_test.go
```diff
@@ -47,6 +47,12 @@ func TestV2BulkVariableGroupInfo(t *testing.T) {
 				1,
 				"bulk_variable_group_info.json",
 			},
+			{
```
n-h-diaz, line 50: nit: could you also add the test to the v3 version? https://github.com/datacommonsorg/mixer/blob/682e5168f45755e43297556abab86eff93914674/internal/server/v3/bulk_variable_group_info/bulk_variable_group_info_test.go (v2 is not yet enabled for spanner by default yet, but I think still useful to have the test here too for when we switch)

n-h-diaz (approved):
Thank you!

PR #1875, opened 2026-04-30
Return stat var definition from spanner in bulk/info/variable-group

juliawu:
This PR updates Mixer's Spanner path to fetch and return the `definition` property for child statistical variables (SVs) within the bulk/info/variable-group endpoint.

In the Spanner graph database, a statistical variable's definition is modeled as an out-arc edge with the predicate `definition` directed toward a literal node. This PR retrieves that literal string value using a scalar subquery, populates it onto the backend data models, and maps it to the resulting protocol buffer response.

Target Endpoints
* V3BulkVariableGroupInfo (Spanner)
* V2BulkVariableGroupInfo (When Spanner is enabled via shouldDivertV2)

## Changes Made
1. Backend Go Models
Updated the following structs in `internal/server/spanner/model.go` to include a Definition field mapped to the Spanner column `definition`:

  * StatVarGroupNode
  * SVGChild
  * ChildSV

2. Spanner SQL Query Statements
Added a scalar subquery in `internal/server/spanner/statements.go` to fetch the definition of a child statistical variable from the Edge and Node tables:

  * getStatVarGroupNode: Added '' AS definition to the child StatVarGroup (SVG) query branch to match the UNION ALL signature, and added the scalar subquery to the child SV query branch.
  * getSVGChildren: Added the scalar subquery to select the definition for baseline SVG child listings.
  * getFilteredChildSVs: Added the scalar subquery to select the definition for child SVs filtered by entity existence and source constraints.

3. Response Conversion & Mapping
Updated `internal/server/spanner/dsutil.go` to map the Definition field to the protobuf output (pb.StatVarGroupNode_ChildSV) inside:

  * svgInfoToBulkVariableGroupInfoResponse()
  * filteredSVGInfoToBulkVariableGroupInfoResponse()

## Testing Strategy

1. Automated Tests
Updated all relevant Spanner query goldens to include the new definition SQL column using ./scripts/update_goldens.sh.

Ran the tests locally to verify the correct generation of SQL strings and backend parsing:

2. Manual Testing
Launched a local Mixer server with the Spanner graph enabled:

```bash
./run_server.sh \
    --feature_flags_path=$PWD/deploy/featureflags/local.yaml \
    --spanner_graph_info="$(cat deploy/storage/spanner_graph_info.yaml)"
```
Verified the response manually using curl:

```bash
curl -s "http://localhost:8081/v2/bulk/info/variable-group?nodes=dc/g/Demographics"
```

The response correctly included definition attributes inside the childStatVars objects.

PR #1877, opened 2026-05-01
Return definitions optionally in bulk/info/variable-group

juliawu:
## Problem

We need to add definitions to the output for `bulk/info/variable-group`.  However, as seen in #1875, in Spanner fetching definitions for statistical variables requires expensive correlated subqueries with joins on the `Node` and `Edge` tables. This was causing timeouts in testing and would likely cause issues in production for large variable groups.

## Solution

We made the fetching of definitions optional by introducing an `include_definitions` boolean flag in the `BulkVariableGroupInfoRequest` message.
*   By default (when `include_definitions` is `false`), definitions are **not** fetched, avoiding the expensive join and returning an empty string for the definition field.
*   Clients that explicitly need definitions can set `include_definitions = true`.

To keep the SQL maintainable and safe for the query planner, I implemented separate query templates in `statements.go` for both cases (with and without definitions) rather than dynamically manipulating the query strings.

### HTTP (via Envoy)
To fetch definitions via the HTTP API, add `includeDefinitions=true` (or `include_definitions=true`) to the query parameters:
```http
GET /v3/bulk/info/variable-group?nodes=dc/g/Demographics&includeDefinitions=true
```

### Proto

*   **[info.proto](file:///Users/juliawu/Repos/datacommons-mixer/proto/v1/info.proto)**: Added `bool include_definitions = 5;` to `BulkVariableGroupInfoRequest`.
### Server (Spanner)
*   **[statements.go](file:///Users/juliawu/Repos/datacommons-mixer/internal/server/spanner/statements.go)**: Added separate query templates for fetching with and without definitions.
*   **[query_builder.go](file:///Users/juliawu/Repos/datacommons-mixer/internal/server/spanner/query_builder.go)**: Updated query builder functions to accept `includeDefinitions` and select the appropriate template.
*   **[client.go](file:///Users/juliawu/Repos/datacommons-mixer/internal/server/spanner/client.go)** & **[query.go](file:///Users/juliawu/Repos/datacommons-mixer/internal/server/spanner/query.go)**: Updated interface and implementations to pass the flag down.
*   **[datasource.go](file:///Users/juliawu/Repos/datacommons-mixer/internal/server/spanner/datasource.go)**: Extracted the flag from the request and passed it to the client methods.

### Tests
*   **[query_cases_test.go](file:///Users/juliawu/Repos/datacommons-mixer/internal/server/spanner/golden/query_cases_test.go)**: Added new test cases with `include_definitions: true` to verify both paths.
*   Updated corresponding golden files for SQL outputs

### Automated Tests

*   Verified locally that all tests pass when `ENABLE_SPANNER_GRAPH=false` (matching CI/CD defaults).
*   Verified locally that the specific tests for these queries pass when `ENABLE_SPANNER_GRAPH=true`.
    *   `go test -v ./internal/server/spanner/golden/ -run "TestGetStatVarGroupNodeQuery|TestGetFilteredSVGChildren|TestGetSVGChildrenQuery"`


internal/server/spanner/golden/query_builder/get_filtered_sv_with_definitions.sql
```diff
@@ -0,0 +1,32 @@
...
+				WHERE e_def.subject_id = n.subject_id
+				AND e_def.predicate = 'definition'
+				LIMIT 1
+			), '') AS definition
+		FROM Node n
+		JOIN (
+			SELECT
+				e.subject_id AS subject_id
+			FROM Edge e
+			JOIN@{JOIN_TYPE=HASH_JOIN} (
```
juliawu, line 17: The Codacy issue is flagging this line. However, per Gemini "he syntax @{JOIN_TYPE=HASH_JOIN} is a Google Cloud Spanner query hint. It tells Spanner's query planner to use a hash join algorithm for that specific join, which is crucial for performance in this query.

Because the linter (SQLint) is checking for strictly standard ANSI SQL, it doesn't recognize Spanner's vendor-specific extensions and flags it as a syntax error."

Because the hint is useful for spanner, I am leaving it in.

proto/v1/info.proto
```diff
@@ -82,6 +82,8 @@ message BulkVariableGroupInfoRequest {
   // (optional) Number of passed in entities that should have data for the stat
   // var (group)s that are returned. Default to 1 if unspecified.
   int32 num_entities_existence = 4;
+  // (optional) Whether to include definitions in the response.
```
n-h-diaz: nit: maybe clarify this refers to sv definition?

n-h-diaz (approved):
Thank you! 

Can be a follow up pr if easier, but could you also add/update the query_tests (test retrieval of data from spanner) and/or the e2e test for this endpoint?

proto/v1/info.proto
juliawu: Good idea. Done!

juliawu:
> Thank you!
> 
> Can be a follow up pr if easier, but could you also add/update the query_tests (test retrieval of data from spanner) and/or the e2e test for this endpoint?

Will do in a followup PR!

PR #1880, opened 2026-05-05
Update the disaster event query by splitting a single match to multiple matches

shixiao-coder:
This change update the node edge match from multiple conditions in a single match

to a match on disaster event first that reduces the number of node to match in later stage and then do matches on other conditions

This reduces py client time from 20s to 2s


n-h-diaz (approved):
wow this is an amazing improvement!!

PR #1884, opened 2026-05-08
Add a new statement to filter the nodes/DCIDs by node types

shixiao-coder:
This PR add statement, query_builder, query that invokes the statement and datasource.go to use the function to filter SVG as well as topics from this logic

### DIFF
--- a/internal/server/spanner/datasource.go
+++ b/internal/server/spanner/datasource.go
@@ -711,33 +711,26 @@ func (sds *SpannerDataSource) BulkVariableGroupInfo(ctx context.Context, req *pb
 	if req.NumEntitiesExistence < 0 {
 		return nil, status.Errorf(codes.InvalidArgument, "numEntitiesExistence must be non-negative")
 	}
-	var dbSvgs, topics []string
-	g, errCtx := errgroup.WithContext(ctx)
-	g.SetLimit(2)
-	g.Go(func() error {
-		var err error
-		dbSvgs, err = sds.client.FilterNodesByType(errCtx, req.GetNodes(), "StatVarGroup")
-		return err
-	})
-	g.Go(func() error {
-		var err error
-		topics, err = sds.client.FilterNodesByType(errCtx, req.GetNodes(), "Topic")
-		return err
-	})
-	if err := g.Wait(); err != nil {
+	nodeToTypes, err := sds.client.FilterNodesByTypes(ctx, req.GetNodes(), []string{"StatVarGroup", "Topic"})
+	if err != nil {
 		return nil, status.Errorf(codes.Internal, "error filtering nodes: %v", err)
 	}
 
 	foundNodes := map[string]bool{}
 	var svgs []string
-	for _, node := range dbSvgs {
+	var topics []string
+
+	for node, matchedTypes := range nodeToTypes {
 		foundNodes[node] = true
-		if _, ok := svgGroupInfoExclusionList[node]; !ok {
-			svgs = append(svgs, node)
+		for _, t := range matchedTypes {
+			if t == "StatVarGroup" {
+				if _, ok := svgGroupInfoExclusionList[node]; !ok {
+					svgs = append(svgs, node)
+				}
+			} else if t == "Topic" {
+				topics = append(topics, node)
+			}
 		}
-	}
-	for _, node := range topics {
-		foundNodes[node] = true
 	}
 
 	invalidNodeMap := map[string]bool{}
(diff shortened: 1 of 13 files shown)


internal/server/spanner/datasource.go
```diff
@@ -711,17 +711,29 @@ func (sds *SpannerDataSource) BulkVariableGroupInfo(ctx context.Context, req *pb
...
-		if strings.HasPrefix(node, prefixTopic) {
-			topics = append(topics, node)
-		} else if strings.HasPrefix(node, prefixSVG) {
-			// Exclude hidden nodes from the database query
-			if _, ok := svgGroupInfoExclusionList[node]; !ok {
-				svgs = append(svgs, node)
-			}
-		} else {
+		if !foundNodes[node] {
 			return nil, status.Errorf(codes.InvalidArgument, "node %s is not a valid StatVarGroup or Topic node", node)
```
nick-nlb: This relates to the discussion we have about what functionality we want when we request non-existent data (as an overview of what we discussed just now):

The decision in terms of balancing behavior and parity was to match the existing (pre-PR) functionality:
* For any nodes given, we only return data for SVG/topics
* If a node either did not exist, or was not an SVG/topic, then we return no information for that node (`{}`)
* If multiple nodes are given, and some are valid/exist and some don't, we return information for the valid ones, and no information for the ones that aren't valid.

internal/server/spanner/query_builder.go
nick-nlb: This is now unused (it's what is causing the lint error during the test runs).

internal/server/spanner/datasource.go
```diff
@@ -711,17 +711,29 @@ func (sds *SpannerDataSource) BulkVariableGroupInfo(ctx context.Context, req *pb
...
 		}
 	}
```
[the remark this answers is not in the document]
nick-nlb: "Correctness: FilterNodesByType returns a DISTINCT list of IDs from the database. By using these results to build the svgs and topics slices directly, any duplicate nodes in the original req.GetNodes() will be lost in the final response. Additionally, the original order of nodes is not guaranteed to be preserved.
"

A note about this: deduping is the behavior in `v1`, so for parity, this is desired anyway. The order being returned is a consideration.

The efficiency comment the autoreviewer made is worth investigation.

internal/server/spanner/statements.go
```diff
@@ -644,6 +646,11 @@ var statements = struct {
 				HAVING COUNT(DISTINCT %s) >= @numEntitiesExistence`,
 	getEmbeddingFromQuery: `		SELECT embeddings.values
 		FROM ML.PREDICT(MODEL %s, (SELECT @search_label AS content, @task_type AS task_type))`,
+	filterNodesByType: `		SELECT DISTINCT subject_id
+		FROM Node
+		WHERE ARRAY_LENGTH(types) > 0
+			AND @type_filter IN UNNEST(types)
+			AND subject_id IN UNNEST(@nodes)`,
```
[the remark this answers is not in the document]
nick-nlb: This comment seems to be correct (and I tried the updated code here, which produces the expected results).

nick-nlb (commented):
I've summarized in a comment what we discussed about parity/behavior (and a couple of other comments).

Could we also do comparative latency tests between the pre-PR and post-PR versions of this endpoint? 

From a usage perspective on the website, it seems noticeably slower, but that may just be in the moment. We should get a side-by-side sense of how much latency this adds (as these calls play a role in NL search as well as the stat var explorer).

internal/server/spanner/datasource.go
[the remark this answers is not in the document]
shixiao-coder: updated for point 2, point 1 is desired behavior

internal/server/spanner/datasource.go
shixiao-coder: as discussed offline, now returning empty results for these nodes instead of throwing error

internal/server/spanner/query_builder.go
shixiao-coder: removed

nick-nlb:
> From a usage perspective on the website, it seems noticeably slower, but that may just be in the moment. We should get a side-by-side sense of how much latency this adds (as these calls play a role in NL search as well as the stat var explorer).

This was definitely just "in the moment" (i.e., it reflected a heavy general load on the spanner DB from other sources rather than a difference between the two versions of this code), and can be ignored.

nick-nlb (approved):
Thank you for all of this Xiao! 

I've make a couple of follow-up updates (see commit messages), and tested everything and it all looks great.

PR #1885, opened 2026-05-11
[DCP - Embeddings] Keep the Indicator resolve through GCS embeddings for now

gmechali:
Ensures the resolver keeps Indicators on V1 such that it goes through the GCS embeddings. This is only required until Spanner embeddings, expected in about a month.

### DIFF
--- a/internal/server/handler_v2.go
+++ b/internal/server/handler_v2.go
@@ -46,7 +46,7 @@ func (s *Server) V2Resolve(
 ) (*pbv2.ResolveResponse, error) {
 	// TODO: Remove this once embeddings search (resolver == "indicator") are
 	// supported thorugh Spanner.
-	if s.shouldDivertV2(ctx) && in != nil && in.GetResolver() != "indicator" {
+	if s.shouldDivertV2(ctx) && (in == nil || in.GetResolver() != "indicator") {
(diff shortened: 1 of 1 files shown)


internal/server/handler_v2.go
```diff
@@ -44,7 +44,10 @@ import (
 func (s *Server) V2Resolve(
 	ctx context.Context, in *pbv2.ResolveRequest,
 ) (*pbv2.ResolveResponse, error) {
-	if s.shouldDivertV2(ctx) {
+	// If request is for embeddings search (resolver == "indicator"),
+	// do not divert to Spanner. Note, we will need to remove this once
+	// Spanner embeddings are supported.
```
juliawu: For completeness, can you add a TODO: to remove this once Spanner embeddings are supported?

gmechali: done.

internal/server/handler_v2.go
```diff
@@ -44,7 +44,10 @@ import (
...
+	if s.shouldDivertV2(ctx) && in.GetResolver() != "indicator" {
```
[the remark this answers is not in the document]
gmechali: fixed.

juliawu (approved):
Thank you for the fix!

PR #1889, opened 2026-05-12
Federate /v2/observation entity expression expansions

clincoln8:
## Goal
Implement expression expansion for the `Observation` API in the `DataSources` layer to support federated queries across local Spanner and Remote Mixer.

## Requirements based on Legacy Flow
*   **Original Expression to Remote**: The Remote Mixer must receive the original request with the expression intact so it can perform its own expansion against its complete graph.
*   **Unified Resolution**: We must resolve the expression against all available sources (local + remote) to get a complete list of child entities.
*   **Targeted Local Queries**: Local sources (like Spanner) must be queried with the resolved list of DCIDs, and the expression must be cleared to prevent them from attempting expansion they cannot support or failing due to constraints (Spanner driver forbids both `Dcids` and `Expression` being set).

## Implementation Details
We implemented this logic in `internal/server/datasources/datasources.go` inside the `Observation` method.

### Why this location?
The `DataSources` struct is the federation hub of Mixer. It knows about all configured sources and handles dispatching and merging. By intercepting the request here, we can:
1.  Perform the expansion once using the existing `Node` API (which already federates).
2.  Shape the requests appropriately for each source before they are dispatched in parallel.
    * Remote Mixer gets the original entity expression
    * Other datasources get the expanded dcid list.

### Architectural Flow
1.  **Intercept**: If the request contains an entity expression and we are in a federated environment (both local and remote sources present), we trigger the expansion flow.
2.  **Resolve**: We call `ds.Node` (our own method) with the expression to get all child places. We added a helper `fetchAllDCIDs` to handle pagination and ensure we get all results.
4.  **Extract**: We look for the arc key `"containedInPlace+"` in the response (handled normalization differences between Remote and Spanner).
5.  **Clone & Dispatch**: We clone the request for local sources, populating `Dcids` with the resolved list and clearing `Expression`. The remote source receives the unmodified original request.
6.  **Merge**: Results are merged using `merger.MergeMultiObservation`.

### Unit Tests
We added tests in `datasources_test.go`:
*   `TestObservation_ExpressionExpansion`: Verifies that Spanner receives resolved DCIDs and Remote receives the original expression.
*   `TestFetchAllDCIDs`: Verifies that the pagination loop correctly follows `NextToken` to fetch all pages of results.

### Manual Verification
To verify this change manually or in the future, I:

#### 1. Set up Spanner Database
I set up a Spanner database with the Data Commons schema and some dummy data. Detailed instructions are in [Verification.md](file:///Users/calinc/.gemini/jetski/brain/4456a06e-03b6-4343-89c0-dd97ea498f9d/Verification.md) to:
*   Upload proto descriptors.
*   Create `Node`, `Edge`, and `Observation` tables.
*   Insert a dummy county linked to California via `linkedContainedInPlace` and `typeOf` edges.

#### 2. Started Envoy
I ran Envoy to handle HTTP to gRPC transcoding:
```bash
```

#### 3. Ran Local Mixer
I started the local Mixer pointing to my Spanner instance and enabling remote federation:
```bash
export SPANNER_CONF="project: <project>\ninstance: <instance>\ndatabase: <database>"
export MIXER_API_KEY="<your_api_key>"

./run_server.sh \
  --use_bigquery=false \
  --use_base_bigtable=false \
  --use_branch_bigtable=false \
  --use_spanner_graph=true \
  --spanner_graph_info="$SPANNER_CONF" \
  --remote_mixer_domain=https://api.datacommons.org
```

#### 4. Verified with Curl
I ran a query with an expression and verified that results contained both remote data (California counties) and my local dummy data:
```bash
curl -X POST -H "Content-Type: application/json" -d '{
  "select": ["variable", "entity", "date", "value"],
  "variable": {"dcids": ["Count_Person"]},
  "entity": {"expression": "geoId/06<-containedInPlace+{typeOf:County}"},
  "date": "2020"
}' http://localhost:8081/v2/observation
```

### DIFF
--- a/internal/server/datasources/datasources.go
+++ b/internal/server/datasources/datasources.go
@@ -16,14 +16,17 @@ package datasources
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/datacommonsorg/mixer/internal/merger"
 	pb "github.com/datacommonsorg/mixer/internal/proto"
 	pbv1 "github.com/datacommonsorg/mixer/internal/proto/v1"
 	pbv2 "github.com/datacommonsorg/mixer/internal/proto/v2"
 	"github.com/datacommonsorg/mixer/internal/server/datasource"
+	v2 "github.com/datacommonsorg/mixer/internal/server/v2"
 	"github.com/datacommonsorg/mixer/internal/server/v2/resolve"
 	"github.com/datacommonsorg/mixer/internal/translator/sparql"
+	"google.golang.org/protobuf/proto"
 
 	"golang.org/x/sync/errgroup"
 )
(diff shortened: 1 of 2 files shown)


clincoln8:
We are going to split this logic into a preprocessor and the spanner datasource implementation. Closing for now and will open a new PR with this new approach.

PR #1892, opened 2026-05-12
[v0/v1 migration] Implement CheckVariableGroupExistence to support StatVarGroup existence checks in Spanner

nick-nlb:
## Issue

[b/505048261](b/505048261)

## Description

Legacy observation existence checks were able to resolve checks for SVGs in addition to stat vars. This functionality was used to power the "Data source" and "Dataset" filters in the Stat Var Explorer. These checks used the BT cache, which is no longer available in the Spanner path and resulted in this functionality being broken.

This PR brings this functionality to Spanner by bringing existence check support to SVGs, specifically for imports / datasets.

## Notes

This PR implements the SVG/topic/SV existence checks against imports and datasets by using the `Cache` table. True entity (place) existence checks do not yet exist in the Spanner flow.

## Testing

In the stat var explorer, when on master but with Spanner on, the "Data source" filter is empty.

When on this branch, the "Data source" filter is populated and should match production. Selecting "Data sources" should then both filter the SVGs and populate the "Dataset" list, with outputs that match production.


internal/server/spanner/query.go
```diff
@@ -173,6 +173,139 @@ func (sc *spannerDatabaseClient) CheckVariableExistence(ctx context.Context, var
...
+			continue
+		}
+		subjectID := row[0]
+		entity := row[1]
+
+		importName := subjectID
+		// Extract import name by dropping the 8-char prefix, mirroring SUBSTR(..., 9) in SQL queries.
+		if len(subjectID) > 8 {
+			importName = subjectID[8:]
+		}
```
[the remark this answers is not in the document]
nick-nlb: This is discussed in the PR description.

internal/server/spanner/query.go
```diff
@@ -173,6 +173,139 @@ func (sc *spannerDatabaseClient) CheckVariableExistence(ctx context.Context, var
...
+			roots = append(roots, v)
+		} else {
+			nonRoots = append(nonRoots, v)
+		}
+	}
+
+	var result [][]string
+
+	// Step 2: Get import_names for the entities
+	predicate := getImportFilterPredicate(entities[0])
```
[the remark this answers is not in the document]
nick-nlb: Agreed with this - will fix.

internal/server/spanner/datasource.go
```diff
@@ -172,15 +172,62 @@ func (sds *SpannerDataSource) Observation(ctx context.Context, req *pbv2.Observa
 	isExistenceRequest := !qo.date && !qo.value && !qo.facet && len(entities) > 0 && entityExpr == ""
 
 	if isExistenceRequest {
-		rows, err := sds.client.CheckVariableExistence(ctx, variables, entities)
+		if len(variables) == 0 {
+			rows, err := sds.client.CheckVariableExistence(ctx, variables, entities)
```
keyurva: Can we create helpers like `handleEntityOnlyExistence`, `sortNodesByType`, etc. for better readability?

internal/server/spanner/datasource.go
keyurva: For my understanding: This is interesting. So the current implementation returns SVs, SVGs and topics under VariableMeasured?

internal/server/spanner/query.go
```diff
@@ -173,6 +173,154 @@ func (sc *spannerDatabaseClient) CheckVariableExistence(ctx context.Context, var
 	return queryDynamic(ctx, sc, *stmt)
 }
 
+// CheckVariableGroupExistence checks for the existence of observations for the given variable groups/topics and entities.
+// Returns a slice of rows, where each row contains [variable, entity] that has at least one observation.
+func (sc *spannerDatabaseClient) CheckVariableGroupExistence(ctx context.Context, variables []string, entities []string) ([][]string, error) {
```
keyurva: `variables` here are actually groups or topics, correct? If so, should we name them `groupsOrTopics` or something similar?

internal/server/spanner/query.go
```diff
@@ -173,6 +173,154 @@ func (sc *spannerDatabaseClient) CheckVariableExistence(ctx context.Context, var
...
+	if len(variables) == 0 || len(entities) == 0 {
+		return [][]string{}, nil
+	}
+
+	// Step 1: Identify which variables are root nodes (have no parents).
+	// We are doing this because a query on root involve a massive join involving
+	// a large number of variables (causing a slow query and risk of timeout).
+	stmtRoots := spanner.Statement{
+		SQL: `SELECT DISTINCT subject_id
```
keyurva: Our pattern is to define statements in `statements.go`. Can we do that for these statements as well?

internal/server/spanner/query.go
```diff
@@ -173,6 +173,154 @@ func (sc *spannerDatabaseClient) CheckVariableExistence(ctx context.Context, var
...
+			FROM Edge
+			WHERE predicate = 'specializationOf'
```
keyurva: Can you describe how topics are handled because they won't have any `specializationOf` predicates.

internal/server/spanner/query.go
```diff
@@ -173,6 +173,154 @@ func (sc *spannerDatabaseClient) CheckVariableExistence(ctx context.Context, var
...
+		if len(row) != 2 {
+			continue
+		}
+		subjectID := row[0]
+		entity := row[1]
+
+		importName := subjectID
+		// Extract import name by dropping the 8-char prefix, mirroring SUBSTR(..., 9) in SQL queries.
+		if len(subjectID) > 8 {
+			importName = subjectID[8:]
```
keyurva: Chatted offline: we can use the `provenance` field for this which will allow us to use the `subjectID` as-is.

internal/server/spanner/query.go
keyurva: To manage the size of this method, can we create helpers - e.g. `identifyRoots`, `handleRoots`, `handleNonRoots`, etc.?

internal/server/spanner/query.go
```diff
@@ -173,6 +173,154 @@ func (sc *spannerDatabaseClient) CheckVariableExistence(ctx context.Context, var
...
+	for k := range entitiesByPredicate {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	var filters []string
+	params := map[string]interface{}{}
+	for _, p := range keys {
+		ents := entitiesByPredicate[p]
+		filters = append(filters, fmt.Sprintf(`(predicate = '%s' AND object_id IN UNNEST(@%s))`, p, p))
```
keyurva: I believe (not certain) these filter statements are declared in `statements.go` as well. Let's declare them there if that's the case for other filters as well.

internal/server/spanner/query.go
```diff
@@ -173,6 +173,154 @@ func (sc *spannerDatabaseClient) CheckVariableExistence(ctx context.Context, var
...
+		params[p] = ents
+	}
+
+	sourceEdgesStmt := spanner.Statement{
+		SQL: `SELECT DISTINCT subject_id, object_id
```
keyurva: Can this be declared in `statements.go` as well?

internal/server/spanner/query.go
```diff
@@ -173,6 +173,154 @@ func (sc *spannerDatabaseClient) CheckVariableExistence(ctx context.Context, var
...
+					result = append(result, []string{root, entity})
+				}
+			}
+		}
+	}
+
+	// Handle non-roots
+	if len(nonRoots) > 0 {
+		existenceQueryStmt := spanner.Statement{
+			SQL: `SELECT DISTINCT e.object_id AS variable, o.import_name
```
keyurva: Similar `statements.go` comment as others.

internal/server/spanner/query.go
keyurva: nit: From offline chat, we learned that `variables` are actually SVGs or topics and `entities` are sources. 

And the outputs are SVGs / topics and import names.

Let's rename variables / add comments wherever appropriate to make this clear.

keyurva (commented):
Thanks Nick! This was a deep change and would've taken a lot of figuring out and getting things in order. Much appreciated!

internal/server/spanner/datasource.go
nick-nlb: The Spanner implementation before this PR only ever surfaced existence checks for SVs, but with this now extended capacity to check variable groups rather than just variables (for now only for sources), those SVGs/Topics would be put in the VariableMeasured (the logic that creates the final payload by grouping SVs and now SVGs/Topics to remain unchanged). Really "variable" in the context of an existence check means "variable or group of variables"

internal/server/spanner/query.go
nick-nlb: It was (and so the params should have been named as you had suggested!) However this function has been reworked. Because we are now checking source existence in the cache, it was just as efficient to also extend this checks to both svs and variable groups (svgs and topics) - so the function is redone and now named to reflect that (but with "variable" in the name and as the parameter - that seems to be how we usually name "variable/variable group" combos).

internal/server/spanner/query.go
nick-nlb: Done! 

These particular SQL statements are no longer there, but the ones that we do have are now in `statements.go`

internal/server/spanner/query.go
nick-nlb: This is now out of date, but topics are handled in source checks by joining via `linkedMember` from the cache (SVGs are joined via `linkedMemberOf`)

internal/server/spanner/query.go
nick-nlb: The method itself is quite different, but I've applied this principle to the newer method and also pulled out the overall existence check functionality it's own separate method.

nick-nlb: While `variable` now does mean all three, I've commented to make it more clear.

internal/server/spanner/datasource.go
```diff
@@ -172,91 +177,243 @@ func (sds *SpannerDataSource) Observation(ctx context.Context, req *pbv2.Observa
...
-			return nil, fmt.Errorf("error getting observations (contained in): %v", err)
+			return nil, fmt.Errorf("error checking SV source existence: %w", err)
 		}
-		observations, err = sds.client.GetObservationsContainedInPlace(ctx, variables, containedInPlace)
+		allRows = append(allRows, rows...)
+	}
+
+	// 3. SVGs + Sources
+	if len(svgs) > 0 && len(sources) > 0 {
+		rows, err := sds.client.CheckVariableSourceExistence(ctx, svgs, sources, "linkedMemberOf")
```
keyurva, line 285: Can we add a TODO to do these existence checks in parallel?

internal/server/spanner/datasource.go
```diff
@@ -172,91 +177,243 @@ func (sds *SpannerDataSource) Observation(ctx context.Context, req *pbv2.Observa
...
-			return nil, fmt.Errorf("error getting observations: %v", err)
+			return nil, fmt.Errorf("error checking Topic source existence: %w", err)
 		}
+		allRows = append(allRows, rows...)
 	}
 
-	observations = filterObservationsByDateAndFacet(observations, date, req.Filter)
+	// 5. Groups + Places (Currently Unsupported)
+	if (len(svgs) > 0 || len(topics) > 0) && len(places) > 0 {
+		slog.Warn("Place existence checks for StatVarGroup/Topic are not supported in Spanner yet, skipping.", "svgs", svgs, "topics", topics)
```
keyurva, line 303: Curious - are they supported in BT? If so, might be worthwhile to file a bug to track this task in spanner.

internal/server/spanner/datasource.go
```diff
@@ -172,91 +177,243 @@ func (sds *SpannerDataSource) Observation(ctx context.Context, req *pbv2.Observa
...
+		} else {
+			places = append(places, e)
+		}
+	}
+	return
+}
+
+// splitVariables splits a list of variables into SVs, SVGs, and Topics.
+func (sds *SpannerDataSource) splitVariables(ctx context.Context, variables []string) (statVars []string, svgs []string, topics []string, err error) {
+	nodeToTypes, err := sds.client.FilterNodesByTypes(ctx, variables, []string{"StatVarGroup", "Topic"})
```
keyurva: I thought we discussed doing this based on dcid string match vs querying the DB (not ideal but good for latency). But leave it up to you.

keyurva (approved):
Thanks for the updates! Appreciate your digging into this and for making the code very easy to follow.

internal/server/spanner/datasource.go
nick-nlb: There is a bug for this already actually. It's the next bug on my plate after this. There is one (somewhat obscure) part of the NL search that does a place check by topic, for example.

internal/server/spanner/datasource.go
nick-nlb: I first started using the DCID prefix-sniffing  method for this.

I then noticed that variables in one of our UN test Database didn't use our own prefixes, i.e. ,`c/g/Root` instead of `dc/g/Root` - leading to concerns that custom data might be less reliable.

However, now that I think about it - we could use the `g` part alone, (i.e., as long as we see a pattern of `*/g/SVG` we ,know it is an `SVG`. If it is `*/topic/Topic`

I saw some new code in fetcher that indicates that this is a valid way to do determine topics (and so I assume SVGs):

```
func isTopicDcid(dcid string) bool {
	return strings.Contains(dcid, "/topic/")
}
```
I'll make this update this morning. It's the last thing and then I'll merge it in.

internal/server/spanner/datasource.go
nick-nlb: For reference, the bug is b/514069768.

internal/server/spanner/datasource.go
nick-nlb: I've made this update now, using a variation on the `isTopicDcid` function (now in utils), with a TODO to consolidate with this function from the fetch path.

PR #1895, opened 2026-05-14
/v2/observation: Fetch all pages for ContainedInPlace relation expression expansion

clincoln8:
## High Level Summary
Follow up to https://github.com/datacommonsorg/mixer/pull/1894

Implemented exhaustive page fetching (traversing all pages) for `/v2/node` calls during relation expression expansion. This ensures that we get the complete set of entities across both remote and local sources, fixing the issue where only the first page of results was processed.

We chose to implement this via a shared helper function `NodeFetchAll` in the `datasource` package. We selected this approach after evaluating several design alternatives (including making DataSources "page-aware" or updating the `DataSource` interface).

**Pros of the Helper Approach:**
*   **Zero Interface Changes**: It avoids the need to modify the `DataSource` interface, which would have required updating all implementations (Spanner, Bigtable, SQL, Mocks) and would have had a high blast radius.
*   **No Code Duplication**: It centralizes the complex logic for handling pagination tokens and deep-merging `NodeResponse` objects, preventing duplication in `RelationExpressionProcessor` and `SpannerDataSource`.
*   **Safety**: It keeps the behavior strictly internal without leaking implementation details (like auto-pagination flags) to the public API via `NodeRequest` protos.

#### `internal/server/datasource/datasource.go`
*   Added `NodeFetchAll` helper function. It repeatedly calls `ds.Node` as long as a `NextToken` is returned, and merges the responses into a single response.
*   Requires `pageSize` to be positive, forcing callers to be explicit.

#### `internal/server/dispatcher/relation_expression_processor.go`
*   Updated `fetchEntities` to use `NodeFetchAll` when calling the remote source, passing `datasources.DefaultPageSize` (500).

#### `internal/server/spanner/datasource.go`
*   Updated `fetchLocalChildPlaces` to use `NodeFetchAll` when fetching local child places, passing `1000`.

## Testing
*   Added unit tests for `NodeFetchAll` in `internal/server/datasource/datasource_test.go` testing:
    *   **Single Page**: Verified that the helper returns response directly without unnecessary looping when `NextToken` is empty.
    *   **Multi-Page**: Verified that the helper correctly loops and deep-merges nested maps in `NodeResponse.Data` across multiple pages.
    *   **Errors**: Verified that errors on both initial and subsequent calls are properly propagated.
*   Ran the full test suite, and all tests passed successfully.
*   **Manual E2E Verification**:
    *   Simulated responses exceeding page size limits to verify pagination traversal.
    *   **Remote Pagination**: Queried for cities in California (returning ~1900 results from remote mixer) to trigger its 500 limit.
    *   **Local Pagination**: Added 1005 dummy cities to local Spanner to trigger the 1000 limit in `fetchLocalChildPlaces`.
    *   Verified that results from both sources are fully fetched and merged correctly by checking the count of returned entities in the response. Tested with both `BaseDCMixer` (local-only) and `LocalMixer` (federated).

## Discussion on Default Page Size
Consider introducing datasource-specific default page sizes in the future.
Currently, we require the caller to specify the page size. However, different backends might perform better with different page sizes. For example, a local Spanner call might handle 1000 nodes efficiently, reducing round-trips, while a remote HTTP call might want to stay at 500 or less to keep payloads manageable.

We could implement this by defining an unexported map and a helper function in the `datasource` package:

```go
var defaultPageSizes = map[DataSourceType]int{
	TypeSpanner: 1000,
	TypeRemote:  0,
	TypeSQL:     500,
	TypeMock:    100,
}

func GetDefaultPageSize(ds DataSource) int {
	if size, ok := defaultPageSizes[ds.Type()]; ok {
		return size
	}
	return datasources.DefaultPageSize // Global fallback
}
```
Hiding the map behind a helper would avoid leaking implementation details to callers.

### DIFF
--- a/internal/server/datasource/datasource.go
+++ b/internal/server/datasource/datasource.go
@@ -16,10 +16,13 @@ package datasource
 
 import (
 	"context"
+	"fmt"
+	"log/slog"
 
 	pb "github.com/datacommonsorg/mixer/internal/proto"
 	pbv1 "github.com/datacommonsorg/mixer/internal/proto/v1"
 	pbv2 "github.com/datacommonsorg/mixer/internal/proto/v2"
+	"google.golang.org/protobuf/proto"
 )
 
 // DataSourceType represents the type of data source.
@@ -45,4 +48,61 @@ type DataSource interface {
 	BulkVariableInfo(context.Context, *pbv1.BulkVariableInfoRequest) (*pbv1.BulkVariableInfoResponse, error)
 	BulkVariableGroupInfo(context.Context, *pbv1.BulkVariableGroupInfoRequest) (*pbv1.BulkVariableGroupInfoResponse, error)
 	SdmxData(context.Context, *pb.SdmxDataQuery) (*pb.SdmxDataResult, error)
+}
+
+// NodeFetchAll fetches all NodeResponse pages for a given request by repeatedly calling ds.Node
+// as long as a NextToken is returned and merges into single response.
+func NodeFetchAll(ctx context.Context, ds DataSource, req *pbv2.NodeRequest, pageSize int) (*pbv2.NodeResponse, error) {
+
+	// NOTE: Consider introducing datasource-specific default page sizes in the future.
+	if pageSize <= 0 {
+		return nil, fmt.Errorf("pageSize must be positive")
+	}
+
+	// Make initial call.
+	resp, err := ds.Node(ctx, req, pageSize)
+	if err != nil {
+		slog.Error("NodeFetchAll: initial fetch failed", "error", err)
+		return nil, err
+	}
+
+	if resp.NextToken == "" {
+		return resp, nil
+	}
+
+	// Clone the request to avoid modifying the caller's object when setting NextToken.
+	reqClone := proto.Clone(req).(*pbv2.NodeRequest)
+	
+	// Initialize accumulated response with data from the first page.
+	accumulatedResp := resp
+	
+	for accumulatedResp.NextToken != "" {
+		reqClone.NextToken = accumulatedResp.NextToken
+		
+		nextResp, err := ds.Node(ctx, reqClone, pageSize)
+		if err != nil {
+			slog.Error("NodeFetchAll: subsequent fetch failed", "nextToken", reqClone.NextToken, "error", err)
+			return nil, err
+		}
+		
+		// Merge nextResp into accumulatedResp.
+		accumulatedResp.NextToken = nextResp.NextToken
+		for subjectID, newGraph := range nextResp.Data {
+			accumulatedGraph, ok := accumulatedResp.Data[subjectID]
+			if !ok {
+				accumulatedResp.Data[subjectID] = newGraph
+				continue
+			}
+			for prop, newNodes := range newGraph.Arcs {
+				accumulatedNodes, ok := accumulatedGraph.Arcs[prop]
+				if !ok {
+					accumulatedGraph.Arcs[prop] = newNodes
+					continue
+				}
+				accumulatedNodes.Nodes = append(accumulatedNodes.Nodes, newNodes.Nodes...)
+			}
+		}
+	}
+
+	return accumulatedResp, nil
 }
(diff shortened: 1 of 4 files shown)


internal/server/datasource/datasource.go
```diff
@@ -46,3 +49,60 @@ type DataSource interface {
...
+			for prop, newNodes := range newGraph.Arcs {
+				accumulatedNodes, ok := accumulatedGraph.Arcs[prop]
+				if !ok {
+					accumulatedGraph.Arcs[prop] = newNodes
+					continue
+				}
+				accumulatedNodes.Nodes = append(accumulatedNodes.Nodes, newNodes.Nodes...)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
clincoln8: updated

internal/server/datasource/datasource_test.go
```diff
@@ -0,0 +1,213 @@
...
+type mockDataSource struct {
+	DataSource
+	nodeFunc func(ctx context.Context, req *pbv2.NodeRequest, pageSize int) (*pbv2.NodeResponse, error)
+}
+
+func (m *mockDataSource) Node(ctx context.Context, req *pbv2.NodeRequest, pageSize int) (*pbv2.NodeResponse, error) {
+	return m.nodeFunc(ctx, req, pageSize)
+}
+
+func TestNodeFetchAll(t *testing.T) {
```
keyurva: totally optional: the tests can be simplified using table-driven test cases like below:

```golang
type mockDataSource struct {
	DataSource
	responses []*pbv2.NodeResponse
	errs      []error
	calls     int
}

func (m *mockDataSource) Node(ctx context.Context, req *pbv2.NodeRequest, pageSize int) (*pbv2.NodeResponse, error) {
	if m.calls >= len(m.responses) {
		return nil, errors.New("too many calls")
	}
	resp := m.responses[m.calls]
	err := m.errs[m.calls]
	m.calls++
	return resp, err
}

// makeNodeResponse helper generates the standard test response graph.
func makeNodeResponse(nextToken string, childDcids ...string) *pbv2.NodeResponse {
	var nodes []*pb.EntityInfo
	for _, dcid := range childDcids {
		nodes = append(nodes, &pb.EntityInfo{Dcid: dcid})
	}
	return &pbv2.NodeResponse{
		NextToken: nextToken,
		Data: map[string]*pbv2.LinkedGraph{
			"geoId/06": {
				Arcs: map[string]*pbv2.Nodes{
					"containedInPlace": {Nodes: nodes},
				},
			},
		},
	}
}

func TestNodeFetchAll(t *testing.T) {
	ctx := context.Background()
	req := &pbv2.NodeRequest{Nodes: []string{"geoId/06"}}

	cases := []struct {
		name      string
		responses []*pbv2.NodeResponse
		errs      []error
		want      *pbv2.NodeResponse
		wantErr   bool
	}{
		{
			name: "SinglePage",
			responses: []*pbv2.NodeResponse{
				makeNodeResponse("", "geoId/06001"),
			},
			errs: []error{nil},
			want: makeNodeResponse("", "geoId/06001"),
		},
		{
			name: "MultiPage",
			responses: []*pbv2.NodeResponse{
				makeNodeResponse("token1", "geoId/06001"),
				makeNodeResponse("token2", "geoId/06002"),
				makeNodeResponse("", "geoId/06003"),
			},
			errs: []error{nil, nil, nil},
			want: makeNodeResponse("", "geoId/06001", "geoId/06002", "geoId/06003"),
		},
		{
			name: "InitialError",
			responses: []*pbv2.NodeResponse{nil},
			errs:      []error{errors.New("initial error")},
			wantErr:   true,
		},
		{
			name: "SubsequentError",
			responses: []*pbv2.NodeResponse{
				makeNodeResponse("token1", "geoId/06001"),
				nil,
			},
			errs: []error{nil, errors.New("subsequent error")},
			wantErr: true,
		},
	}

	for _, tc := range cases {
		t.Run(tc.name, func(t *testing.T) {
			mockDS := &mockDataSource{
				responses: tc.responses,
				errs:      tc.errs,
			}

			got, err := NodeFetchAll(ctx, mockDS, req, 10)
			if (err != nil) != tc.wantErr {
				t.Fatalf("NodeFetchAll() error = %v, wantErr %v", err, tc.wantErr)
			}
			if tc.wantErr {
				return
			}

			if diff := cmp.Diff(tc.want, got, protocmp.Transform()); diff != "" {
				t.Errorf("NodeFetchAll() mismatch (-want +got):\n%s", diff)
			}
		})
	}
}

```

keyurva (approved):
Thanks!

internal/server/datasource/datasource_test.go
clincoln8: done

PR #1896, opened 2026-05-14
Implement topic hierarchy cache with background refresh

keyurva:
* Add data structures and protocol buffer representations for storing topic hierarchy and variables.
* Implement synchronous loading and recursive hierarchy building from Spanner and Knowledge Graph.
* Integrate Redis as an L2 cache to persist hierarchies across server restarts.
* Add a background refresher to periodically sync the hierarchy and maintain cache freshness.
* Wire up the cache manager at server startup for immediate availability.


internal/server/topic/fetcher.go
```diff
@@ -0,0 +1,335 @@
...
+	for _, n := range nodes.GetNodes() {
+		dcid := n.GetDcid()
+		if dcid == "" {
+			continue
+		}
+		topics[dcid] = &pbv2.TopicNode{
+			Dcid: dcid,
+			Name: extractName(n),
+		}
+	}
```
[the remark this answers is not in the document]
clincoln8: Would using datasource.NodeFetchAll work here?

proto/topic.proto
clincoln8, line 1: Will we be exposing this in the V2 API? If not, consider moving this to proto/topic.proto and keeping proto/v2/ reserved for public APIs.

internal/server/topic/topic_cache.go
```diff
@@ -0,0 +1,199 @@
...
+
+			for {
+				select {
+				case <-m.stopCh:
+					return
+				case <-ctx.Done():
+					return
+				case <-m.ticker.C:
+					slog.Info("Background topic cache refresher triggered")
+					if _, err := m.LoadHierarchy(ctx); err != nil {
```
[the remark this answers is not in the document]
clincoln8: Is gemini trying to tell the "master class of distributed systems" guru to bypass his own l2 cache?

internal/server/topic/topic_cache.go
```diff
@@ -0,0 +1,199 @@
...
+// It first checks the local L1 in-memory cache.
+// If L1 is empty/cold, it falls back to LoadHierarchy to load it from Redis or the KG.
+func (m *TopicCacheManager) GetHierarchy(ctx context.Context) (*pbv2.TopicHierarchy, error) {
+	m.mu.RLock()
+	if m.cache != nil && m.cache.TopicHierarchy != nil {
+		h := m.cache.TopicHierarchy
+		m.mu.RUnlock()
+		return h, nil
+	}
+	m.mu.RUnlock()
```
clincoln8: should this call CachedHierarchy?
```suggestion
if h := m.CachedHierarchy(); h != nil {
    return h, nil
}
```

internal/server/topic/topic_cache.go
```diff
@@ -0,0 +1,199 @@
...
+		TopicHierarchy: hierarchy,
+	}
+
+	slog.Info("Topic cache updated in memory", "cache", m.cache.String())
+}
+
+// GetHierarchy retrieves the cached TopicHierarchy.
+// It first checks the local L1 in-memory cache.
+// If L1 is empty/cold, it falls back to LoadHierarchy to load it from Redis or the KG.
+func (m *TopicCacheManager) GetHierarchy(ctx context.Context) (*pbv2.TopicHierarchy, error) {
```
clincoln8: Callers can mutate the cached proto since GetHierarchy returns a pointer. Should we return a clone, comment this as intentionally it read-only, or is direct mutation expected?

internal/server/topic/fetcher.go
clincoln8, line 1: anytime we do `return fmt.Errorf("...", err)`, should we log an error in addition to just returning?

internal/server/topic/topic_cache.go
```diff
@@ -0,0 +1,199 @@
...
+		} else if err != nil {
+			slog.Error("Failed to read topic cache from Redis", "error", err)
+		}
+	}
+
+	// L2 Miss: synchronous load from KG
+	slog.Info("Topic cache miss: loading synchronously from KG")
+	hierarchy, err := m.FetchTopicsFromKG(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load topic cache from KG during miss: %w", err)
```
clincoln8, line 181: nit: maybe log error here too? or  might not be needed if we add errors logs in fetcher.go

internal/server/topic/golden/topic_cache_golden_test.go
clincoln8, line 1: jetski noticed that in all the tests, we aren't testing using a mocked redis client for l2 cache behavior. Should we add some automated testing for that?

clincoln8 (commented):
Thanks!! I've left some comments for this first round, but I haven't gotten to `fetcher.go` yet. I'll do a second pass for it soon!

internal/server/topic/fetcher.go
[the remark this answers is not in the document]
keyurva: Yes. Done.

internal/server/topic/fetcher.go
```diff
@@ -0,0 +1,335 @@
...
+	resp, err := m.ds.Node(ctx, req, defaultPageSize)
+	if err != nil {
+		return fmt.Errorf("failed to fetch topic properties: %w", err)
+	}
+
+	for dcid, graph := range resp.GetData() {
+		if nodes, ok := graph.GetArcs()["relevantVariableList"]; ok {
+			topicToChildren[dcid] = parseTopicMembers(nodes, svpgSet)
+		}
+	}
```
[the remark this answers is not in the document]
keyurva: Number of topic nodes is manageable so does not need chunking right now.

internal/server/topic/fetcher.go
```diff
@@ -0,0 +1,335 @@
...
+	resp, err := m.ds.Node(ctx, req, defaultPageSize)
+	if err != nil {
+		return fmt.Errorf("failed to fetch svpg properties: %w", err)
+	}
+
+	for dcid, graph := range resp.GetData() {
+		if nodes, ok := graph.GetArcs()["memberList"]; ok {
+			svpgToMembers[dcid] = parseSvpgMembers(nodes)
+		}
+	}
```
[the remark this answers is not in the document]
keyurva: Number of svpg nodes is manageable so does not need chunking right now.

internal/server/topic/topic_cache.go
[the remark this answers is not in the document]
keyurva: I know - the audacity!

For future reference: in this case, we want to load from redis first which expires naturally based on TTL. If new data needs to be fetched before the TTL expires, which typically happens during deployments, we explicitly clear the cache which will trigger a KG fetch.

internal/server/topic/golden/topic_cache_golden_test.go
keyurva: L2 Redis cache interactions are tested via the `redis` package unit tests, so we pass `nil` for redis here.

internal/server/topic/fetcher.go
keyurva: Overall agree. jetski recommends only returning in internal functions and logging in "boundary" functions (basically public ones) like `LoadHierarchy`, `Start`, etc.

internal/server/topic/topic_cache.go
keyurva: Done. Added doc comment.

internal/server/topic/topic_cache.go
keyurva: Good call. Done.

internal/server/topic/topic_cache.go
keyurva: Done.

proto/topic.proto
keyurva: Good call. Not exposing in APIs. Moved.

keyurva (commented):
Thanks for the review!

internal/server/topic/fetcher.go
```diff
@@ -0,0 +1,336 @@
...
+	return relevantVarsMap
+}
+
+// fetchTopicMembers batches the datasource requests to fetch Topic children (relevantVariableList).
+// It populates topicToChildren mapping and registers linked SVPGs in svpgSet.
+func (m *TopicCacheManager) fetchTopicMembers(ctx context.Context, topicDcids []string, topicToChildren map[string][]string, svpgSet map[string]struct{}) error {
+	req := &pbv2.NodeRequest{
+		Nodes:    topicDcids,
+		Property: "->relevantVariableList",
```
clincoln8, line 224: just ooc, what was the reason for using the string list over the individual triples in "->relevantVariable"? easier parsing?

internal/server/topic/fetcher.go
```diff
@@ -0,0 +1,336 @@
...
+// expandTopicMembers expands referenced SVPG member variables in-memory under each topic, maintaining curated order.
+func expandTopicMembers(topicToChildren map[string][]string, svpgToMembers map[string][]string) map[string][]string {
+	relevantVarsMap := make(map[string][]string)
+	for topicDcid, children := range topicToChildren {
+		var expanded []string
+		seen := make(map[string]struct{})
+		for _, childDcid := range children {
+			if isSvpgDcid(childDcid) {
+				// Expand SVPG members directly inline
+				if members, ok := svpgToMembers[childDcid]; ok {
```
clincoln8, line 197: Should we add a warning log if we ever hit the else condition here, ie if an SVPG is referenced in topicToChildren but not found in svpgToMembers? Otherwise it's just silently dropped.

internal/server/topic/fetcher.go
```diff
@@ -0,0 +1,336 @@
...
+		if _, isChild := childTopicsSet[dcid]; !isChild {
+			roots = append(roots, dcid)
+			if dcid == defaultRootTopicDcid {
+				hasDefaultRoot = true
+			}
+		}
+	}
+
+	// Fallback Hack: If there are too many roots, and the default root "dc/topic/Root" exists,
+	// set it as the only root topic and log a warning instructing KG curation to be trimmed.
```
clincoln8: I'm curious to hear more about this decision and how we chose the threshold number.

In this case where we force the single default root, any other detected roots are dropped from the RootTopicDcids list. Is it okay that they are not linked under the single root and are unreachable just from the hierarchy perspective?

Instead of just dropping them, should we programmatically add the other roots as children of the default root here? That way they remain part of the hierarchy?

clincoln8 (approved):
Thanks!! Just a couple clarifications / nits. 
Masterful implementation :)

internal/server/topic/fetcher.go
keyurva: 2 reasons:
* To fetch all in 1 edge vs multiple (not the most important reason but helps and the list is already available)
* To maintain curated order of variables (not relevant for MCP but is for NL)

PR #1898, opened 2026-05-15
[v0/v1 migration] FilterStatVarsByEntity migration

nick-nlb:
## Issue

[b/513324218](b/513324218)

## Description

The `api/stats/stat-var-search` was broken under the following circumstances:

* VAI was enabled
* BT was not in use
* Entities were passed into the payload (these are used to filter the results from VAI.

No results were returned in these circumstances (whereas results were successfully filtered by sv/entity observation existence and returned when using BT).

The reason for this was that `FilterStatVarsByEntity` had not been migrated and was relying on the existence of the BT cache.

This PR creates a v3 (and diverted v2) implementation of the FilterStatVarsByEntity. 

## Testing

The problem is immediately visible on the [dev scatter tool](https://dev.datacommons.org/tools/scatter).

You can replicate it by:
* opening up the hierarchy without choosing a place and searching for "farms"
* the search will work (bring up results) as no filtering is occurring.
* delete the search from the search bar
* select United States as a place and "Counties" as the breakdown contained place.
* once again, search for "farms"
* you will now get no results.

Performing the same steps on production will give you a list of farm related stat vars that have been filtered to show only ones relevant to the place chosen.

### Local Testing

To replicate the scenario we see in dev, you can start mixer with:

```
./run_server.sh \
    --use_spanner_graph=true \
    --use_base_bigtable=false
```

The key flag is `use_base_bigtable`. By setting it to false, we disable BT, and in fact, we should be doing all our local testing this way if we want to truly test a Spanner only environment.

With big table set to false, in master you will see the same error as in dev. In this branch, you will see filtered results that should be identical to those found in production.


internal/server/datasources/datasources.go
```diff
@@ -213,3 +213,21 @@ func (ds *DataSources) SdmxData(ctx context.Context, in *pb.SdmxDataQuery) (*pb.
...
+		},
+		func(all []*pb.FilterStatVarsByEntityResponse) (*pb.FilterStatVarsByEntityResponse, error) {
+			if len(all) == 0 {
+				return &pb.FilterStatVarsByEntityResponse{}, nil
+			}
+			res := all[0]
+			for i := 1; i < len(all); i++ {
+				res = merger.MergeFilterStatVarsByEntityResponse(res, all[i])
+			}
+			return res, nil
```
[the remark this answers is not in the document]
nick-nlb: I effected this but in a way that matches more with how other similar functions did it (using a MergeMulti function).

internal/server/remote/client.go
```diff
@@ -130,3 +130,12 @@ func (rc *RemoteClient) BulkVariableGroupInfo(req *pbv1.BulkVariableGroupInfoReq
 	}
 	return resp, nil
 }
+
+func (rc *RemoteClient) FilterStatVarsByEntity(req *pb.FilterStatVarsByEntityRequest) (*pb.FilterStatVarsByEntityResponse, error) {
+	resp := &pb.FilterStatVarsByEntityResponse{}
+	err := util.FetchRemote(rc.metadata, rc.httpClient, "/v2/variable/filter", req, resp)
+	if err != nil {
+		return nil, err
```
clincoln8, line 140: nit: consider logging an error here before returning

internal/server/spanner/datasource.go
```diff
@@ -932,3 +932,51 @@ func filterVariableGroupInfo(data []*pbv1.VariableGroupInfoResponse, invalidNode
...
+	}
+
+	ids := make([]string, 0, len(svList))
+	for _, item := range svList {
+		ids = append(ids, item.Dcid)
+	}
+
+	rows, err := sds.client.CheckVariableExistence(ctx, ids, entities)
+	if err != nil {
+		return nil, fmt.Errorf("error checking variable existence: %w", err)
```
clincoln8, line 966: nit: either log an error here or ensure that that lower level method CheckVariableExistence has adequate logging.

clincoln8 (approved):
Thanks Nick, this is beautifully done!

Just a small nit to add some error logging. Mixer has been historically lacking in helpful logging/debugging traces, so we're trying to push for more observability in general (of course where reasonable).

nick-nlb:
> Thanks Nick, this is beautifully done!
> 
> Just a small nit to add some error logging. Mixer has been historically lacking in helpful logging/debugging traces, so we're trying to push for more observability in general (of course where reasonable).

Thank you very much for the review! Error logging is now added in!

PR #1900, opened 2026-05-15
Redis cache null pointer issue

nick-nlb:
## Description

This PR refactors the redisCacheClient initialization in main.go to declare the interface type directly. By instantiating the concrete client locally and deferring Close(), we prevent passing a typed-nil pointer to NewTopicCacheManager. This resolving the crash when running without Redis.


keyurva (approved):
Thanks for the quick fix Nick!

Tested locally with and without redis and both worked as expected!

PR #1902, opened 2026-05-18
[DCP - Mixer] Use V2API when Spanner Graph is enabled.

gmechali:
This PR sets the base feature flags file. 

In website, when we create the cdc_services image, we will update the run.sh script to make updates to this file before starting mixer.


clincoln8 (commented):
Instead of using the feature flags, can we just use the --use_spanner_graph mixer flag?

gmechali:
good catch thanks!
Closing this PR

PR #1907, opened 2026-05-20
Populate ProvenanceSummary for SQL-backed StatVarSummary

jm-rivera:
## Problem

`/api/variable/info` returns an empty `provenanceSummary` for StatisticalVariables sourced from custom imports on Custom DC instances. The base-DC path (Bigtable) populates the field correctly, and `placeTypeSummary` works for custom SVs, but `provenanceSummary` is never populated for SQL-backed data. Custom DC consumers are left without the facet metadata (provenance, measurement method, observation period, scaling factor, unit) they need to enumerate per-series details for their own variables.

Reported by ONE : https://issuetracker.google.com/issues/513611549

`sqlquery.GetStatVarSummaries` populates `PlaceTypeSummary` on the response proto but never sets `ProvenanceSummary`. The Bigtable-side helper populates both. The data needed to populate the missing fields is already present in the SQL schema (`observations` + `triples`) but not aggregated.

## Changes

- **New SQL aggregation** `getStatVarProvenanceSummaries` in `internal/sqldb/statements.go`. Groups `observations` by the 7-tuple `(variable, provenance, measurement_method, observation_period, scaling_factor, unit, entity_type)` and joins `triples` for the provenance display name. Uses the existing `observations_variable` and `triples_subject_id_predicate` indexes.
- **New model and client method**: `SVProvenanceSummary` in `internal/sqldb/model.go`, `GetSVProvenanceSummaries` on `SQLClient` in `internal/sqldb/query.go`.
- **`GetStatVarSummaries` now runs the existing and the new query in parallel** via `errgroup`, matching the convention at `server/statvar/statvar_summary.go`. An in-memory pivot folds per-(7-tuple) rows into the nested `ProvenanceSummary` / `SeriesSummary` / `PlaceTypeSummary` proto shape, computing per-series and per-provenance rollups in Go.
- **Extended SQLite test fixture** in `test/sqlquery/statvar_summary/` covers multi-provenance, multi-series, multi-entity-type cases, including the bare-ID fallback for provenances without a `name` triple.
- **Golden update** in `bulk_variable_info/bulk_bt_and_sql.json` for the SQL-backed `test_var_1`.

The new SQL query reads only from columns that already exist in the Custom DC schema (`observations.{variable, provenance, measurement_method, observation_period, scaling_factor, unit, entity, date, value}` and `triples.{subject_id, predicate, object_id, object_value}`). No schema migration required.

## Working example

This fix is already deployed on the ONE Data Commons production instance (`datacommons.one.org`, Cloud SQL MySQL backend), so the behavior is reproducible against a real Custom DC deployment without standing one up.

Pick any SQL-backed Custom SV and hit `/api/variable/info`. Example using a `CRS` (DAC aid flows) variable:
```bash
curl -sL -A "Mozilla/5.0" -H "Accept: application/json" \
  "https://datacommons.one.org/api/variable/info?dcids=ONE/CRS_DAC/Health-ODAGrants-GTM" \
  | python3 -m json.tool 
```

**Before this fix** the response for a SQL-backed SV looks like:

```json
{
  "ONE/CRS_DAC/Health-ODAGrants-GTM": {
    "placeTypeSummary": { "...": "..." }
  }
}
```

**After this fix** the same SV returns the populated structure:

```json
{
  "ONE/CRS_DAC/Health-ODAGrants-GTM": {
    "placeTypeSummary": { "Class": "...", "Country": "...", "GeoRegion": "..." },
    "provenanceSummary": {
      "c/p/2": {
        "importName": "CRS",
        "observationCount": 12982,
        "seriesSummary": [
          {
            "earliestDate": "1981",
            "latestDate": "2024",
            "observationCount": 950,
            "minValue": 434.05,
            "maxValue": 119753592.42,
            "seriesKey": {
              "measurementMethod": "Commitment",
              "unit": "ConstantEUR_2023"
            },
            "placeTypeSummary": { "Class": "...", "Country": "...", "...": "..." }
          },
          "... 17 more series, one per (measurementMethod, unit) tuple ..."
        ]
      }
    }
  }
}
```

The 18 `seriesSummary` entries correspond to the 3 distinct measurement methods (`Commitment`, `GrantEquivalent`, `GrossDisbursement`) times the 6 unit variants (`EUR`, `GBP`, `USD`, plus their `ConstantXXX_2023` constant-currency series), each with its own date range and place-type rollup. This is exactly the facet metadata Custom DC consumers need to render per-series UIs.


internal/sqldb/statements.go
```diff
@@ -143,6 +144,108 @@ var statements = struct {
...
+					o.provenance             AS provenance,
+					o.measurement_method     AS measurement_method,
+					o.observation_period     AS observation_period,
+					o.scaling_factor         AS scaling_factor,
+					o.unit                   AS unit,
+					t.object_id              AS entity_type,
+					Min(o.date)              AS earliest_date,
+					Max(o.date)              AS latest_date,
+					Min(o.value + 0.0)       AS min_value,
+					Max(o.value + 0.0)       AS max_value,
```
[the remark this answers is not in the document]
jm-rivera: This matches the existing convention in this file, `getStatVarSummaries` at lines 95-96 already uses Min(o.value + 0.0) / Max(o.value + 0.0) for the same reason. Let me know if I should try to address this anyway (though I think it would require changing the type which seems like a bigger change than ideal)

internal/sqldb/sqlquery/statvar_summary.go
```diff
@@ -68,3 +85,126 @@ func toPlaces(entityIds []string) []*pb.StatVarSummary_Place {
...
+		measurementMethod, observationPeriod, scalingFactor, unit string
+	}
+	// seriesByVarProv dedups *SeriesSummary pointers by (variable, provenance, seriesKey)
+	// so repeated rows with the same key (one per entity_type) update the same series.
+	seriesByVarProv := map[string]map[string]map[seriesKey]*pb.StatVarSummary_SeriesSummary{}
+
+	for _, r := range rows {
+		summary := ensureSummary(summaries, r.Variable)
+
+		ps, ok := summary.ProvenanceSummary[r.Provenance]
```
keyurva, line 129: nit: Can we create helpers here for better readability? e.g. populateProvenanceSummary, populatePlaceTypeSummary, sortSummaries, etc.

internal/sqldb/statements.go
```diff
@@ -143,6 +144,108 @@ var statements = struct {
 		SELECT *
 		FROM   aggregate;
 	`,
+	getStatVarProvenanceSummaries: `
+		WITH obs_groups
```
keyurva, line 148: Can you comment on the performance of this query with your data? Providing some context on the number of obs, variables, triples, etc. will help.

internal/sqldb/statements.go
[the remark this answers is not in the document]
keyurva: Agree. This matches the existing convention so ok to go with this for now.

keyurva (commented):
Thanks for putting this together, Jorge! Had a question on performance but if it is reasonable, happy to get this in!

internal/sqldb/sqlquery/statvar_summary.go
jm-rivera: Done!

internal/sqldb/statements.go
jm-rivera: Performance really depends on what people need to pull / how many statvars. Generally this isn't the best solution, it is computing a lot. 

For a single mid-sized StatVar (8.5k obs): ~600ms
For a single large StatVar (24k obs): ~1700ms
For 5 StatVars at once (111k obs): ~14s

Our short-term use case involves pulling for 1 StatVar at a time. 

The challenge is that SQL path scales linearly with total observations (it aggregates on read), while the BT path reads a precomputed row. Improving the performance here would require precomputed StatVarSummary materialisation at import time, which doesn't seem worth it since you're supporting this with DCP / Spanner soon.

internal/sqldb/sqlquery/statvar_summary.go
keyurva: Thanks.

internal/sqldb/statements.go
keyurva: Looks like perf won't be great in some cases but ok to go ahead given DCP.

proto/service/mixer.proto
```diff
@@ -172,7 +173,18 @@ service Mixer {
...
-      post : "/v2/variable/filter" 
+      post : "/v2/variable/filter"
+      body : "*"
+    };
+  }
+
+  // Returns precomputed date coverage for the requested variables (and
+  // optionally entities) from the in-memory coverage map. Custom-DC only;
+  // absent variables and pairs are omitted from the response.
+  rpc V2VariableCoverage(VariableCoverageRequest)
```
keyurva: A new endpoint will need discussion. Any chance you can exclude it from this PR? We can get the summary functionality submitted and handle this separately?

keyurva (commented):
Thanks for the updates, Jorge. If you're ok with excluding the coverage endpoint from here, we can get this merged right away. If you'd like to include it, let's have a discussion around it.

proto/service/mixer.proto
jm-rivera: Sorry @keyurva! I was accidentally committing to the same branch in our fork on some additional experiments (more on that soon). I've removed the last 2 commits to bring it back to a focused PR  on the provenanceSummary

keyurva: No worries. Thanks for reverting.

keyurva (approved):

jm-rivera:
Just in case, a note that I do not have write access to this repo so I can't merge myself. I've autoupdated the branch so it has all the latest from upstream, if helpful.

keyurva:
> Just in case, a note that I do not have write access to this repo so I can't merge myself. I've autoupdated the branch so it has all the latest from upstream, if helpful.

Oh sorry, didn't realize you didn't have merge access. I've enabled auto merge. Can you rebase the branch? I don't see any option to do that. I've also re-triggered the failing checks. Hope it auto merges once these have been addressed. Feel free to ping if it does not.

keyurva:
Ah - looks like it merged after all. All good!

PR #1909, opened 2026-05-26
[v0/v1 migration] Implement existence checks for variable groups against entities

nick-nlb:
## Issue

b/514069768

## Related PR

This is a follow up of [1892](https://github.com/datacommonsorg/mixer/pull/1892).

## Description

In Mixer’s Spanner implementation we support SV/place existence checks, and SV/source, SVG/source and topic/source existence checks. The added in the SVG/source and topic/source checks in the related PR linked above. However, we did not yet yet support SVG/place and topic/place existence checks.

This PR adds existence checks for topics and SVGs against observations. 

The website never checks SVGs against places using this existence check, but does use topic/place existence checks in the NL search.

## Testing

```
curl -X POST 'http://127.0.0.1:8081/v2/observation' \
-H 'Content-Type: application/json' \
-d '{
  "select": ["variable", "entity"],
  "entity": {
    "dcids": [
      "geoId/06",
      "country/USA"
    ]
  },
  "variable": {
    "dcids": ["dc/g/Demographics"]
  }
}'
{"byVariable":{"dc/g/Demographics":{}}}
```

```
curl -X POST 'http://127.0.0.1:8081/v2/observation' \
-H 'Content-Type: application/json' \
-d '{
  "select": ["variable", "entity"],
  "entity": {
    "dcids": [
      "geoId/06",
      "country/USA"
    ]
  },
  "variable": {
    "dcids": ["dc/topic/Demographics"]
  }
}'
```

The above two `curl` calls test the SVG and topic branches respectively. These should produce no results in master, but results in this branch.

Note that by no results, we mean that the query will return with the variable list, but the variable list will contain no places. 

```
{"byVariable":{"dc/g/Demographics":{}}}%
```

If the query successfully determined existence, it will return something like this:

```
{"byVariable":{"dc/g/Demographics":{"byEntity":{"country/USA":{},"geoId/06":{}}}}}%
```


internal/server/spanner/statements.go
```diff
@@ -684,4 +686,10 @@ var statements = struct {
 		  AND e2.predicate IN ('source', 'isPartOf')
 		  AND e3.predicate = @predicate
 		  AND e3.object_id IN UNNEST(@variables)`,
+	checkGroupPlaceExistence: `		SELECT DISTINCT e.object_id AS variable, o.observation_about AS entity
+		FROM Edge@{FORCE_INDEX=InEdge} e
+		JOIN@{JOIN_TYPE=APPLY_JOIN} Observation@{FORCE_INDEX=VariableMeasuredObservationAbout} o ON e.subject_id = o.variable_measured
+		WHERE e.predicate = @predicate
+		  AND e.object_id IN UNNEST(@variableGroups)
+		  AND o.observation_about IN UNNEST(@entities)`,
```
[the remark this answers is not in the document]
nick-nlb: This is done on the query added in this PR, with the caveat that this needs to also be done to other queries in this file (created in another bug).

internal/server/spanner/query_builder.go
```diff
@@ -253,6 +253,18 @@ func FilterStatVarsByEntityQuery(variables []string, entities []string) (*spanne
 	}, nil
 }
 
+// CheckGroupPlaceExistenceQuery returns a query to check existence of svg/topic for places.
+func CheckGroupPlaceExistenceQuery(variableGroups []string, entities []string, predicate string) *spanner.Statement {
```
keyurva, line 257: Can we add a query builder test for this? Helps to see what the final query looks like.

internal/server/spanner/client.go
```diff
@@ -39,6 +39,7 @@ type SpannerClient interface {
 	GetObservations(ctx context.Context, variables []string, entities []string) ([]*Observation, error)
 	CheckVariableExistence(ctx context.Context, variables []string, entities []string) ([][]string, error)
 	CheckVariableSourceExistence(ctx context.Context, variables []string, sources []string, predicate string) ([][]string, error)
+	CheckVariableGroupPlaceExistence(ctx context.Context, variableGroups []string, entities []string, predicate string) ([][]string, error)
```
keyurva, line 42: nit: Since this can be used for both groups and topics, the absence of the latter in the name could cause confusion. Consider either changing the method name or adding a comment for clarification.

keyurva (approved):
Thanks Nick!

internal/server/spanner/client.go
nick-nlb: This points to an existing "nomenclature" issue when working with SVGs and topics: the need for a single concept that describes them both.

I ended up going with using the word  "variable group" includes both SVGs and topic (i.e., any "folder" or "bundle" of variables). It follows how we use it in the sense of the bulk variable group endpoint (which supports topic). I liked "variable group" over just "group" (which is also used in some places - because group is very abstract).

As you say, that isn't really explicitly commented anywhere.

I updated the comment for this function itself to note that variable group means either SVG or topic.

PR #1910, opened 2026-05-30
[v2/resolve] Enable base_multi_entity index via request header

clincoln8:
### Goal
Allow clients to dynamically select the embeddings index for V2 Resolve requests via the `X-V2Resolve-Index` header.

### Approach & Implementation
*   **Request Header**: Added a new header `X-V2Resolve-Index` to allow clients to specify the desired index label, and a generic utility `util.GetSingleHeaderValue` to read it from the context.
*   **Index Mapping**: Updated the resolve flow to call a new method `resolve.SelectEmbeddingsIndex` to determine which index to query, allowing the client to override the server's default index. Supported labels are `"multi-entity"` (mapped to `"base_multi_entity"`) and `"base-nl"` (mapped to `"base_uae_mem"`).
* Deployment Configuration: Added base_multi_entity to the enabled indexes list in deploy/helm_charts/mixer/values.yaml to ensure it is loaded by the embeddings sidecar.

### Edge Cases, Errors & Logging
*   **Incompatible Header Label**: If the client requests a label that is not in our mapping (e.g., `invalid`), the system logs an `ERROR` and returns an `InvalidArgument` gRPC error, preventing fallback to an unintended index.
*   **No Header**: If no header is provided, it safely falls back to the default index without error or warning.

### Testing
Added unit tests in `embeddings_test.go` to thoroughly cover the new logic and defend against future regressions.
#### Local Integration Test:
* **Steps**: 
    * Started the embeddings server (`datacommons-nl` image) locally in a Docker container. 
    * Mounted a custom environment configuration (`custom_env.yaml`) to enable `base_multi_entity` and set `base_uae_mem` as the default index.
    * Verified that calling Mixer API without the header correctly routed to the default index.
    * Verified that calling Mixer API with the `X-V2Resolve-Index: multi-entity` header successfully routed the request to the `base_multi_entity` index in the sidecar. This was confirmed by searching for a specific statistical variable (`Annual_TwoWayTrade_EconomicActivity_NAICSMiningExceptOilAndGas_WithTradePartner`) that only exists in the `base_multi_entity` index, which returned a perfect match score (approx 1.0).

* **Environment File (`tmp/custom_env.yaml`) used:**
    ```yaml
    nl:
      env:
        enabled_indexes:
          - base_uae_mem
          - base_multi_entity
        default_indexes:
          - base_uae_mem
        vertex_ai_models:
          uae-large-v1-model:
            project_id: datcom-nl
            location: us-central1
            prediction_endpoint_id: "8110162693219942400"
    ```
* **Commands used for testing:**
    1. Start Embeddings Server:
    ```bash
    docker run -p 5001:6060 \
      -e NUM_WORKERS=1 \
      -e ADDITIONAL_CATALOG_PATH=/workspace/catalog.yaml \
      -e GOOGLE_APPLICATION_CREDENTIALS=/tmp/keys/creds.json \
      -v $PWD/tmp/custom_env.yaml:/workspace/deploy/helm_charts/envs/autopush.yaml \
      -v /Users/calinc/.config/gcloud/application_default_credentials.json:/tmp/keys/creds.json \
      gcr.io/datcom-ci/datacommons-nl:latest
    ```
    2. Start Mixer:
    ```bash
    ./run_server.sh \
        --embeddings_server_url=http://localhost:5001
    ```
    3. Verify via grpcurl:
    ```bash
    grpcurl -plaintext \
        -import-path proto \
        -proto proto/service/mixer.proto \
        -rpc-header "X-V2Resolve-Index: multi-entity" \
        -d '{"nodes": ["Annual Two Way Trade Economic Activity: Naics Naics/ 21 2"], "resolver": "indicator"}' \
        localhost:12345 datacommons.Mixer/V2Resolve
    ```

### Future Work / Follow-ups
* **Header Propagation in Custom DC**: Currently, the X-V2Resolve-Index header is not propagated to remote mixers when Custom DC routes requests. We need to decide how we want to handle this case and whether to propagate?

### DIFF
--- a/internal/featureflags/featureflags.go
+++ b/internal/featureflags/featureflags.go
@@ -49,8 +49,8 @@ type Flags struct {
 	// Whether to enable Spanner search embeddings.
 	// switch this to true in local.yaml for easier development
 	EnableSpannerSearchEmbeddings bool `yaml:"EnableSpannerSearchEmbeddings"`
-	// Whether to enable v2 resolve multi entity index.
-	EnableV2ResolveMultiEntity bool `yaml:"EnableV2ResolveMultiEntity"`
+	// Whether to enable dynamic index request for V2 Resolve.
+	EnableDynamicIndexRequest bool `yaml:"EnableDynamicIndexRequest"`
 }
 
 // setDefaultValues creates a new Flags struct with default values.
@@ -66,7 +66,7 @@ func setDefaultValues() *Flags {
 		UseStatisticalCalculation:     false,
 		EnableSDMXDataApi:             false,
 		EnableSpannerSearchEmbeddings: false,
-		EnableV2ResolveMultiEntity:    false,
+		EnableDynamicIndexRequest:    false,
 	}
 }
 
--- a/internal/featureflags/featureflags.go
+++ b/internal/featureflags/featureflags.go
@@ -49,6 +49,8 @@ type Flags struct {
 	// Whether to enable Spanner search embeddings.
 	// switch this to true in local.yaml for easier development
 	EnableSpannerSearchEmbeddings bool `yaml:"EnableSpannerSearchEmbeddings"`
+	// Whether to enable v2 resolve multi entity index.
+	EnableV2ResolveMultiEntity bool `yaml:"EnableV2ResolveMultiEntity"`
 }
 
 // setDefaultValues creates a new Flags struct with default values.
@@ -64,6 +66,7 @@ func setDefaultValues() *Flags {
 		UseStatisticalCalculation:     false,
 		EnableSDMXDataApi:             false,
 		EnableSpannerSearchEmbeddings: false,
+		EnableV2ResolveMultiEntity:    false,
 	}
 }
 
--- a/internal/featureflags/featureflags.go
+++ b/internal/featureflags/featureflags.go
@@ -49,8 +49,6 @@ type Flags struct {
 	// Whether to enable Spanner search embeddings.
 	// switch this to true in local.yaml for easier development
 	EnableSpannerSearchEmbeddings bool `yaml:"EnableSpannerSearchEmbeddings"`
-	// Whether to enable dynamic index request for V2 Resolve.
-	EnableDynamicIndexRequest bool `yaml:"EnableDynamicIndexRequest"`
 }
 
 // setDefaultValues creates a new Flags struct with default values.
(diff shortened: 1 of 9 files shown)


internal/server/handler_core.go
```diff
@@ -49,7 +50,11 @@ func (s *Server) V2ResolveCore(
 		if !s.flags.EnableEmbeddingsResolver {
 			return nil, status.Errorf(codes.Unimplemented, "Resolving indicators is not enabled for this environment.")
 		}
-		return resolve.ResolveUsingEmbeddings(ctx, s.httpClient, s.embeddingsServerURL, s.resolveEmbeddingsIndexes, in.Request.GetNodes(), in.TypeOfValues, adapter, in.Request.GetExpandTopics())
+		idx := s.resolveEmbeddingsIndexes
+		if s.useMultiEntity(ctx) {
+			idx = "base_multi_entity"
```
keyurva: nit: Consider creating a const and possibly a helper for index selection.

internal/server/handler_core.go
```diff
@@ -77,6 +82,17 @@ func (s *Server) V2ResolveCore(
 	}
 }
 
+func (s *Server) useMultiEntity(ctx context.Context) bool {
```
keyurva: Should we move this to resolver and make it return the index to use? If we make the header say `X-V2Resolve-Index` with values `multientity`, `allsvs`, etc., the method could return the corresponding index id or the default one.

WDYT?

keyurva (approved):
Thanks!

internal/server/handler_core.go
clincoln8: updated!

internal/featureflags/featureflags.go
```diff
@@ -49,8 +49,8 @@ type Flags struct {
 	// Whether to enable Spanner search embeddings.
 	// switch this to true in local.yaml for easier development
 	EnableSpannerSearchEmbeddings bool `yaml:"EnableSpannerSearchEmbeddings"`
-	// Whether to enable v2 resolve multi entity index.
-	EnableV2ResolveMultiEntity bool `yaml:"EnableV2ResolveMultiEntity"`
+	// Whether to enable dynamic index request for V2 Resolve.
+	EnableDynamicIndexRequest bool `yaml:"EnableDynamicIndexRequest"`
```
keyurva: your call: since this is fully opt-in functionality, wondering if we need a feature flag at all?

keyurva (approved):

internal/server/handler_core.go
clincoln8: done

internal/featureflags/featureflags.go
clincoln8: since it's opt-in and we want to test this against prod anyways, it makes sense to drop the feature flag. updated!

PR #1911, opened 2026-06-01
Add the Yaml file to be used in dcp website that reads feature featureflags

shixiao-coder:
This allows a default set of feature flags to be read in DCP website setup and also act as placeholder for DCP website setup script to append more feature flags based on environment variables


deploy/featureflags/custom.yaml
```diff
@@ -0,0 +1,2 @@
+flags:
+  EnableSDMXDataApi: true
```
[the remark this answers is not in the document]
shixiao-coder: added trailing line

deploy/featureflags/custom.yaml
```diff
@@ -0,0 +1,5 @@
+# This should only be used for DCP, do not use it for base DC
```
gmechali: This is in custom.yaml, so i meant only use it for DCP - not CDC!

gmechali (approved):

deploy/featureflags/custom.yaml
shixiao-coder: Ah, gotcha, just updated to do not use it for CDC

PR #1912, opened 2026-06-01
Filter inferior facets from spanner obs responses

n-h-diaz:
This is to fix a dev bug where the summary on the place page for Japan was showing different data sources (and a more recent population date) than in prod.

This is because for legacy V2 we [filter out "inferior" (ie low rank) facets](https://github.com/datacommonsorg/mixer/blob/f381be7cac0718ca84b9590d6e277974fbd40ad4/internal/server/v2/observation/direct.go#L135-L142) from certain requests, including for the population request for the place page summary (which comes from /api/observations/point?entities=country/JPN&variables=Count_Person, which will call /v2/observation?variable.dcids=Count_Person&entity.dcids=country/JPN&date=LATEST&select=variable&select=entity&select=date&select=value).

This PR updates the spanner implementation to do the same truncation. 

Tested with a local mixer with spanner that the place page is now showing the same response
image


internal/server/spanner/dsutil.go
```diff
@@ -471,7 +472,10 @@ func mergeEntityOrderedFacets(
 }
 
 func obsToObsResponse(req *pbv2.ObservationRequest, observations []*Observation) *pbv2.ObservationResponse {
-	response := generateObsResponse(req.Variable, observations, true /*includeObs*/)
+	// Whether the response should filter out low ranked facets from the response.
+	// Note that the behavior differs for containedInPlace requests vs direct requests.
+	shouldFilterInferiorFacets := (req.Entity.Expression != "" && req.Date != "") || (req.Entity.Expression == "" && req.Date == shared.LATEST)
```
nick-nlb: Mini-nit: To make this a bit more easy to mentally parse (and self documenting), could we break this up into separate named booleans? 

I.e., 
```
isContainedInWithDate := (req.Entity.Expression != "" && req.Date != "")
isDirectWithLatestDate := (req.Entity.Expression == "" && req.Date == shared.LATEST)
shouldFilterInferiorFacets = isContainedInWithDate || isDirectWithLatestDate
```

internal/server/v3/observation/golden/observation_test.go
```diff
@@ -59,10 +59,10 @@ func TestV3Observation(t *testing.T) {
 			{
 				req: &pbv2.ObservationRequest{
 					Variable: &pbv2.DcidOrExpression{
-						Dcids: []string{"AirPollutant_Cancer_Risk"},
+						Dcids: []string{"Count_Person"},
 					},
 					Entity: &pbv2.DcidOrExpression{
-						Dcids: []string{"geoId/01001", "geoId/02013"},
+						Dcids: []string{"country/JPN"},
```
nick-nlb: To double check here: 

In this case, we're replacing a test (rather than creating a new one). We should make sure that we aren't losing anything with the removal of the old test (if so, we could rather create a new test).

nick-nlb (approved):
A couple of small comments but tests well and looks good!

internal/server/spanner/dsutil.go
n-h-diaz: good idea, done - thanks!

internal/server/v3/observation/golden/observation_test.go
n-h-diaz: I don't think this would lose anything, but for completeness I added a new test instead (I'm hoping to clean up our spanner tests a bit, but can do this in follow up prs)

PR #1915, opened 2026-06-03
[v0/v1 migration] Targetted language filtering

nick-nlb:
## Issue

[b/515782168](b/515782168)

## Description

Mixer currently overfetches all available translations for a given node because our property expression syntax cannot filter by terminal node values. When using the Spanner backend for places with hundreds of translations, this overfetch exceeds Spanner's default page size. Mixer only retrieves the first page, meaning the specifically requested language is often truncated from the results, causing missing translations on the website.

As such, the fetch as we are currently performing it looks like this:

```
curl -X POST "http://localhost:8081/v2/node" \
-d '{
  "nodes": ["country/JPN"],
  "property": "->nameWithLanguage"
}'
```
An important takeaway is that while we can filter the terminal node by a subsequent edge (i.e. `{typeOf:City}`), mixer currently has no method to allow filtering on the terminal node itself. 

##  Solution

This solution implements query operators (in particular for this purpose, the `$lang` query operator that performs filtering on the terminal node.

It implements this for both Spanner and BT (with Spanner being the more significant implementation, because it involves an actual enhancement on the underlying query, whereas the BT branch simply filters the final results similar in principle to what Flask is doing on the frontend.

The query operators are implemented via a separate hander in anticipation of further query operators being added in the future.

## Testing

Once this PR is ready, a corresponding PR will be implemented on website to take advantage of this. In the meantime, the output can be tested with curl calls.

Run Mixer locally with Spanner (the commands are provided below but may be slightly different depending on your environment:

```
./run_server.sh \
    --use_spanner_graph=true \
    --use_base_bigtable=false

```

The following calls demonstrate the update:

### No query operator filter applied

This is what is currently being run on the website. The output should be the same both in masterand in our branch:

```
curl -X POST "http://localhost:8081/v2/node" \
-d '{
  "nodes": ["country/CAN"],
  "property": "->nameWithLanguage"
}'
```
Output:
```
{"data":{"country/JPN":{"arcs":{"nameWithLanguage":{"nodes":[{"provenanceId":"dc/base/WikidataOtherIdGeos","value":"An t-Seapan@gd"},{"provenanceId":"dc/base/WikidataOtherIdGeos","value":"An tSeapáin@ga"},
... [dozens more languages]
```

### Single arc with filter
```
curl -X POST "http://localhost:8081/v2/node" \
-d '{
  "nodes": ["country/JPN"],
  "property": "->nameWithLanguage{$lang:es}"
}'
```
```
{"data":{"country/JPN":{"arcs":{"nameWithLanguage":{"nodes":[{"provenanceId":"dc/base/WikidataOtherIdGeos","value":"Japón@es"}]}}}}}%
```

### Multi arc with filter (useful for selecting the default name in the same call)
```
curl -X POST "http://localhost:8081/v2/node" \
-d '{
  "nodes": ["country/JPN"],
  "property": "->[name, nameWithLanguage{$lang:es}]"
}'
```

```
{"data":{"country/JPN":{"arcs":{"name":{"nodes":[{"provenanceId":"dc/base/WikidataOtherIdGeos","value":"Japan"}]},"nameWithLanguage":{"nodes":[{"provenanceId":"dc/base/WikidataOtherIdGeos","value":"Japón@es"}]}}}}}
```

## Testing in BT

All of the above changes should also be visible with identical output using BT:

You can start Mixer locally in BT with:

```
./run_server.sh
```

## Codacy

Codacy errors are false positives.


internal/server/spanner/query_builder.go
```diff
@@ -70,8 +70,35 @@ const (
...
+	}
+	var parts []string
+	for j, lang := range langs {
+		paramName := fmt.Sprintf("%s_%d", prefix, j)
+		// Database stores values exactly like "Value"@en
+		params[paramName] = "@" + lang
+		parts = append(parts, fmt.Sprintf("ENDS_WITH(n.value, @%s)", paramName))
+	}
+	return strings.Join(parts, " OR ")
+}
```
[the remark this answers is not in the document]
nick-nlb: Done (in both BT and Spanner) - with the caveat that the results will still of course be lowercase (and so the suffix may not precisely match the user's query string.

internal/server/spanner/query_builder.go
```diff
@@ -139,27 +180,66 @@ func GetNodeEdgesByIDQuery(ids []string, arc *v2.Arc, pageSize, offset int) *spa
...
+
+			// Sort operator keys to guarantee deterministic SQL output
+			opKeys := make([]string, 0, len(propFilters))
+			for k := range propFilters {
+				opKeys = append(opKeys, k)
+			}
+			sort.Strings(opKeys)
+
+			for _, opKey := range opKeys {
+				if handler, isOperator := queryOperatorRegistry[opKey]; isOperator {
```
n-h-diaz, line 209: do we need to support bracket filtering for non operator props too?

internal/server/v2/parser.go
```diff
@@ -106,39 +115,19 @@ func parseArc(arrow, expr string) (*Arc, error) {
 		if expr[i] == '{' {
 			// <-containedInPlace{p:v}
 			arc.SingleProp = expr[0:i]
+			if arc.SingleProp == "" {
+				return nil, status.Errorf(
+					codes.InvalidArgument, "invalid property expression: %s", rawExpr)
```
n-h-diaz: optional (here and elsewhere in parsing code): explain the syntax error to help clients understand valid requests

n-h-diaz (approved):
Thanks! cool parsing code :) 

Overall looks good, just had a high level question about extending the request syntax for bracket filtering and if we need to support that more generally (or alternatively if we should clarify to api users that this is only for $ filtering). Not blocking though

nick-nlb:
> Overall looks good, just had a high level question about extending the request syntax for bracket filtering and if we need to support that more generally (or alternatively if we should clarify to api users that this is only for $ filtering). Not blocking though

Definitely, I think we should follow this up with an extension of this to include more extensive filtering on terminal nodes. I'll start a document to outline the plan once this current milestone is met!

PR #1918, opened 2026-06-04
Add V2 definition for get locations rankings

n-h-diaz:
This PR creates /v2/place/ranking, which for now is just a wrapper of V1 with legacy=true.

This also updates the current response proto with the new fields that will be used in the eventual V2 response. These will not be used though until the spanner implementation is ready. 

I added a dummy test to show what the new response will look like -- please take a look and let me know if this will work! I can make any changes as needed. One small diff from what we discussed is that I removed the "orderedFacets" level and just am using "info" to reuse the existing "Ranking" message, but I can change if preferred.


juliawu (approved):
Thanks for the fast implementation! I mainly checked the protos and return structure, and this looks good to me.

PR #1919, opened 2026-06-04
Update the DCP configuration to default to use spanner embeddings

shixiao-coder:


deploy/featureflags/custom.yaml
```diff
@@ -3,3 +3,5 @@ flags:
   UseSpannerGraph: true
   UseStaleReads: true
   V2DivertFraction: 1.0
```
gmechali: can you just delete these other ones since they arent needed

deploy/featureflags/custom.yaml
```diff
@@ -3,3 +3,5 @@ flags:
...
+  EnableEmbeddingsResolver: false
+  EnableSpannerSearchEmbeddings: true
```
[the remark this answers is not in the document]
shixiao-coder: this is not master switch, tested locally

deploy/featureflags/custom.yaml
shixiao-coder: deleted

gmechali (approved):

PR #1922, opened 2026-06-04
[v0/v1 migration] Existence check queries: deterministic return order.

nick-nlb:
## Issue

[b/516807725](b/516807725)

## Description

Add ORDER BY clause to source and group existence SQL queries to force deterministic return order (to ensure consistent results for tests, as well as for payload return order in the future).


juliawu (approved):
Thanks!

PR #1923, opened 2026-06-05
[v0/v1 migration]  Update API to support multiple typeOf filters for place-in queries

nick-nlb:
## Description

When traversing recursive (`+`) wildcard properties such as `containedInPlace+`, Mixer previously restricted `typeOf` filter constraints on BT to exactly a single target entity type (`len(typeOfs) == 1`).

Prior to [Website PR #6328](https://github.com/datacommonsorg/website/pull/6328), the Flask frontend queried `containedInPlace+` without any `typeOf` filter constraint, retrieving all contained nodes and filtering unwanted types in Flask.

However, to fix paginated Spanner responses where hundreds of non-place entities caused relevant places to be dropped, PR #6328 updated the request to explicitly filter by multiple target child place type.

While Spanner was able to handle this new syntax, BT was not (triggering an error).

## Solution

This solution implements multi-type filter support for BT.

## Testing

[Japan Place Page](http://localhost:8080/place/country/JPN)

### When running locally on BT:

On master, this page will fail (place not found).

In this branch it will succeed.

### When running locally on Spanner:

Both will succeed.

### DIFF
--- a/internal/server/handler_core.go
+++ b/internal/server/handler_core.go
@@ -100,7 +100,7 @@ func (s *Server) V2NodeCore(
 			// Examples:
 			//   <-containedInPlace+{typeOf:City}
 			typeOfs, ok := arc.Filter["typeOf"]
-			if !ok || len(typeOfs) != 1 {
+			if !ok || len(typeOfs) == 0 {
 				return nil, status.Errorf(codes.InvalidArgument,
 					"invalid filter for %s", in.GetProperty())
 			}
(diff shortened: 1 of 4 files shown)


n-h-diaz (approved):
Thanks!

would it be useful to distinguish the type of the returned places?

nick-nlb:
> Thanks!
> 
> would it be useful to distinguish the type of the returned places?

That's a good idea. The lack of distinction becomes much more obvious when you pull multiple places at the same time. I'll add this in in a follow-up PR!

PR #1924, opened 2026-06-05
Add option to divert request to spanner via header

n-h-diaz:
This is to enable easier website diversion for specific V2 requests (vs random diversion). 

To use, set the header `"X-Divert-Spanner: true"`

For example
```
curl -X POST -H "X-Divert-Spanner: true" \
  http://localhost:8081/v2/observation \
  -d '{"variable": { "dcids": ["Count_Person"] }, "entity": { "dcids": ["country/USA"] }, "select": ["facet"]}'
```

Note that this is currently lower priority than the overall Mixer server config flag `use_spanner_graph` (so if that flag is true, all requests will be diverted regardless of header). However to use this header, spanner needs to be enabled with the experimental feature flag `UseSpannerGraph` and the corresponding `spanner_graph_info` needs to be set.

For example, to test the flag locally,
```
./run_server.sh     --feature_flags_path=$PWD/deploy/featureflags/local.yaml     --spanner_graph_info="$(cat deploy/storage/spanner_graph_info.yaml)" 
``` 
where local.yaml has `UseSpannerGraph: true`


internal/server/server.go
```diff
@@ -339,10 +340,19 @@ func (s *Server) TopicCacheManager() *topic.TopicCacheManager {
...
 	if s.useSpannerGraph {
 		return true
 	}
 
+	// Second, check if the specific request has the header to divert to Spanner.
+	if md, ok := metadata.FromIncomingContext(ctx); ok {
+		if vals := md.Get(util.XDivertSpanner); len(vals) > 0 && vals[0] == "true" {
+			return true
+		}
+	}
```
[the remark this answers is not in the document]
n-h-diaz: thanks, done!

internal/server/server.go
```diff
@@ -339,10 +339,17 @@ func (s *Server) TopicCacheManager() *topic.TopicCacheManager {
 
 // shouldDivertV2 returns true if the request should be diverted to the dispatcher.
 func (s *Server) shouldDivertV2(ctx context.Context) bool {
+	// First check if the overall Mixer Spanner flag is set.
 	if s.useSpannerGraph {
 		return true
 	}
 
+	// Second, check if the specific request has the header to divert to Spanner.
```
nick-nlb: A client could set the diversion x-header to be true against a mixer that doesn't have Spanner enabled. It won't work (as is mentioned in the PR), but the calls still go through (and produce no data).

If Spanner isn't enabled, should we have the diversion header completely ignored (making it a no-op)? This would make the flag setting from outside safer, as it would only have any effect on a mixer that has Spanner enabled. For a non-Spanner enabled mixer, the flag would just be ignored and it would give a regular response.

This same logic could also apply to the fractional diversion, but I'm not as worried about there because that is entirely internal to Mixer (whereas any client could set this x-header from outside against any mixer, even one not yet prepared for Spanner).

internal/server/server.go
```diff
@@ -339,10 +339,17 @@ func (s *Server) TopicCacheManager() *topic.TopicCacheManager {
...
+	if util.IsHeaderTrue(ctx, util.XDivertSpanner) {
+		return true
```
nick-nlb, line 356: Could we set up some unit tests to test the various combinations of headers and flags against the decision to divert? (i.e. that maps various combinations of the main spanner flag, the flags spanner flag, the x-header and the expected diversion result).

internal/server/server.go
```diff
@@ -339,10 +339,17 @@ func (s *Server) TopicCacheManager() *topic.TopicCacheManager {
 
 // shouldDivertV2 returns true if the request should be diverted to the dispatcher.
 func (s *Server) shouldDivertV2(ctx context.Context) bool {
+	// First check if the overall Mixer Spanner flag is set.
 	if s.useSpannerGraph {
```
nick-nlb, line 343: Not related directly to this PR, but I did find the naming of the two `useSpannerGraph` flags unclear.

If my interpretation is correct:
* s.useSpannerGraph is the flag set via `--use_spanner_graph`, and if this is set, both enables Spanner and sets mixer to use Spanner 100% of the time.
* if s.flags.useSpannerGraph is set to true, Spanner is enabled, but no traffic is diverted to Spanner unless the fractional diversion flag is set to some number (or with this PR, the diversion x-header is set.

Would it make more sense for the flag version to be called something like "enableSpannerGraph"? I definitely wouldn't make the change in this PR (as it would have a pretty big surface area) but a possibility for later.

Again, not for this PR!

nick-nlb (commented):
Thank you for this! It looks great.

Just the couple of comments and thoughts.

internal/server/server.go
n-h-diaz: good call! updated to not divert if neither flag is set regardless of headers

internal/server/server.go
n-h-diaz: good idea- added a test

internal/server/server.go
n-h-diaz: yep that's correct

and yes that makes sense - will clean up the flags in follow up

nick-nlb (approved):
LGTM! Thank you for this.

PR #1925, opened 2026-06-05
[v2/resolve] Routing Refactoring & Header Overrides

clincoln8:
## Goal
Refactor V2 Resolve routing logic to support embeddings-based resolution (Indicator resolver) via Spanner or legacy backend, controlled by a new boolean header and feature flags. Optimize the embeddings client to fetch available indexes at startup to avoid per-request latency.

## Key Aspects of the Approach
*   **Tri-state Routing Header**: Introduced `X-V2Resolve-Indicator-Spanner` header to allow forcing Spanner (`true`), forcing legacy (`false`), or falling back to default routing based on flags (`empty`).
*   **Startup Configuration Fetch**: Registered a lifecycle hook for `EmbeddingsServiceClient` to fetch and cache available indexes from the NL server at startup. This avoids making a blocking HTTP call to the NL server config endpoint on every request that requires index validation.
*   **Centralized HTTP to gRPC Error Mapping**: Added `util.HTTPStatusToGRPCCode` to map HTTP response statuses from the embeddings server to standard gRPC error codes.

## Minor but Important Details
*   Added `util.GetOptionalBoolHeader` to handle tri-state boolean header parsing with validation.
*   Configured the `http.Client` for embeddings service with a 10-second timeout to prevent indefinite hangs.
*   Optimized the default routing check order in `handler_v2.go` to evaluate feature flags before checking Spanner status (short-circuiting).

## Edge Cases, Errors & Logging
*   Invalid values for `X-V2Resolve-Indicator-Spanner` return `codes.InvalidArgument`.
*   If Spanner is forced via header but not enabled on the server, it fails fast with `codes.FailedPrecondition`.
*   If the NL server is unreachable at startup, the lifecycle hook fails, causing Mixer to fail-fast.
*   If indexes are not yet loaded (e.g., in tests or due to load failures), validation is skipped (lenient fallback) to avoid breaking requests.

## Testing & E2E Verification
*   **E2E/Manual Verification Setup**:
    *   Start a local `nl_server` (embeddings server) on port 5001: `--embeddings_server_url=http://localhost:5001`.
    *   Start Mixer pointing to it: `./run_server.sh ... --embeddings_server_url=http://localhost:5001`.
    *   *Note*: Since we now fetch indexes at startup, `nl_server` must be running *before* Mixer starts, otherwise Mixer will fail to start.

*   **Verification Scenarios (via `grpcurl`)**:
    *   **Place Resolver**: Verify that place queries succeed.
        ```bash
        grpcurl -plaintext -import-path ./proto -proto proto/service/mixer.proto -d '{"resolver": "place", "nodes": ["wikidataId/Q30"]}' localhost:12345 datacommons.Mixer/V2Resolve
        ```
    *   **Topic Resolver**: Verify that topic queries succeed and bypass Spanner dispatcher (handled locally).
        ```bash
        grpcurl -plaintext -import-path ./proto -proto proto/service/mixer.proto -d '{"resolver": "topic", "nodes": ["dc/topic/Agriculture"]}' localhost:12345 datacommons.Mixer/V2Resolve
        ```
    *   **Indicator Resolver (Default Route)**: Verify default routing to legacy remote service.
        ```bash
        grpcurl -plaintext -import-path ./proto -proto proto/service/mixer.proto -d '{"resolver": "indicator", "nodes": ["population of USA"]}' localhost:12345 datacommons.Mixer/V2Resolve
        ```
    *   **Indicator Resolver (Force Legacy)**: Verify that setting `X-V2Resolve-Indicator-Spanner: false` forces routing to legacy.
        ```bash
        grpcurl -plaintext -import-path ./proto -proto proto/service/mixer.proto -rpc-header 'x-v2resolve-indicator-spanner: false' -d '{"resolver": "indicator", "nodes": ["population of USA"]}' localhost:12345 datacommons.Mixer/V2Resolve
        ```
    *   **Indicator Resolver (Force Spanner - Disabled)**: Verify that setting `X-V2Resolve-Indicator-Spanner: true` when Spanner is not initialized fails immediately.
        ```bash
        grpcurl -plaintext -import-path ./proto -proto proto/service/mixer.proto -rpc-header 'x-v2resolve-indicator-spanner: true' -d '{"resolver": "indicator", "nodes": ["population of USA"]}' localhost:12345 datacommons.Mixer/V2Resolve
        ```
        *Expected Error*:
        ```
        ERROR: Code: FailedPrecondition Message: Spanner backend is not enabled in this mixer
        ```
    *   **Indicator Resolver (Invalid Header Value)**: Verify that invalid header values fail validation.
        ```bash
        grpcurl -plaintext -import-path ./proto -proto proto/service/mixer.proto -rpc-header 'x-v2resolve-indicator-spanner: invalid' -d '{"resolver": "indicator", "nodes": ["population of USA"]}' localhost:12345 datacommons.Mixer/V2Resolve
        ```
        *Expected Error*:
        ```
        ERROR: Code: InvalidArgument Message: Invalid X-V2Resolve-Indicator-Spanner header value: "invalid". Valid values are 'true' or 'false'
        ```
    *   **Valid Index Test**: Verify validation succeeds for a valid index (e.g. `base_uae_mem`).
        ```bash
        grpcurl -plaintext -import-path ./proto -proto proto/service/mixer.proto -rpc-header 'x-v2resolve-index: base_uae_mem' -d '{"resolver": "indicator", "nodes": ["population of USA"]}' localhost:12345 datacommons.Mixer/V2Resolve
        ```
    *   **Invalid Index Test**: Verify validation fails for an invalid index.
        ```bash
        grpcurl -plaintext -import-path ./proto -proto proto/service/mixer.proto -rpc-header 'x-v2resolve-index: non_existent_index' -d '{"resolver": "indicator", "nodes": ["population of USA"]}' localhost:12345 datacommons.Mixer/V2Resolve
        ```
        *Expected Error*:
        ```
        ERROR: Code: InvalidArgument Message: Embeddings index "non_existent_index" is not available in the embeddings server
        ```

## Follow-up Work (PR 2)
We will migrate `resolver=topic` requests to resolve through the dispatcher when Spanner is enabled. This will involve:
1. Allowing `SpannerDataSource` to delegate topic resolution to `TopicExpander` via closure-based lazy injection to break circular startup dependencies.
2. Decoupling `TopicCacheManager` initialization from the embeddings resolver flag.
3. Updating routing to divert topic resolver requests to the dispatcher when Spanner is enabled.


internal/server/v2/resolve/embeddings.go
```diff
@@ -302,3 +328,75 @@ func fetchSVPropertyInfos(ctx context.Context, topicExpander TopicExpander, resu
...
+	}
+
+	indexes, err := c.fetchAvailableIndexes(ctx)
+	if err != nil {
+		slog.Warn("Failed to fetch available embeddings indexes, skipping validation", "error", err)
+		return true
+	}
+
+	return c.checkIndexes(idx, indexes)
+}
```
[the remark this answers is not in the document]
clincoln8: this is temp dev-only workflows. i don't think we need to cache, unless the need explicitly arises

internal/server/spanner/datasource.go
keyurva, line 71: Given the number of places this is called, should we start using the `SpannerDataSourceOptions` for this? Ok to do in a follow up PR.

internal/server/v2/resolve/embeddings.go
```diff
@@ -51,19 +51,34 @@ var labelToIndex = map[string]string{
...
 
-// SelectEmbeddingsIndex determines the correct index to use for embeddings resolution.
-func SelectEmbeddingsIndex(ctx context.Context, defaultIndex string) (string, error) {
+type EmbeddingsServiceClient struct {
+	httpClient          *http.Client
+	embeddingsServerURL string
+	defaultIndexes      string
+}
+
+func NewEmbeddingsServiceClient(httpClient *http.Client, embeddingsServerURL string, defaultIndexes string) *EmbeddingsServiceClient {
```
keyurva, line 63: WDYT about an `Options` struct here?

keyurva, line 63: Should `defaultIndexes` be a slice?

internal/server/v2/resolve/embeddings.go
```diff
@@ -163,14 +174,29 @@ func callEmbeddingsServer(
 
 	if httpResp.StatusCode != http.StatusOK {
 		bodyBytes, _ := io.ReadAll(httpResp.Body)
-		slog.Error("Embeddings server returned non-200 status", "status_code", httpResp.StatusCode, "body", string(bodyBytes), "url", embeddingsServerURL, "queries", nodes)
-		return nil, status.Errorf(codes.Internal, "The resolution service encountered an error processing your request.")
+		slog.Error("Embeddings server returned non-200 status", "status_code", httpResp.StatusCode, "body", string(bodyBytes), "url", c.embeddingsServerURL, "queries", nodes)
+		
+		var grpcCode codes.Code
```
keyurva: nit: seems like a useful generic util. Does a `util.toGRPCCode` make sense?

internal/server/v2/resolve/embeddings.go
```diff
@@ -302,3 +328,79 @@ func fetchSVPropertyInfos(ctx context.Context, topicExpander TopicExpander, resu
...
+func (c *EmbeddingsServiceClient) fetchAvailableIndexes(ctx context.Context) (map[string]struct{}, error) {
+	if c.embeddingsServerURL == "" {
+		return nil, fmt.Errorf("embeddings server URL is not configured")
+	}
+
+	u, err := url.Parse(c.embeddingsServerURL)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse embeddings server URL: %w", err)
+	}
+	u.Path = path.Join(u.Path, "/api/server_config/")
```
keyurva, line 340: What are you thinking in terms of a similar config for spanner embeddings? File based, in spanner, something else?

I would've asked for consistency here but given this is going away soon, ok to get it from the nl server.

internal/server/handler_v2.go
```diff
@@ -117,6 +125,56 @@ func (s *Server) V2Resolve(
...
+			return true, nil
+		case util.V2ResolveIndicatorBackendLegacy:
+			// Force Legacy Remote Service
+			return false, nil
+		default:
+			if backendHeader != "" {
+				return false, status.Errorf(codes.InvalidArgument, "Invalid X-V2Resolve-Indicator-Backend header value: %q. Valid values are %q or %q", backendHeader, util.V2ResolveIndicatorBackendSpanner, util.V2ResolveIndicatorBackendLegacy)
+			}
+			// Default: use Spanner if configured AND default routing flag is true
+			return s.isSpannerEnabled() && s.flags != nil && s.flags.EnableSpannerSearchEmbeddings, nil
```
keyurva: nit: `s.flags != nil && s.flags.EnableSpannerSearchEmbeddings` can be the first thing we check.

internal/server/handler_core.go
```diff
@@ -49,11 +49,14 @@ func (s *Server) V2ResolveCore(
 		if !s.flags.EnableEmbeddingsResolver {
 			return nil, status.Errorf(codes.Unimplemented, "Resolving indicators is not enabled for this environment.")
 		}
-		idx, err := resolve.SelectEmbeddingsIndex(ctx, s.resolveEmbeddingsIndexes)
+		idx, err := s.embeddingsServiceClient.SelectIndex(ctx)
 		if err != nil {
 			return nil, err
 		}
-		return resolve.ResolveUsingEmbeddings(ctx, s.httpClient, s.embeddingsServerURL, idx, in.Request.GetNodes(), in.TypeOfValues, adapter, in.Request.GetExpandTopics())
+		if idx != "" && !s.embeddingsServiceClient.ValidateIndex(ctx, idx) {
```
keyurva, line 56: Looks like this will be called on every resolve call that includes the index header. And this method in turn calls fetch which calls the NL server for the config. Is this intentional?

Should we fetch and maintain the valid configs at startup to not incur this cost? Or not fetch them at all and simply propagate the parameter as is?

cmd/main.go
```diff
@@ -470,7 +471,8 @@ func main() {
 	slog.Info("Dispatcher initialized", "processorsCount", len(processors))
 
 	// Create server object
-	mixerServer := server.NewMixerServer(store, metadata, c, mapsClient, dispatcher, flags, *writeUsageLogs, *embeddingsServerURL, *resolveEmbeddingsIndexes, *useSpannerGraph, topicCacheManager)
+	embeddingsServiceClient := resolve.NewEmbeddingsServiceClient(&http.Client{}, *embeddingsServerURL, *resolveEmbeddingsIndexes)
```
keyurva: Jetski suggests to initialize the http client with a timeout. By default it has none.

keyurva (approved):
Thanks! Consider the fetch index question. Rest are nits and optional.

internal/server/v2/resolve/embeddings.go
clincoln8: done

internal/server/v2/resolve/embeddings.go
clincoln8: i think this requires some more discussion. we can chat about it during the sync tomorrow about embeddings schema?

internal/server/handler_core.go
clincoln8: modified this to do a fetch at server startup.

internal/server/handler_v2.go
clincoln8: done

internal/server/v2/resolve/embeddings.go
clincoln8: I think it's fine to keep as a string because the input flag expects this to be a comma-separated string and then we pass it to the embeddings server as a comma separated string. So we'd parse and then re-concatenate. 
I can make it a slice if there's strong desire for the validation.

internal/server/spanner/datasource.go
clincoln8: will discuss offline as a potential followup

internal/server/v2/resolve/embeddings.go
clincoln8: will discuss offline as a potential followup

cmd/main.go
clincoln8: added

PR #1927, opened 2026-06-05
Enable embeddings sidecar for default Mixer service group

clincoln8:
Enables the embeddings sidecar on the default service group. This ensures that endpoints like `/v2/agent/search_indicators`, which require embeddings search but are routed to the default service group, do not fail with FailedPrecondition errors in deployed environments.


keyurva (approved):
Thank you!

PR #1930, opened 2026-06-08
Update prod spanner db to January build

n-h-diaz:
To test with traffic on most up-to-date build and monitor metrics

As follow up:
- will enable incremental ingestion on January build 
- slowly ramp up mirroring traffic to 100%

Will revert to November build if there are any issues


keyurva (approved):

vish-cs:
Please let me know when we should switch incremental ingestion to the January instance.

vish-cs (approved):

PR #1932, opened 2026-06-09
Clean up unused spanner flags

n-h-diaz:
This PR cleans up the flags EnableV3 and UseStaleReads from feature flag yaml files. The functionality of these flags has been rolled into a single feature flag "UseSpannerGraph" and the relevant code changes have made it to production

Once each environment has been redeployed with the flags removed, I'll send a follow up PR to delete the flag definitions

This is part of the Mixer cleanup to prepare for spanner on base dc

### DIFF
--- a/deploy/featureflags/autopush.yaml
+++ b/deploy/featureflags/autopush.yaml
@@ -2,9 +2,7 @@ clusters:
   - projects/datcom-mixer-autopush/locations/us-central1/clusters/mixer-us-central1
 liveUrl: https://autopush.api.datacommons.org
 flags:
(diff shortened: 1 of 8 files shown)


gmechali (approved):

gmechali:
Well actually we should separate the prod* files in their own PR right? These flighst get deployed immediately on submit I believe.

n-h-diaz:
> Well actually we should separate the prod* files in their own PR right? These flighst get deployed immediately on submit I believe.

removed prod from this pr and will do as a follow up

Thanks for the review!

PR #1933, opened 2026-06-10
Ramp up v3 mirror traffic on prod to 80%

n-h-diaz:
To test with more traffic and monitor metrics 

Will ramp down if any issues


dhotchkiss (approved):
LGTM! 70 --> 80%

PR #1937, opened 2026-06-12
Provenance ids in facet definitions

nick-nlb:
## Issue
[b/519184012](b/519184012)

## Description

With the move to no longer prefix provenance ids with `dc/base/` in environments other than base, website can no longer rely on a clean mapping between a prefixed `importName` and a `provenanceId`.

Website was previously adding this prefix explicitly when constructing a provenance id from an importName when preparing for calls back to mixer to request metadata.

This PR adds provenanceId into the facet definition as returned by mixer endpoints, and incorporates it into the hash that defines a facet id. 

## Notes

BT does not return the provenanceId directly, and will likely not be updated to do so during the transition period from BT to Spanner. Because of this, Mixer creates the provenance id (by appending the prefix) before returning it. This is acceptable because BT will never contain provenance ids that do not follow this pattern.

Current Schema in Spanner is supported, with the caveat that the provenance ids here are backfilled. 

New Schema is not yet supported (as new schema in general is not supported in Mixer).

## Goldens

Because the hashing logic has been updated to include provenance ids, a significant number of goldens have been updated.

## Testing

Run your local mixer using Spanner:

```
./run_server.sh \
    --use_spanner_graph=true \
    --use_base_bigtable=false
```

Restart envoy

```
```

The following curl call should return provenance ids in the facets:

```
curl -X POST -H "Content-Type: application/json" -H "" \
  http://localhost:8081/v2/observation \
  -d '{
    "variable": { "dcids": ["Count_Person"] },
    "entity": { "dcids": ["country/USA"] },
    "select": ["variable", "entity", "value", "date", "facet"]
  }'
```

Restart Mixer using BT. 

```
./run_server.sh
```
The same curl call will also produce provenanceIds in the facets.


n-h-diaz (approved):
Thanks!

optional: use constant for prefix

PR #1938, opened 2026-06-12
Centralize V2 Resolve indicator routing and introduce X-V2Resolve-Indicator-Spanner header

clincoln8:
Changes:
* Centralized Routing Logic: Refactored the inline checks in V2Resolve into a dedicated helper shouldRouteResolveToDispatcher. The routing decision is specific to the resolver type:
  * Place Resolver ("place" or empty): Uses standard diversion logic (returns s.shouldDivertV2(ctx)).
  * Topic Resolver ("topic"): Always routes to the local legacy backend (false) because it relies on the local in-memory TopicCache. **Follow-up needed to handled this via dispatcher**
  * Indicator Resolver ("indicator"): Uses custom request-time routing:
    * Header Override: If X-V2Resolve-Indicator-Spanner is present:
       * "true": Forces routing to Spanner (fails fast with FailedPrecondition if Spanner is not enabled).
       * "false": Forces routing to the legacy backend.
    * Default (No Header): If the header is missing, it routes to Spanner only if BOTH the feature flag EnableSpannerSearchEmbeddings is true AND the Spanner database backend is enabled on the server (s.isSpannerEnabled()). Otherwise, it defaults to the legacy backend.
* Simplified Spanner Datasource: Removed the duplicate enableSpannerSearchEmbeddings flag check inside the SpannerDataSource constructor and methods. The Spanner datasource now assumes that any indicator request routed to it is valid to execute, centralizing the dispatch logic.

Verification:
* Spanner Disabled (Default Dev Server):
  * Default requests (no header) correctly route to legacy backend.
  * Header X-V2Resolve-Indicator-Spanner: false correctly routes to legacy backend.
  * Header X-V2Resolve-Indicator-Spanner: true correctly fails fast with FailedPrecondition ("Spanner backend is not enabled in this mixer").
  * Invalid header values correctly fail with InvalidArgument.
* Spanner Enabled (Feature Flag toggling): Created temporary feature flag configurations to test default routing behaviors
  * Verified default queries route to legacy backend when EnableSpannerSearchEmbeddings is false.
  * Verified default queries route to Spanner Vector Search when EnableSpannerSearchEmbeddings is true.
  * Verified that explicitly passing the header override X-V2Resolve-Indicator-Spanner: false successfully overrides a true feature flag, diverting the request to legacy.
  * Verified that explicitly passing X-V2Resolve-Indicator-Spanner: true successfully overrides a false feature flag, diverting the request to Spanner.


internal/server/handler_v2.go
```diff
@@ -117,6 +125,56 @@ func (s *Server) V2Resolve(
...
+
+	// Topic resolver always goes to V2ResolveCore (local) because it uses in-memory TopicCache
+	// TODO: implement resolver==topic handling in spanner datasource for dispatch based fulfillment
+	if resolver == resolve.ResolveResolverTopic {
+		return false, nil
+	}
+
+	// Indicator resolver (embeddings-based) has custom request-time toggling
+	if resolver == resolve.ResolveResolverIndicator {
+		divertVal, err := util.GetOptionalBoolHeader(ctx, util.XV2ResolveIndicatorSpanner)
```
keyurva, line 153: What do you think about this - if a request explicitly includes this header, we straightaway route to dispatcher. If spanner is not enabled, the request will fail anyway. Since they had explicitly opted in, that's probably a fair outcome.

If you agree, the rest of the logic becomes simpler.

I'm fine either way. Was suggesting simply because this flag is temporary so was thinking about the least amount of work needed to implement it and then remove it.

keyurva (approved):

internal/server/handler_v2.go
clincoln8: i'll just keep it as is since I already wrote it 😅
I like the idea of being able to override back to legacy if needed -- after we divert to spanner as the default, it might be helpful to explicitly request legacy embeddings for some time, just for comparison.

PR #1947, opened 2026-06-17
Migrate final schema reference to TimeSeries and Observation table

rohitkumarbhagat:
- Replace TimeSeries_final_v2 -> TimeSeries, Observation_final_v2 -> Observation
- Made final schema tables configurable
- removed querying observation attributes
- updated following fields
  - facets -> facet
  - facets_id -> facet_id
  - string_value to value


ajaits (approved):

internal/server/spanner/query_builder_multientity.go
```diff
@@ -26,8 +26,21 @@ import (
 	"github.com/datacommonsorg/mixer/internal/server/v2/shared"
 )
 
+func formatMultiEntityObservationSQL(sql string, cfg TableConfig) string {
+	return fmt.Sprintf(sql, cfg.ObservationTable, cfg.TimeSeriesTable)
+}
+
+func formatMultiEntityTimeSeriesSQL(sql string, cfg TableConfig) string {
```
keyurva, line 33: I had a few comments in the way to config is passed. Current approach couples statements and queries and requires passing the config to every builder function.

One suggestion is to create a statement struct and constructor in `statements_multientity.go`. Something like this:

```golang
type multiEntityStatements struct {
    getObsBasic             string
    getStatVarsByEntityBoth string
    // ...
}

func newMultiEntityStatements(cfg TableConfig) *multiEntityStatements {
    return &multiEntityStatements{
        getObsBasic: fmt.Sprintf("... FROM %s ...", cfg.ObservationTable),
        // ...
    }
}
```

Then store `*multiEntityStatements` on `multiEntityClient` and use the pointer for generating statements.

This centralizes sql construction in one file, eliminates runtime `fmt.Sprintf` on every request and removes explicit `TableConfig` passing.

rohitkumarbhagat: addressed in https://github.com/datacommonsorg/mixer/pull/1951

PR #1948, opened 2026-06-18
Enable SDMX flags in custom feature flags for DCP

clincoln8:
This PR enables UseMultiEntitySchema and EnableSDMXDataApi in custom.yaml for DCP deployments, and updates the comment to clarify its usage.


deploy/featureflags/custom.yaml
```diff
@@ -1,4 +1,6 @@
-# This should only be used for DCP, do not use it for CDC
+# This is read in for DCP deployments when resolve via spanner is enabled. It is not used for CDC.
 flags:
+  UseMultiEntitySchema: true
```
gmechali, line 3: So we kinda had a hack going on here https://github.com/datacommonsorg/website/blob/151ecc7c6457e7d001db810805790231175295d8/build/cdc_services/run.sh#L116

we only passed in the feature flags if the resolve with spanner embeddings was true. Meant to be temporary but if we re gonna use feature flags for this then we need to fix that 😓 

Can you confirm with Xiao that we will deprecate the  RESOLVE_WITH_SPANNER_EMBEDDINGS TF varaible now that its ready, and always pass in the custom.yaml? 

If we do that, then this LGTM. 

But also it's getting tough to see the distinction between feature flags and env vars for DCP...

gmechali (approved):

deploy/featureflags/custom.yaml
clincoln8: RESOLVE_WITH_SPANNER_EMBEDDINGS defaults to true for now. I'll discuss with Xiao to holistically re-assess how we want to do this. Agreed that this is confusing and a temporary stage as we figure out the long term.

PR #1950, opened 2026-06-22
Update variable metadata tool response per offline feedback

keyurva:
- Redesigned GetVariableMetadata response to use structured proto and lock-free concurrency
- Updated the response schema to return strongly-typed facet metadata.
- Refactored the backend handler to implement a lock-free concurrent fetch pipeline.
- Added an include list for dataset properties to prevent internal metadata leaks.
- Implemented graceful degradation for downstream metadata fetch failures.

### Local Verification

**Single variable + place with constrained properties (short response)**
```
http://localhost:8081/v2/agent/get_variable_metadata?variable_dcids=Mean_WagesDaily_Amount_Male_Rural_ConstructionWorker&entity_dcids=country/IND
```

**Multiple variables + places with no constrained props (large response)**
```
http://localhost:8081/v2/agent/get_variable_metadata?variable_dcids=Count_Person&variable_dcids=Median_Age_Person&entity_dcids=country/IND&entity_dcids=geoId/01
```

### DIFF
--- a/internal/agent/get_variable_metadata.go
+++ b/internal/agent/get_variable_metadata.go
@@ -16,16 +16,18 @@ package agent
 
 import (
 	"context"
-	"log/slog"
-	"sync"
+	"log"
+	"sort"
 	"time"
 
+	pb "github.com/datacommonsorg/mixer/internal/proto"
 	pbv1 "github.com/datacommonsorg/mixer/internal/proto/v1"
 	pbv2 "github.com/datacommonsorg/mixer/internal/proto/v2"
 	"github.com/datacommonsorg/mixer/internal/util"
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/structpb"
 )
 
 const (
@@ -33,7 +35,67 @@ const (
 	maxConcurrentFetchers = 20
 )
 
-var obsMetadataSelect = []string{"variable", "entity", "facet"}
+var (
+	obsMetadataSelect = []string{"variable", "entity", "facet"}
+
+	// properties to exclude from Statistical Variable properties Struct
+	excludedSVProperties = map[string]struct{}{
+		"name":           {},
+		"description":    {},
+		"label":          {},
+		"typeOf":         {},
+		"provenance":     {},
+		"definition":     {},
+		"memberOf":       {},
+		"linkedMemberOf": {},
+		"linkedMember":   {},
+	}
+
+	// properties to explicitly include for Provenance and Source dataset properties Struct.
+	// These must be the exact raw property keys returned from the knowledge graph (V2Node).
+	includedDatasetProperties = map[string]struct{}{
+		"url":                     {},
+		"license":                 {},
+		"licenseType":             {},
+		"lastDataRefreshDate":     {},
+		"nextDataRefreshDate":     {},
+		"nextSourceReleaseDate":   {},
+		"sourceReleaseFrequency":  {},
+		"latestObservationDate":   {},
+		"earliestObservationDate": {},
+		"isPartOf":                {},
+		"source":                  {},
+		"domain":                  {},
+		"description":             {},
+		"descriptionUrl":          {},
+	}
+)
+
+// rawVariableData holds the raw, unmapped graph and summary data for a single Statistical Variable.
+type rawVariableData struct {
+	properties *pbv2.LinkedGraph
+	summary    *pb.StatVarSummary
+}
+
+// rawDatasetData holds the raw, unmapped graphs for a dataset's provenance and parent source.
+type rawDatasetData struct {
+	provenance *pbv2.LinkedGraph
+	source     *pbv2.LinkedGraph
+}
+
+// rawFetchResult aggregates all raw data retrieved during the concurrent fetch phase.
+type rawFetchResult struct {
+	variables    map[string]*rawVariableData
+	observations *pbv2.ObservationResponse
+	provenances  map[string]*rawDatasetData
+}
+
+// translatedProperties holds the resolved name, description, and filtered dynamic properties Struct.
+type translatedProperties struct {
+	name        string
+	description string
+	properties  *structpb.Struct
+}
 
 // GetVariableMetadata assesses Statistical Variables by retrieving their definitions,
 // temporal/entity coverage, and source provenance descriptions.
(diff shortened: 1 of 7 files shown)


internal/agent/get_variable_metadata.go
```diff
@@ -87,326 +143,528 @@ func dedup(dcids []string) []string {
...
+		return nil, err
+	}
+
+	// Now hydrate provenances using the fetched data (lock-free!)
+	if err := s.fetchRawProvenances(ctx, raw); err != nil {
+		return nil, err
+	}
+
+	return raw, nil
+}
```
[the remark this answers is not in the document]
keyurva: Done.

internal/agent/get_variable_metadata.go
```diff
@@ -87,326 +143,528 @@ func dedup(dcids []string) []string {
...
-		if vMeta.StatVarSummary != nil {
-			for provID := range vMeta.StatVarSummary.GetProvenanceSummary() {
-				if pMeta, ok := provenancesMap[provID]; ok {
-					vMeta.Provenances[provID] = pMeta
-				}
-			}
+	// 2. Resolve root Description
+	if descArc, ok := graph.GetArcs()["description"]; ok {
+		tp.description = resolveFirstValue(descArc.GetNodes())
+	}
```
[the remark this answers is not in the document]
keyurva: Done.

internal/agent/get_variable_metadata.go
```diff
@@ -87,326 +143,528 @@ func dedup(dcids []string) []string {
...
+	if srcArc, ok := graph.GetArcs()["source"]; ok {
+		for _, n := range srcArc.GetNodes() {
+			if n != nil && n.GetDcid() != "" {
+				return n.GetDcid()
+			}
 		}
-		for eDcid, eObs := range vObs.GetByEntity() {
-			if eObs == nil {
+	}
```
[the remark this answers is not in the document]
keyurva: Done.

internal/agent/get_variable_metadata.go
```diff
@@ -87,326 +143,528 @@ func dedup(dcids []string) []string {
...
+		Properties: &structpb.Struct{Fields: make(map[string]*structpb.Value)},
 	}
-	if resp != nil && resp.GetData() != nil {
-		if graph, ok := resp.GetData()[pDcid]; ok && graph != nil {
-			props := toPropertyValuesMap(graph)
-			mu.Lock()
-			provenancesMap[pDcid] = &pbv2.GetVariableMetadataResponse_ProvenanceMetadata{
-				Dcid:       pDcid,
-				Properties: props,
+
```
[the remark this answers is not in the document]
keyurva: Done.

proto/v2/agent.proto
```diff
@@ -114,48 +112,49 @@ message GetObservationsResponse {
...
+
+    message DateRange {
+      string start = 1;
+      string end = 2;
+    }
+    DateRange date_range = 4;
+
+    message Scope {
+      repeated string entity_coverage = 1;
+      repeated string entity_granularity = 2;
```
clincoln8, line 137: is entity_granularity specific to places? How will this expand to non-places, like companies?

proto/v2/agent.proto
```diff
@@ -114,48 +112,49 @@ message GetObservationsResponse {
...
+    string provenance_id = 2;
+    int32 obs_count = 3;
+
+    message DateRange {
+      string start = 1;
+      string end = 2;
+    }
+    DateRange date_range = 4;
+
+    message Scope {
```
clincoln8, line 135: what if the message was `Coverage` and the first field was just `entities`?

```
message Coverage {
  repeated string entities = 1;
  repeated string granularities = 2; 
}
```

proto/v2/agent.proto
clincoln8, line 137: Will we consider giving back counts of each type?

internal/agent/get_variable_metadata.go
```diff
@@ -16,24 +16,86 @@ package agent
...
 
-var obsMetadataSelect = []string{"variable", "entity", "facet"}
+var (
+	obsMetadataSelect = []string{"variable", "entity", "facet"}
+
+	// properties to exclude from Statistical Variable properties Struct
+	excludedSVProperties = map[string]struct{}{
+		"name":           {},
+		"description":    {},
+		"label":          {},
```
clincoln8: Is this a property in the KG?

internal/agent/get_variable_metadata.go
```diff
@@ -16,24 +16,86 @@ package agent
...
+		"typeOf":         {},
+		"provenance":     {},
+		"definition":     {},
+		"memberOf":       {},
+		"linkedMemberOf": {},
+		"linkedMember":   {},
```
clincoln8: long term: is there a way for us to label "internal" properties to auto-exclude them from API respoonses like this one? This way we can have more control based on how we define the graph, rather than having hardcoded lists of properties in mixer.

For example, what if we made `linkedMember` subClassOf `generatedProperty` or `internalOnlyProperty`. And in this read path, we check the subClassOf for each property

internal/agent/get_variable_metadata.go
```diff
@@ -16,24 +16,86 @@ package agent
...
+	}
+
+	// properties to explicitly include for Provenance and Source dataset properties Struct.
+	// These must be the exact raw property keys returned from the knowledge graph (V2Node).
+	includedDatasetProperties = map[string]struct{}{
```
clincoln8, line 80: similarly, is there something we could define something in the graph so that this isn't a static, hardcoded list? Then this could also be customizable for DCP.

internal/agent/get_variable_metadata.go
```diff
@@ -87,326 +143,543 @@ func dedup(dcids []string) []string {
...
+	for i, vDcid := range varDcids {
+		idx := i
 		dcid := vDcid
+
+		// Worker 1: Fetch Node Graph (Non-critical: log and skip on error)
 		g.Go(func() error {
-			return s.fetchVariableProperties(gCtx, dcid, variables, &mu)
+			graph, err := s.fetchNodeGraph(gCtx, dcid)
+			if err != nil {
+				log.Printf("Warning: failed to fetch node graph for variable %s: %v", dcid, err)
```
clincoln8: is log.Printf the right thing here?

internal/agent/get_variable_metadata.go
```diff
@@ -45,31 +107,25 @@ func (s *Service) GetVariableMetadata(
...
+		return nil, status.Error(codes.InvalidArgument, "variable_dcids cannot be empty")
 	}
-
-	variables := initVariableMetadata(varDcids)
-
-	if err := s.fetchCoreMetadata(ctx, varDcids, entityDcids, variables); err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to fetch core variable metadata: %v", err)
+	if len(req.GetEntityDcids()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "entity_dcids cannot be empty")
 	}
```
clincoln8, line 147: should we cap the max number of variable_dcids and entity_dcids here (like max 50 vars / 100 places)? since we're doing concurrent fetches and a bulk Spanner query, would a massive request cause a latency spike or resource issues?

internal/agent/get_variable_metadata.go
```diff
@@ -87,326 +143,543 @@ func dedup(dcids []string) []string {
...
 
-	if err := g.Wait(); err != nil {
-		return err
+	if graph == nil || graph.GetArcs() == nil {
+		return tp
+	}
+
+	// 1. Resolve root Name (primary name or label)
+	if nameArc, ok := graph.GetArcs()["name"]; ok && nameArc != nil {
```
clincoln8: nit: should we make these property names constants?

internal/agent/get_variable_metadata.go
```diff
@@ -87,326 +143,543 @@ func dedup(dcids []string) []string {
...
+			if _, ok := stats.covered[reqPlace]; ok {
+				activeCoverage = append(activeCoverage, reqPlace)
+			}
+		}
 
-	var ids []string
-	for id := range seen {
-		ids = append(ids, id)
+		// Build flat dynamic properties map for facet details
+		fPropsMap := map[string]any{
```
clincoln8, line 540: long term: holistically, how do we want to handle facet metadata across these endpoints? right now, we're building a hardcoded map here in fPropsMap for a generic properties Struct, but elsewhere (like GetObservationsResponse) these are strongly-typed fields.

is there a cleaner, more systemic way to handle standard facet properties than maintaining these hardcoded maps in Go?

clincoln8 (approved):

internal/agent/get_variable_metadata.go
keyurva: No. Removed.

internal/agent/get_variable_metadata.go
keyurva: Probably. However for performance, we'll probably need to fetch and cache all such properties once vs looking up the relationship on every call.

internal/agent/get_variable_metadata.go
keyurva: That's a good call. I like this better than the sub-class concept. We can have, say, a `metadataProperties` property on a type node that define the properties that should be considered metadata.

internal/agent/get_variable_metadata.go
keyurva: Capping makes sense. And don't need such a high number for agentic calls. Using a max of 10 for now.

internal/agent/get_variable_metadata.go
keyurva: Changed to slog.

internal/agent/get_variable_metadata.go
keyurva: Done.

internal/agent/get_variable_metadata.go
keyurva: It's hardcoded currently because the facet object only supports this fixed set. I made the properties here a generic struct so we _can_ return arbitrary observation attributes once our apis support them.

proto/v2/agent.proto
keyurva: @carojahn - thoughts?

I'll submit as-is for now but will update in a follow-up PR based on Caro's input.

proto/v2/agent.proto
keyurva: Ideally these should be generic and not confined to just places.

These come from SV summary. For non-place entities we may need to define a similar hierarchy (say sector, sub-sector), populate SV summaries based on those hierarchies and return in this call.

keyurva: If SV summaries include it we can. Don't think we want to compute it on the fly.

PR #1958, opened 2026-07-01
[Stale Timestamp] Make mixer use the right stale timestamp. 

gmechali:
Change the logic for computing the staleness timestamp. Before we just sought out the most recent Successful run which opened up the possibility of this timestamp being out of retention period, which then defaults to strong reads. And that would mean the next time we start a run, we would be reading from a database as it's getting updated, defeating the purpse of the stale timestamp.

The update ingestion history logic will log a new entry when the ingestion starts (status pending), and updates it to running as it works through the stages. now the query will look at the creationTimestamp of the ingestionhistory row to determine the timestamp BEFORE the current ingestion started. 

The only way that we end up in a bad state would be if the duration of the ingestion is greater than the spanner retention window. 

Note that this introduces a mixer flag to use this new logic, and we remain backwards compatible on the old schema. We will be migrating Base DC before killing this logic.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -208,7 +208,7 @@ func main() {
 	var spannerClient spanner.SpannerClient
 	if shouldUseSpannerGraph {
 		var err error
-		spannerClient, err = spanner.NewSpannerClient(ctx, *spannerGraphInfo, flags.SpannerGraphDatabase, flags.UseMultiEntitySchema)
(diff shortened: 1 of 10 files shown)


clincoln8 (approved):
Looks good to me, thanks!! 

nit: You could wrap the flags in a SpannerClientOptions struct to keep the constructor clean to align with docs/architecture.md if you wanted. But feel free to merge as is, I'm planning a larger refactor of these interfaces anyways and can incorporate that.

PR #1964, opened 2026-07-08
[DCP - Stale Timestamp] Just use the oldest CreationTimestamp from Running/pending ingestion history recods.

gmechali:
Last PR had mistakenly kept the requirement for the SUCCESS row as well. This is no longer needed for the new path. We just rely on the min creationTimestamp for IngestionHistory rows with status PENDING/Running. Or, we fallback to 15 seconds ago.

### DIFF
--- a/internal/server/spanner/statements.go
+++ b/internal/server/spanner/statements.go
@@ -146,19 +146,11 @@ var statements = struct {
 			CompletionTimestamp DESC
 		LIMIT 1`,
 	getIngestionHistoryTimestamp: `		SELECT
-			CompletionTimestamp
+			MIN(CreationTimestamp)
 		FROM
 			IngestionHistory
 		WHERE
-			Status = 'SUCCESS'
-			AND CompletionTimestamp < (
-				SELECT COALESCE(MIN(CreationTimestamp), TIMESTAMP '9999-12-31T23:59:59Z')
-				FROM IngestionHistory
-				WHERE Status IN ('RUNNING', 'PENDING')
-			)
-		ORDER BY 
-			CompletionTimestamp DESC
-		LIMIT 1`,
+			Status IN ('RUNNING', 'PENDING')`,
 	getParam:  `= @%s`,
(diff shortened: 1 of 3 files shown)


internal/server/spanner/statements.go
```diff
@@ -146,19 +146,11 @@ var statements = struct {
...
-			Status = 'SUCCESS'
-			AND CompletionTimestamp < (
-				SELECT COALESCE(MIN(CreationTimestamp), TIMESTAMP '9999-12-31T23:59:59Z')
-				FROM IngestionHistory
-				WHERE Status IN ('RUNNING', 'PENDING')
-			)
-		ORDER BY 
-			CompletionTimestamp DESC
-		LIMIT 1`,
+			Status IN ('RUNNING', 'PENDING')`,
```
rohitkumarbhagat, line 153: workflow failed  at T1 -> new workflow started with fix -> new row created with Pending status at T2 -> mixer picks T2 -> failed changes between T1-T2 gets picked. Should this be avoided.

rohitkumarbhagat, line 153: Since this is hot query and history will accumulate overtime, consider creating a secondary index on status to avoid full table scan.

gmechali: If workflow failed at T1, it should not release the lock just yet, which means workflow 2 will not yet be allowed to acquire the lock to start. 

Updated in chat, but I think worfklow1 should trigger the reversion, run it within its workflow, and presumably will end up with status REVERTED, where it undoes the changes that workflow1 tried to do. 

I do think this query will need to get a little more complicated - pending Vishal's changes on reversion. But the crux of it - a workflow should only release the lock once it left the database in a good state.

If the reversion fails, then we'll need manual intervention.

gmechali: ack - will track it for follow up. thanks for the suggestion.

gmechali:
Actually closing this - the previous query was the right one.

PR #1974, opened 2026-07-10
Distinguish between timeouts/cancels in logging

n-h-diaz:
Follow up to https://github.com/datacommonsorg/mixer/pull/1971

Uses duration as an estimate of whether a canceled state was via proxy (API timed out, so log as error) vs client initiated (log as info)

All deadline exceeded will still be considered error

Updates the buffer to be 2 seconds

### DIFF
--- a/internal/server/spanner/query.go
+++ b/internal/server/spanner/query.go
@@ -62,8 +62,12 @@ const (
 	// Default fallback staleness duration when staleness timestamp is unavailable or expired.
 	defaultStalenessDuration = 15 * time.Second
 
-	// Default timeout for API requests.
+	// Default proxy timeout for API requests.
 	ApiTimeout = 60 * time.Second
+
+	// Buffer for proxy request timeout.
+	// This is to avoid race conditions and account for jitter in timing.
+	ApiTimeoutBuffer = 2 * time.Second
 
 	// Special edge predicates.
 	predAffectedPlace      = "affectedPlace"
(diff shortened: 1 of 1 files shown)


internal/server/spanner/query.go
```diff
@@ -1214,3 +1221,53 @@ func isTimeoutOrCanceled(err error) bool {
...
+
+	baseArgs := []any{
+		"reason", reason,
+		"duration", duration.String(),
+	}
+	allArgs := append(baseArgs, extraArgs...)
+
+	if reason == ClientCanceled {
+		slog.InfoContext(ctx, fmt.Sprintf("%s canceled by client.", action), allArgs...)
+	} else {
```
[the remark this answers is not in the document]
n-h-diaz: done thanks!

internal/server/spanner/query.go
```diff
@@ -1203,14 +1210,53 @@ func processCacheRows[T proto.Message](iter *spanner.RowIterator, newProto func(
...
+	}
+
+	baseArgs := []any{
+		"reason", reason,
+		"duration", duration.String(),
+	}
+	allArgs := append(baseArgs, extraArgs...)
+
+	if reason == ClientCanceled {
+		slog.InfoContext(ctx, fmt.Sprintf("%s canceled.", action), allArgs...)
```
nick-nlb: Thinking about our ability later to aggregate and query the logs: this creates a high cardinality of log entries (a wide variety of actions, each one creating a different error message). It looks like slog can take (and structure separately) key/value pairs. Perhaps the message could be fixed as something like "Spanner operation canceled", and then the action could follow as separate parameters `"action", action`

internal/server/spanner/query.go
```diff
@@ -1012,7 +1025,7 @@ func (sc *spannerDatabaseClient) executeQuery(
 	} else {
 		// Apply an absolute upper bound with a small buffer to protect the DB
 		// in case ESP is bypassed.
-		fallbackLimit := ApiTimeout + (5 * time.Second)
+		fallbackLimit := ApiTimeout + ApiTimeoutBuffer
```
nick-nlb, line 1030: This drops the buffer from 5 seconds to 2 (probably correct but just to verify that this is intentional).

nick-nlb (commented):
Thank you Natalie! A couple of comments (one actionable). Let me know what you think.

internal/server/spanner/query.go
n-h-diaz: yes intentional! I think 5 was probably too excessive

internal/server/spanner/query.go
n-h-diaz: good idea! I updated both to be "Spanner operation aborted." and have action as a separate key. The "reason" can distinguish between timeout and cancel

nick-nlb (approved):

PR #1979, opened 2026-07-13
Force InEdge for filter queries

n-h-diaz:
There was a regression in query plans for filter queries where the base table was being used, force a more stable query plan for filter queries, specifically using InEdge index.


nick-nlb (approved):
Thank you Natalie!

PR #1980, opened 2026-07-13
Point autopush mirrored traffic to staging db

n-h-diaz:
Jan db isn't really used anymore, and all of the relevant code for running staging db should also be in autopush.

Note this is still just mirrored traffic, but this is to help prep for eventually diverting traffic


deploy/featureflags/autopush_website.yaml
```diff
@@ -6,4 +6,5 @@ flags:
   V3MirrorFraction: 1.0
   WriteUsageLogs: 1.0
   UseSpannerGraph: true
-  SpannerGraphDatabase: dc_graph_2026_01_27
+  SpannerGraphDatabase: projects/datcom-store/instances/dc-graph-staging/databases/dc_graph
+  UseMultiEntitySchema: true
```
rohitkumarbhagat: can we also enable sdmx on autopush mixer? EnableSDMXDataApi: true

rohitkumarbhagat (approved):

deploy/featureflags/autopush_website.yaml
n-h-diaz: done, thanks for the review!

PR #1985, opened 2026-07-14
update the mixer to use NL SVs and also updating staging to default to use Spanner NL embeddings

shixiao-coder:


keyurva (approved):
Thanks!

deploy/featureflags/staging.yaml
keyurva, line 1: Given that this is the staging yaml, the PR description should say staging correct?

shixiao-coder: yep, updated to staging

PR #1988, opened 2026-07-15
Support facetId as filterable series attribute in sdmx

rohitkumarbhagat:
## Summary

  - Return facetId as a series-level attribute in SDMX data responses.
  - Allow facetId filtering through REST and dispatcher data requests using the stored facet_id column.
  - Keep other attributes blocked unless explicitly added to the filterable-attribute allowlist.
  - Require SDMX data queries to specify variableMeasured and at least one observation property.
  - Keep SDMX availability filtering unchanged.

  ## Impacted APIs

  - V3SdmxData
  - /sdmx/v3/data


internal/server/spanner/query_multientity.go
```diff
@@ -694,6 +739,13 @@ func populateSdmxFacetComponents(series *sdmxpb.SdmxTimeSeries, facets map[strin
 	}
 }
 
+func populateSdmxFacetID(series *sdmxpb.SdmxTimeSeries, facetID string) {
+	if series.Attributes == nil {
+		series.Attributes = map[string]string{}
+	}
+	series.Attributes[datacommons.ComponentFacetID] = facetID
+}
```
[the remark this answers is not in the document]
rohitkumarbhagat: done

internal/server/sdmx/rest/v2/component.go
```diff
@@ -126,15 +126,15 @@ func validateAvailabilityRequest(path AvailabilityPath, constraints map[string][
 		return status.Errorf(codes.Unimplemented, "unsupported SDMX availability component %q", path.ComponentID)
 	}
 
-	return validateComponentFilters(constraints)
+	return validateComponentFilters(constraints, isFilterableDimensionCandidate)
 }
 
-func validateComponentFilters(constraints map[string][]string) error {
+func validateComponentFilters(constraints map[string][]string, isFilterable func(string) bool) error {
```
keyurva, line 132: Is there a reason for `isFilterable` to be a closure vs just a standard function that gets called?

internal/server/spanner/query_multientity.go
```diff
@@ -627,25 +636,56 @@ func validateSdmxConstraintValues(constraints map[string]*sdmxpb.ConstraintList)
...
 }
 
-func validateSdmxConstraintComponents(
-	constraints map[string]*sdmxpb.ConstraintList,
+func validateSdmxDataConstraintComponents(
+	constraints map[string]*sdmxpb.SdmxComponentConstraint,
 	shape *sdmxpb.SdmxDataShape,
 ) error {
-	filterableDimensions := sdmxFilterableDimensions(shape)
+	return validateSdmxConstraintComponents(constraints, sdmxFilterableDataComponents(shape), "components")
```
keyurva, line 643: nit: consider using consts here and elsewhere

keyurva (approved):
Thanks!

internal/server/sdmx/rest/v2/component.go
rohitkumarbhagat: Good question. The parameter is a function value and can accept a closure, though the current call sites pass named functions. We need the parameter because data and availability use different filtering policies: data queries allow dimensions plus filterable attributes such as facetId, while availability allows dimensions only. Passing the predicate keeps the common validation loop shared without introducing a mode flag or duplicating it

internal/server/spanner/query_multientity.go
rohitkumarbhagat: Thanks for the suggestion. I checked the existing component-kind constants, but these two values are human-readable error labels rather than component IDs or kinds: the data set includes both dimensions and attributes, while the availability set includes dimensions only. Reusing ComponentKindDimension would also require deriving the plural form. Since the labels are local presentation text and do not represent an external value, I suggest keeping the literals here.

rohitkumarbhagat:
Thanks Keyur!!

PR #1995, opened 2026-07-15
This change update the resolve logic to default use genai client

shixiao-coder:
We discovered a latency for calling embedding model from spanner client remote model call into vertex ai model.

This PR uses genai client by default to invoke embeddings


internal/server/spanner/datasource.go
```diff
@@ -556,14 +579,35 @@ func (sds *SpannerDataSource) vectorSearchResolution(
...
+				)
+				if err != nil {
+					return status.Errorf(codes.Internal, "failed to get term embedding for %s via GenAI SDK: %v", node, err)
+				}
+				if len(res.Embeddings) > 0 {
+					embeddings = make([]float64, len(res.Embeddings[0].Values))
+					for idx, val := range res.Embeddings[0].Values {
+						embeddings[idx] = float64(val)
+					}
+				}
```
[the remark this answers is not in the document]
shixiao-coder: genai result doesn't accept get method

internal/server/spanner/embedding_config.go
```diff
@@ -54,8 +54,9 @@ const (
 // SearchConfig holds the configuration for the parameters necessary to search.
 type SearchConfig struct {
 	SearchAlgorithm  SearchMethod     `json:"search_algorithm" yaml:"search_algorithm"`
-	EmbeddingModel   string           `json:"embedding_model" yaml:"embedding_model"` // the model name registered in spanner to invoke
-	EmbeddingTable   string           `json:"embedding_table" yaml:"embedding_table"` // the table name in spanner for embeddings
+	EmbeddingModel         string           `json:"embedding_model" yaml:"embedding_model"` // the model name registered in spanner to invoke
+	EmbeddingModelEndpoint string           `json:"embedding_model_endpoint" yaml:"embedding_model_endpoint"`
```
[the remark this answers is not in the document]
shixiao-coder: added

internal/server/spanner/datasource.go
```diff
@@ -556,14 +579,35 @@ func (sds *SpannerDataSource) vectorSearchResolution(
...
-				errCtx,
-				cfg.SearchConfig.EmbeddingModel,
-				node,
-				string(cfg.SearchConfig.QueryTaskType),
-			)
-			if err != nil {
-				return status.Errorf(codes.Internal, "failed to get term embedding for %s: %v", node, err)
+			var embeddings []float64
+			var err error
+			if sds.genaiClient != nil {
```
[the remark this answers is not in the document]
shixiao-coder: added the alert

internal/server/spanner/datasource.go
```diff
@@ -91,6 +94,26 @@ func NewSpannerDataSource(
...
+	if client != nil {
+		dbURI := client.Id()
+		project, _, _, err := parseDatabaseURI(dbURI)
+		if err == nil && project != "" {
+			ctx := context.Background()
+			genaiClient, err := genai.NewClient(ctx, &genai.ClientConfig{
+				Project:  project,
+				Location: "us-central1",
+				Backend:  genai.BackendVertexAI,
+			})
```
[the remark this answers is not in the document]
shixiao-coder: updated the location to read from Yaml

internal/server/spanner/datasource.go
```diff
@@ -63,6 +65,7 @@ type SpannerDataSource struct {
 	mapsClient      internalmaps.MapsClient
 	searchConfig    *SpannerSearchConfig
 	topicExpander   resolvev2.TopicExpander
+	genaiClient     *genai.Client
```
keyurva: Currently this creates a tight coupling between our spanner client and the genai client.

WDYT about abstracting embedding generation into a thin interface, say:

```golang
// Embedder generates vector embeddings for input text.
type Embedder interface {
    Embed(ctx context.Context, text string) ([]float64, error)
}
```

Then define a `GenAIEmbedder` struct that implements `Embedder` using `*genai.Client`.
`SpannerDataSource` accepts `Embedder` in constructor/options remaining decoupled from genai client and making it easier to test the embedder.

shixiao-coder: agreed, I think it's great move it to an interface. I'm moving it further into an independant folder like the map client to be fully decoupled from spanner

internal/embedder/embedder.go
```diff
@@ -0,0 +1,91 @@
...
+		for idx, val := range res.Embeddings[0].Values {
+			embeddings[idx] = float64(val)
+		}
+		return embeddings, nil
+	}
+	return nil, nil
+}
+
+// NewEmbedder creates a new Embedder instance wrapping Google GenAI SDK client.
+func NewEmbedder(ctx context.Context, dbURI string, location ...string) (Embedder, error) {
```
keyurva: Consider accept a `project` string itself. Otherwise it gets coupled with spanner specific logic to extract the project from the DB URI.

Also, consider making `location` just a string. You can default to default location if the string is empty.

internal/embedder/embedder.go
```diff
@@ -0,0 +1,91 @@
...
+	res, err := e.client.Models.EmbedContent(
+		ctx,
+		endpoint,
+		genai.Text(text),
+		&genai.EmbedContentConfig{TaskType: taskType},
+	)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get term embedding for %s via GenAI SDK: %v", text, err)
+	}
+	if len(res.Embeddings) > 0 {
```
keyurva: Consider adding a check for `res.Embeddings[0] != nil` to prevent potential NPEs when accessing `es.Embeddings[0].Values`.

keyurva (approved):
Thanks Xiao - very cool optimization!

internal/embedder/embedder.go
shixiao-coder: Sure added the project directly and use HostProject from main
also updated the location to a simple string and default to us-central1 if empty

internal/embedder/embedder.go
shixiao-coder: Sure, added a check

PR #1996, opened 2026-07-16
Trigger topic cache sync on Spanner ingestion completions

clincoln8:
## Goal
Implements event-driven reactive in-memory cache synchronization (topic hierarchy and agent service) in `datcom-mixer` by hooking into the Spanner `IngestionHistory` timestamp poller.

## Key Aspects of the Approach
* **Spanner Timestamp Trigger (`query.go`)**: Detects DB initialization transitions (`wasInitialized == false -> true`) and timestamp jumps (`prevNano != newNano`) to trigger registered `SetOnIngestionUpdate` callbacks.
* **Thread-Safe Drain Loop (`query.go`)**: Isolates in-memory cache reloading into `maybeTriggerReload` and `runIngestionReloadLoop`. Uses `sc.hookMutex` and atomic state flags to drain burst updates safely without lost updates or concurrent goroutine storms.
* **Redis Bypass for Fresh In-Memory Loading (`topic_cache.go` & `cmd/main.go`)**: Adds `ReloadHierarchy` to bypass stale L2 Redis cache reads during ingestion updates, loading fresh topic data directly into L1 in-memory cache and re-warming L2 Redis.
* **Fail-Safe In-Memory Preservation (`topic_cache.go`)**: Retains existing populated in-memory topic hierarchy if Spanner transiently returns zero topics.

## Edge Cases, Errors & Logging
* Gracefully ignores table-not-found on uninitialized DB boot.
* Deferred panic recovery resets reloading flags safely if callbacks fail.
* 2-minute context timeout bounds each reactive reload execution.

## Testing & Verification
* **Unit Tests**: Pass cleanly across `internal/server/...`, including concurrent update draining (`TestSetOnIngestionUpdate_DrainsSubsequentUpdate`).
* **E2E Verification**: Tested against live Spanner database (`test-latest-dcp-dc-db`), reloading topic graphs in memory within ~30s of `datacommons admin ingest` completion without pod restart.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -572,9 +572,14 @@ func main() {
 	// Bind reactive ingestion callback and start background Spanner polling ONLY after server components are initialized
 	if spannerClient != nil {
 		spannerClient.SetOnIngestionUpdate(func(ctx context.Context) {
-			slog.Info("Executing reactive periodic hooks triggered by Spanner ingestion state change")
-			if err := mixerServer.RunPeriodicHooks(ctx); err != nil {
-				slog.Error("Reactive RunPeriodicHooks failed during Spanner sync", "error", err)
+			slog.Info("Executing reactive cache reloads triggered by Spanner ingestion state change")
+			if topicCacheManager != nil {
+				if _, err := topicCacheManager.LoadHierarchy(ctx); err != nil && !spanner.IsTableNotFoundError(err) {
+					slog.Error("Failed to reload topic cache hierarchy on Spanner ingestion update", "error", err)
+				}
+			}
+			if agentSvc := mixerServer.AgentService(); agentSvc != nil {
+				agentSvc.Reset()
 			}
 		})
 		spannerClient.Start()
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -571,17 +571,20 @@ func main() {
 
 	// Bind reactive ingestion callback and start background Spanner polling ONLY after server components are initialized
 	if spannerClient != nil {
-		spannerClient.SetOnIngestionUpdate(func(ctx context.Context) {
-			slog.Info("Executing reactive cache reloads triggered by Spanner ingestion state change")
-			if topicCacheManager != nil {
-				if _, err := topicCacheManager.LoadHierarchy(ctx); err != nil && !spanner.IsTableNotFoundError(err) {
-					slog.Error("Failed to reload topic cache hierarchy on Spanner ingestion update", "error", err)
+		if flags.EnableReactiveInMemoryCacheRefresh {
+			slog.Info("Enabling reactive in-memory cache reloads on Spanner ingestion state changes")
+			spannerClient.SetOnIngestionUpdate(func(ctx context.Context) {
+				slog.Info("Executing reactive in-memory cache reloads triggered by Spanner ingestion state change")
+				if topicCacheManager != nil {
+					if _, err := topicCacheManager.ReloadHierarchy(ctx); err != nil && !spanner.IsTableNotFoundError(err) {
+						slog.Error("Failed to reload topic cache hierarchy on Spanner ingestion update", "error", err)
+					}
 				}
-			}
-			if agentSvc := mixerServer.AgentService(); agentSvc != nil {
-				agentSvc.Reset()
-			}
-		})
+				if agentSvc := mixerServer.AgentService(); agentSvc != nil {
+					agentSvc.Reset()
+				}
+			})
+		}
 		spannerClient.Start()
 	}
 
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -235,7 +235,6 @@ func main() {
 			slog.Error("Failed to create Spanner client", "error", err)
 			os.Exit(1)
 		}
-		spannerClient.Start()
 		defer spannerClient.Close()
 	}
 	slog.Info("After Spanner client creation")
@@ -568,6 +567,17 @@ func main() {
 	if err := mixerServer.RunInitHooks(ctx); err != nil {
 		slog.Error("Failed to initialize server components during startup", "error", err)
 		os.Exit(1)
+	}
+
+	// Bind reactive ingestion callback and start background Spanner polling ONLY after server components are initialized
+	if spannerClient != nil {
+		spannerClient.SetOnIngestionUpdate(func(ctx context.Context) {
+			slog.Info("Executing reactive periodic hooks triggered by Spanner ingestion state change")
+			if err := mixerServer.RunPeriodicHooks(ctx); err != nil {
+				slog.Error("Reactive RunPeriodicHooks failed during Spanner sync", "error", err)
+			}
+		})
+		spannerClient.Start()
 	}
 
 	// Start the periodic scheduler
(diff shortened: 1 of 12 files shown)


cmd/main.go
```diff
@@ -533,6 +532,17 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Bind reactive ingestion callback and start background Spanner polling ONLY after server components are initialized
+	if spannerClient != nil {
+		spannerClient.SetOnIngestionUpdate(func(ctx context.Context) {
```
gmechali: Instead of maknig it part of the reactive calling, could we push to the topic cache when it's time to upadte?

https://github.com/datacommonsorg/mixer/blob/b61a906f82ae0fb67ab17674331fec4bb865265f/internal/server/spanner/staleness_tracker.go#L60

This staleness tracker now logs here that we've noticed a change in the staleness timestamp, which means our spanner requests have now moved up and we should read new data. 

Could we just connect that part to the TopicCacheManager and trigger the reload?

clincoln8: I'm not totally sure what you mean here. I've updated the code a bit, do you mind taking another look and see if it's aligned with this?

cmd/main.go
```diff
@@ -560,6 +559,22 @@ func main() {
...
+			slog.Info("Executing reactive cache reloads triggered by Spanner ingestion state change")
+			if topicCacheManager != nil {
```
keyurva: If `m.redisClient != nil`, `LoadHierarchy(ctx)` checks the L2 Redis cache first (`GetCachedResponse`). If Redis has the pre-ingestion hierarchy cached, `LoadHierarchy` will hit Redis and return the stale hierarchy without fetching newly ingested topics from Spanner (KG).

Should `TopicCacheManager` provide a `ReloadHierarchy(ctx)` method (or a bypass parameter) that forces a fresh fetch from KG via `FetchTopicsFromKG(ctx)` and overwrites the L2 Redis cache?

internal/server/topic/topic_cache.go
```diff
@@ -243,6 +243,11 @@ func (m *TopicCacheManager) LoadHierarchy(ctx context.Context) (*pb.TopicHierarc
 	}
 	if len(hierarchy.GetTopics()) == 0 {
 		slog.Warn("Loaded empty topic hierarchy.")
+		// Fail-safe preservation: never overwrite an existing functional L1 cache with an empty structure
+		if existing := m.CachedHierarchy(); len(existing.GetTopics()) > 0 {
```
keyurva, line 247: Good fail-safe guard to prevent an empty yield from overwriting a valid cache.

Just a quick clarification: if topics were ever legitimately cleared in Spanner, this guard would retain the old cache until pod restart. Is this trade-off intended for production safety?

internal/server/spanner/timestamp_test.go
```diff
@@ -170,3 +170,65 @@ func TestTimestampResetOnNull(t *testing.T) {
 		t.Fatalf("Expected error when staleness timestamp is zero, but got nil")
 	}
 }
+
+func TestSetOnIngestionUpdate_TriggersOnTimestampChange(t *testing.T) {
```
keyurva, line 174: Good unit tests for the basic trigger and no-op paths!

Could we also add a unit test verifying the atomic drain loop when a second timestamp update arrives while `hook` is actively executing?

cmd/main.go
```diff
@@ -560,6 +559,22 @@ func main() {
...
+				if _, err := topicCacheManager.LoadHierarchy(ctx); err != nil && !spanner.IsTableNotFoundError(err) {
```
keyurva: any reason table not found is not logged as an error?

keyurva: `LoadHierarchy` could load from redis which itself might be stale. I think you mentioned yesterday that redis is cleared on import but wanted to ensure that is the case.

keyurva: Also, is this the case for base DC as well?

internal/server/spanner/query.go
keyurva, line 993: This method now seems to be handling a lot of low level responsibilities. I had a session with jetski and sent you a simplification proposal. Use it if you like it otherwise ok to go as-is.

internal/server/topic/topic_cache.go
keyurva, line 247: If topics were legitimately cleared in spanner, this will still retain the old cache. Just checking if this is an intentional trade-off?

keyurva (approved):
Thanks!

cmd/main.go
gmechali: I think this is fine too. You're basically doing the same thing but in a different place I had in mind, but Im not sure my suggestion was better anyways

gmechali (dismissed):

gmechali:
Thanks Christie for resolving this so quick!

internal/server/spanner/query.go
clincoln8: Thanks! I've refactored processStalenessTimestamp into focused helper methods (maybeTriggerReload and runIngestionReloadLoop) as proposed, while keeping sc.hookMutex around the drain loop check and state transitions to prevent lost-update race conditions.

cmd/main.go
clincoln8: Great catch! I've added ReloadHierarchy(ctx) to TopicCacheManager, which bypasses L2 Redis reads, fetches fresh topic data directly from Spanner/KG, and updates both L1 memory and L2 Redis. SetOnIngestionUpdate in cmd/main.go now calls ReloadHierarchy.

internal/server/spanner/timestamp_test.go
clincoln8: added TestSetOnIngestionUpdate_DrainsSubsequentUpdate in timestamp_test.go to verify that a second timestamp update arriving mid-execution is drained and executed by the active loop.

internal/server/topic/topic_cache.go
clincoln8: according to jetski, "yes, in Data Commons an empty topic yield is almost always due to transient DB connection issues or table bootstrapping. Retaining the populated in-memory cache until pod restart prevents serving broken {} responses during transient outages."

cmd/main.go
clincoln8: !spanner.IsTableNotFoundError(err) is ignored on initial boot because Spanner tables may not exist yet during fresh database startup; the timestamp poller logs a fallback warning when missing.

clincoln8: updated to use ReloadHierarchy to bypass redis anyways. i'm not sure if redis is cleared on each ingest for baseDC

cmd/main.go
```diff
@@ -570,6 +569,25 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Bind reactive ingestion callback and start background Spanner polling ONLY after server components are initialized
+	if spannerClient != nil {
+		if flags.EnableReactiveInMemoryCacheRefresh {
```
keyurva, line 574: One downside to this flag is that DCP and base DC behavior will be different. But jetski thinks this is ok. One recommendation it had is to default it to true (or simply remove it) once this has proven to be stable in DCP instances.

keyurva (dismissed):
This was tricky - thanks for the multiple iterations and getting it done!

cmd/main.go
keyurva: Should we log that condition? Might be useful for debugging.

PR #2005, opened 2026-07-17
fix(spanner): Reset staleness timestamp when IngestionHistory returns NULL

clincoln8:
### Problem
When `IngestionHistory` yields `NULL` (`nullTime.Valid == false`) upon the completion of a successful ingestion workflow, `fetchAndUpdateTimestamp` prints out a warning indicating fallback to exact staleness reads (`15s`) and immediately exits (`return nil`).

However, during Spanner query evaluation inside `executeQuery`, exact staleness fallback reads are only triggered when `getStalenessTimestamp()` returns an error—which occurs strictly when `sc.timestamp` equals `0`:

```go
func (sc *spannerDatabaseClient) getStalenessTimestamp() (time.Time, error) {
    val := sc.timestamp.Load()
    if val != 0 {
        return time.Unix(0, val).UTC(), nil
    }
    slog.Warn("Spanner staleness timestamp not available. Falling back to default staleness.")
    return time.Time{}, fmt.Errorf("error getting staleness timestamp") # <-- returns err ONLY WHEN TIMESTAMP==0
}
// Inside executeQuery:
ts, err := sc.getStalenessTimestamp()
if err != nil {
    return runQuery(spanner.ExactStaleness(defaultStalenessDuration)) # <-- on err, return default 15 sec
}
return runQuery(spanner.ReadTimestamp(ts))
```

Because `sc.timestamp` is left completely untouched across memory (`val != 0`) upon receiving `NULL` from Spanner, `getStalenessTimestamp()` never triggers the exact staleness fallback and permanently locks queries to historical pre-ingestion database snapshots.

### Manifestation Scenario
1. An ingestion workflow (`datacommons admin ingest start`) runs. Mixer polls Spanner and caches the active workflow timestamp right inside `sc.timestamp`.
2. The workflow completes cleanly (`Status = 'SUCCESS'`). Subsequent Spanner polling iterations yield `NULL`.
3. Because `sc.timestamp` is never zeroed across memory right right when hitting `NULL`, `getStalenessTimestamp()` keeps returning the cached start time, trapping API reads in outdated pre-ingestion states until the container restarts.

### Solution
Explicitly reset `sc.timestamp.Store(0)` inside the `!nullTime.Valid` check when a non-zero snapshot exists (`sc.timestamp.Load() != 0`). This allows `getStalenessTimestamp()` to encounter zero and cleanly invoke exact staleness fallback reads across post-ingestion windows.


gmechali (approved):
Thanks you for catching this Christie! Sorry :)

PR #2006, opened 2026-07-17
  Support remote containedInPlace+ expansion in SDMX data queries

rohitkumarbhagat:
## Summary

  - Fetch remote place hierarchies for SDMX containedInPlace+ filters and combine them with locally available containment data.
  - Continue serving SDMX observations from local sources; remote configuration is used only for place expansion.
  - Add a configurable expansion limit, bounded concurrency, relation deduplication, and local-only fallback for remote failures.
  - Let Spanner choose join order while preserving existing APPLY_JOIN and index hints.

  ## Testing

  - Added dispatcher and feature-flag unit coverage.
  - Added query-builder golden tests.
  - Added Spanner emulator coverage for remote-only results, local and remote unions, multiple components, shared relations, entity slots, and State containment.

  ## Impacted API

  - GET /sdmx/v3/data[/{tail}] (V3SdmxData)

  SDMX Availability is unchanged. There are no request or response schema changes.


internal/server/dispatcher/relation_expression_processor.go
```diff
@@ -81,6 +104,165 @@ func (p *RelationExpressionProcessor) PreProcess(rc *RequestContext) (Outcome, e
...
+					return limitErr
+				}
+				slog.Warn("RelationExpressionProcessor: falling back to local-only SDMX containment expansion",
+					"ancestor", relation.Ancestor,
+					"childPlaceType", relation.ChildPlaceType,
+					"duration", time.Since(start),
+					"error", err,
+				)
+				return nil
+			}
```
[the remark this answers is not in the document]
rohitkumarbhagat: this is as per existing behavior. will add a todo to fix later.

internal/server/dispatcher/relation_expression_processor.go
```diff
@@ -81,6 +104,168 @@ func (p *RelationExpressionProcessor) PreProcess(rc *RequestContext) (Outcome, e
...
+}
+
+type sdmxRemoteExpansionSource struct {
+	datasource.DataSource
+	relation datacommons.ContainedInPlaceConstraint
+	limit    int
+	dcids    map[string]struct{}
+}
+
+func (s *sdmxRemoteExpansionSource) Node(
```
keyurva, line 230: Since this method mutates the shared `dcids` list and looks up it's length, consider guarding it with a mutex:

```golang
type sdmxRemoteExpansionSource struct {
	...
	mu       sync.Mutex
}

func (s *sdmxRemoteExpansionSource) Node(
	ctx context.Context,
	req *pbv2.NodeRequest,
	pageSize int,
) (*pbv2.NodeResponse, error) {
	resp, err := s.DataSource.Node(ctx, req, pageSize)
	if err != nil {
		return nil, err
	}
	s.mu.Lock()
	defer s.mu.Unlock()
    ....
}
```

internal/server/dispatcher/relation_expression_processor.go
```diff
@@ -81,6 +104,168 @@ func (p *RelationExpressionProcessor) PreProcess(rc *RequestContext) (Outcome, e
 	return Continue, nil
 }
 
+func (p *RelationExpressionProcessor) preProcessSdmxData(rc *RequestContext) (Outcome, error) {
```
keyurva, line 110: nit: As a general best practice, consider breaking up longer methods using helpers wherever it makes sense.

internal/server/dispatcher/relation_expression_processor.go
```diff
@@ -81,6 +104,168 @@ func (p *RelationExpressionProcessor) PreProcess(rc *RequestContext) (Outcome, e
...
+						"ancestor", relation.Ancestor,
+						"childPlaceType", relation.ChildPlaceType,
+						"limit", limitErr.limit,
+					)
+					return limitErr
+				}
+				// TODO: Propagate context cancellation and deadline errors instead of
+				// falling back to local-only results when the shared Observation
+				// relation-expansion behavior is updated.
+				slog.Warn("RelationExpressionProcessor: falling back to local-only SDMX containment expansion",
```
keyurva, line 167: Any reason we're not propagating it right now?

internal/server/remote/datasource.go
```diff
@@ -46,6 +46,8 @@ func (rds *RemoteDataSource) Id() string {
 
 func (rds *RemoteDataSource) Node(ctx context.Context, req *pbv2.NodeRequest, pageSize int) (*pbv2.NodeResponse, error) {
 	// The remote datasource currently calls V2 node, which does not use custom pageSize.
+	// TODO: Propagate ctx through RemoteClient and FetchRemote and configure a
+	// bounded HTTP timeout so remote requests honor cancellation and deadlines.
 	return rds.client.Node(req)
```
keyurva, line 51: Good callout. Let's plan to do this in a follow-up. Will be useful for other remote calls as well.

internal/server/spanner/query_builder_multientity.go
```diff
@@ -484,56 +487,89 @@ func (b *multiEntityQueryBuilder) getSdmxContainedInPlaceObservationsQuery(
...
 	}
 
+	// The anchor is the containment-constrained entity used to drive the
+	// contained_places-to-TimeSeries join. Its entity slot selects the access
+	// path; remaining containment constraints are applied as semi-join filters.
 	anchor := resolved[0]
 	index := "_BASE_TABLE"
-	statementHint := ""
+	// Production containment queries favor total scan and join throughput.
+	statementHints := []string{"SCAN_METHOD=COLUMNAR", "EXECUTION_METHOD=BATCH"}
```
keyurva, line 542: Is there a way to add these hints in the statements file itself vs here? I worry about maintainability when it comes to manipulating queries and hints like this. OK to do in a follow-up.

keyurva (approved):
Thanks!

internal/server/dispatcher/relation_expression_processor.go
rohitkumarbhagat: I deferred this to keep behavior aligned with the existing Observation path. More importantly, RemoteDataSource.Node currently does not propagate the context through RemoteClient, so handling cancellation only here would be partial. The follow-up should first propagate context and configure a bounded HTTP timeout, then propagate cancellation and deadline errors consistently from both Observation and SDMX expansion.

internal/server/remote/datasource.go
rohitkumarbhagat: sure

internal/server/dispatcher/relation_expression_processor.go
rohitkumarbhagat: Agreed that this should be revisited as the flow grows. I added a TODO to break the orchestration into reusable helpers when SDMX Availability adopts remote containment, so we can extract the shared flow once rather than introduce single-use helpers now.

internal/server/dispatcher/relation_expression_processor.go
rohitkumarbhagat: Thanks for pointing this out. The current request contains one ancestor and NodeFetchAll paginates it sequentially, so the accumulator is not accessed concurrently today. However, that relies on an internal implementation detail that could change. I added the mutex preventively so this wrapper remains safe if pagination becomes concurrent in the future. The remote call stays outside the critical section, so the current cost is limited to locking around the small in-memory accumulation and limit check.

internal/server/spanner/query_builder_multientity.go
rohitkumarbhagat: Agreed. I’m deferring this because the stacked time-period change adds explicit-date and latest containment templates in the same builder and statements files. I added a TODO here and will consolidate hint selection across all containment variants after that change lands

rohitkumarbhagat:
Thanks Keyur!!

PR #2019, opened 2026-07-21
Prevent observation OOM when variables are not specified

rohitkumarbhagat:
## Summary

  Prevent Spanner observation requests with an empty variable list from fetching complete observation histories for every variable associated with an entity.

  The multi-entity query previously interpreted an empty variable list as all variables. When date or value fields were selected, this produced an unbounded observation query and could cause Mixer to run out of memory.

  This change matches the legacy Bigtable behavior:

  - Requests without date or value fields use the variable-existence path.
  - Facet-only requests without variables return the variables available for the entity.
  - Date or value requests without variables return an empty response without calling `GetObservations`.
  - Requests with explicit variables continue using the existing observation and facet paths.

  ## Testing

  - `go test ./internal/server/spanner/... -count=1`
  - `./run_test.sh -l`
  - `git diff --check`

### DIFF
--- a/internal/server/spanner/datasource.go
+++ b/internal/server/spanner/datasource.go
@@ -175,21 +175,24 @@ func (sds *SpannerDataSource) Observation(ctx context.Context, req *pbv2.Observa
 		return nil, fmt.Errorf("variable must be specified for entity.expression")
 	}
 
-	var observations []*Observation
-	var err error
-
 	qo := selectFieldsToQueryOptions(req.Select)
 
-	// Check if this is an existence-only request:
-	// 1. Only 'variable' and 'entity' are requested (no date, value, or facet).
-	// 2. A simple list of entities is provided (no complex entity expression).
-	isExistenceRequest := !qo.date && !qo.value && !qo.facet && len(entities) > 0 && entityExpr == ""
+	// A request without date/value is existence-only unless facets were
+	// requested for explicitly specified variables.
+	isExistenceRequest := entityExpr == "" &&
+		!qo.date &&
+		!qo.value &&
+		(!qo.facet || len(variables) == 0)
 
-	if isExistenceRequest {
+	switch {
+	case isExistenceRequest:
 		return sds.handleExistenceRequest(ctx, req, variables, entities)
+	case len(variables) == 0:
+		// Match legacy behavior: observation fields without variables return no data.
+		return &pbv2.ObservationResponse{}, nil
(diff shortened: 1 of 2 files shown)


internal/server/spanner/datasource.go
```diff
@@ -175,21 +175,24 @@ func (sds *SpannerDataSource) Observation(ctx context.Context, req *pbv2.Observa
...
+	isExistenceRequest := entityExpr == "" &&
+		!qo.date &&
+		!qo.value &&
+		(!qo.facet || len(variables) == 0)
 
-	if isExistenceRequest {
+	switch {
+	case isExistenceRequest:
 		return sds.handleExistenceRequest(ctx, req, variables, entities)
+	case len(variables) == 0:
```
keyurva, line 190: I think the `isExistenceRequest` check about covers it but can we ensure that existence requests (with entity dcids and no variables) and facet requests continue to work?

keyurva (approved):
Thanks for the super fast fix!

internal/server/spanner/datasource.go
```diff
@@ -175,21 +175,24 @@ func (sds *SpannerDataSource) Observation(ctx context.Context, req *pbv2.Observa
...
-	// Check if this is an existence-only request:
-	// 1. Only 'variable' and 'entity' are requested (no date, value, or facet).
-	// 2. A simple list of entities is provided (no complex entity expression).
-	isExistenceRequest := !qo.date && !qo.value && !qo.facet && len(entities) > 0 && entityExpr == ""
+	// A request without date/value is existence-only unless facets were
+	// requested for explicitly specified variables.
+	isExistenceRequest := entityExpr == "" &&
+		!qo.date &&
+		!qo.value &&
+		(!qo.facet || len(variables) == 0)
```
shixiao-coder, line 185: Is it OK to remove len(entities) > 0 condition?

shixiao-coder (approved):
LGTM with a comment to resolve

internal/server/spanner/datasource.go
keyurva: Let's do this in a followup.

internal/server/spanner/datasource.go
keyurva: Let's do this separately.

keyurva:
Submitting this so folks can test during office hours in the US. Can revert / update if needed.

PR #2030, opened 2026-07-22
Avoid duplicate counts and rows across import-scoped provenances

SandeepTuniki:
(Raising this PR for quick feedback as I don't know full the scope of changes necessary for this.)

This PR makes the changes necessary to accommodate the change which updates the provenance for generated nodes and edges from `dc/base/GeneratedGraphs` to `dc/base/generated/<import_name>`

### DIFF
--- a/internal/sqldb/statements.go
+++ b/internal/sqldb/statements.go
@@ -256,7 +256,7 @@ var statements = struct {
 		WHERE lookup_key = :key;
 	`,
 	getAllStatVarGroups: `
-		SELECT DISTINCT t1.subject_id svg_id, t2.object_value svg_name, t3.object_id svg_parent_id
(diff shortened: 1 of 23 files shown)


internal/sqldb/statements.go
```diff
@@ -256,7 +256,7 @@ var statements = struct {
 		WHERE lookup_key = :key;
 	`,
 	getAllStatVarGroups: `
-		SELECT t1.subject_id svg_id, t2.object_value svg_name, t3.object_id svg_parent_id
+		SELECT DISTINCT t1.subject_id svg_id, t2.object_value svg_name, t3.object_id svg_parent_id
```
n-h-diaz: I don't think we'd need to modify any of these

n-h-diaz (approved):
Thank you!

PR #2031, opened 2026-07-22
update the logic to read 2 * limit from vector search and dedup in go and filter down to limit

shixiao-coder:


keyurva (approved):

internal/server/spanner/datasource.go
```diff
@@ -580,11 +580,15 @@ func (sds *SpannerDataSource) vectorSearchResolution(
...
+				fetchLimit = cfg.SearchConfig.Limit
+			}
 			searchResults, err := sds.client.VectorSearchQuery(
 				errCtx,
 				cfg.SearchConfig.EmbeddingTable,
-				cfg.SearchConfig.Limit,
+				fetchLimit,
 				embeddings,
 				cfg.SearchConfig.NumLeaves,
 				cfg.SearchConfig.Threshold,
```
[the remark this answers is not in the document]
shixiao-coder: updated

internal/server/spanner/datasource.go
```diff
@@ -580,11 +580,15 @@ func (sds *SpannerDataSource) vectorSearchResolution(
 				return nil
 			}
 
-			// 2. Vector search
+			// 2. Vector search (fetch 2x limit to ensure unique candidates count reaches limit after deduplication)
+			fetchLimit := cfg.SearchConfig.Limit * 2
```
clincoln8, line 584: Could we double the limit and dedupe directly in the spanner query? Or is it more efficient to return more rows and dedupe in Go?

shixiao-coder: dedup in spanner would require a group by and join, this is highly likely reducing effeciency and less preferred than a quick linear scan in go

keyurva: This maybe worth exploring (can be done later though). 

Since we're only interested in the top scored row for a given dcid, is there an efficient way to do this in sql itself. Is this the sql that is run for vector search? https://github.com/datacommonsorg/mixer/blob/7b88c5d045daf42ccf5f2cc12969e453315e0904/internal/server/spanner/golden/query_builder/vector_search_node.sql#L14

PR #2044, opened 2026-07-23
Optimize multi-type direct `containedInPlace` Node queries

rohitkumarbhagat:
## Summary

- Extend Node containment planning to direct inbound `containedInPlace` queries.
- For queries with exactly one `typeOf` filter:
  - Two or more distinct type values use ancestor-first execution.
  - A single type value retains the existing unforced query.
- Preserve the configured behavior for `linkedContainedInPlace` and the generic fallback for other graph queries.
- Retain the existing `FORCE_INDEX=InEdge` hint.

The ancestor-first path emits `FORCE_JOIN_ORDER=TRUE` and `SCAN_METHOD=COLUMNAR`.

## Motivation

We receive frequent requests for direct children of small places with a broad list of place types. When such a request returns no results, Spanner may start from the global `typeOf` side, scan millions of edges, and then perform containment lookups for every candidate.

For example:

`geoId/5508971700<-containedInPlace{typeOf:[EurostatNUTS3,...,State]}`

The saved plans used the same optimizer version and statistics package and both returned zero rows:

| Metric | Unforced | Forced ancestor-first |
| --- | ---: | ---: |
| Elapsed time | 9.66 s | 118.54 ms |
| CPU time | 17.65 s | 102.65 ms |
| Rows scanned | 2,331,598 | 1 |
| Remote calls | 127 | 1 |
| Data read | 50.3 MB | 416 KB |

The unforced plan scanned approximately 2.3 million `typeOf` edges and performed approximately 2.3 million containment probes. Forcing ancestor-first started from the requested place and avoided that global scan.

The saved A/B comparison isolated `FORCE_JOIN_ORDER`; columnar scanning was validated separately and is already part of the existing ancestor-first Node query path.

## Testing

- Added planner coverage for direct single-value, multi-value, duplicate-value, and unsupported filter shapes.
- Added SQL goldens verifying:
  - Single-value direct containment remains unforced.
  - Multi-value direct containment uses ancestor-first ordering, columnar scan, and `InEdge`.
- Confirmed existing linked-containment goldens remain unchanged.
- Ran:
  - `go test ./internal/server/spanner/... -count=1`
  - `./run_test.sh -f`

### DIFF
--- a/internal/server/spanner/golden/query_builder/get_node_edges_contained_in_place_multiple_types_ancestor_first.sql
+++ b/internal/server/spanner/golden/query_builder/get_node_edges_contained_in_place_multiple_types_ancestor_first.sql
@@ -1,0 +1,38 @@
+		@{SCAN_METHOD=COLUMNAR}
+		GRAPH DCGraph MATCH (m:Node
+		WHERE
+			m.subject_id = 'geoId/5508971700')<-[e:Edge
+		WHERE
+			e.predicate = 'containedInPlace']-(n:Node),
+		@{FORCE_JOIN_ORDER=TRUE}
+		(n)-[@{FORCE_INDEX=InEdge}filter0:Edge
+		WHERE
+			filter0.predicate = 'typeOf'
+			AND filter0.object_id IN ('County','City')]->
+		RETURN
+			m.subject_id,
+			n.subject_id AS object_id,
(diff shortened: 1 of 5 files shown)


n-h-diaz (approved):
THANK YOU!! 😃

PR #2049, opened 2026-07-24
Expose target parameter in SearchIndicatorsRequest

keyurva:
- Added an optional `target` field to `SearchIndicatorsRequest` proto to allow controlling resolution sources (`custom_only`, `base_only`, `base_and_custom`).
- Propagated the `target` parameter from the Agent Service indicator search request to candidate resolution.

### DIFF
--- a/internal/agent/search_indicators.go
+++ b/internal/agent/search_indicators.go
@@ -87,7 +87,7 @@ func (s *Service) SearchIndicators(
 		var err error
 		oversampledLimit := limit * 2
 		// Fetch candidates using embeddings search. Leaf expansion is enabled depending on expandTopics.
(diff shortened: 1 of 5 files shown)


internal/agent/search_indicators.go
clincoln8, line 137: nit: also check if target is a valid value. relying on v2/resolve's validation only would result in returning a less helpful 500 internal error.

clincoln8, line 521: does topic resolver also accept target?

clincoln8 (approved):
thanks!

internal/agent/search_indicators.go
keyurva: Done.

keyurva: Yes it does.

PR #2052, opened 2026-07-24
update the resolve v2 diverting logic to allows prod rollout with percentage

shixiao-coder:


internal/server/handler_v2.go
```diff
@@ -160,14 +160,14 @@ func (s *Server) shouldRouteResolveToDispatcher(ctx context.Context, resolver st
...
-				return true, nil
+				return s.shouldDivertV2(ctx), nil
 			} else {
 				// Force Legacy
 				return false, nil
 			}
 		}
 		// Default: use Spanner if configured AND default routing flag is true
-		return s.flags != nil && s.flags.EnableSpannerSearchEmbeddings && s.isSpannerEnabled(), nil
+		return s.flags != nil && s.flags.EnableSpannerSearchEmbeddings && s.isSpannerEnabled() && s.shouldDivertV2(ctx), nil
```
clincoln8: EnableSpannerSearchEmbeddings needs to be flipped in prod

is still correct to use isSpannerEnabled here? should it just be s.flags.EnableSpannerSearchEmbeddings && s.shouldDivertV2(ctx)?

shixiao-coder: I'm fine either way, should be fine to remove

shixiao-coder: updated

internal/server/handler_v2.go
nick-nlb, line 152: I think this change might have an unanticipated effect relating to this header:
X-V2Resolve-Indicator-Spanner (XV2ResolveIndicatorSpanner).

The website never actually sends this header, and so this won't have a bearing on the fractional rollout. The website only sends (or doesn't send) the X-Divert-Spanner header.

The goal is:
* if `EnableSpannerSearchEmbeddings` is true and we would divert generally to Spanner, then we divert resolve to Spanner.
* if `EnableSpannerSearchEmbeddings` is false and we would divert generally to Spanner, then we do not divert resolve to Spanner.
* if `XV2ResolveIndicatorSpanner` is true (not used by website) and Spanner is available, the nwe divert resolve to Spanner (regardless of whether we would generally divert other calls.

With this change, `XV2ResolveIndicatorSpanner` no longer diverts unless we would do general Spanner diversion. I don't know what the original purpose of this flag was, or where it is used, but I think it should still divert (so long as Spanner is available). regardless of whether other paths would divert.

Another way to think about the goal is, if Spanner is enabled but not in global use (which will be the state when we start ramp up), then calls from website will use Spanner embeddings if `EnableSpannerSearchEmbeddings` is true and we are in Spanner diversion mode.

clincoln8: Just for context, I introduced X-V2Resolve-Indicator-Spanner for dev-only purposes so maintaining current behavior isn't too important. Ultimately it's supposed to help us test stuff, so it's up to us configure as needed.

That being said, it sounds like there's a couple tweaks needed to reach desired behavior here.

nick-nlb: Following from this and our offline discussion:

@shixiao-coder Let's update this to match this logic:

* `XV2ResolveIndicatorSpanner` would behave as it did before, diverting regardless of other logic for testing.
*  Otherwise, if `EnableSpannerSearchEmbeddings` is false, we never use Spanner for embeddings
* Otherwise, if `EnableSpannerSearchEmbeddings` is true, then we follow the shouldDivert logic, and use embeddings if we would be using Spanner for other v2 calls.

shixiao-coder: Sure, updated this logic, lmk if this is fine now

n-h-diaz (approved):

PR #2066, opened 2026-07-27
fix(server): return gRPC InvalidArgument (HTTP 400) on client input errors to prevent 5xx alerts (#1858)

beets:
### Summary
This PR fixes an issue where client-side input errors on `/v2/` API endpoints returned plain Go `fmt.Errorf(...)` errors. In Go gRPC (`google.golang.org/grpc/status`), any error that is not a `status.Status` defaults to code `2` (`codes.Unknown`), which Envoy/ESPv2 translates to **HTTP 500 Internal Server Error** (`via_upstream`).

Bursts of empty/malformed requests from automated probes and external clients were inflating server 5xx error metrics and repeatedly triggering the Cloud Monitoring alert policy **Mixer API request count/latency alert** (`Produced API - Error Count (5xx response)` condition).

By wrapping client input validation failures in `status.Errorf(codes.InvalidArgument, ...)`, Go gRPC returns status code `3` (`INVALID_ARGUMENT`), which Envoy/ESPv2 maps to **HTTP 400 Bad Request**. Since 4xx responses are excluded from 5xx server error metrics, client mistakes will no longer trigger server alert incidents.

---

### Changes
* **`internal/server/recon`**:
  * **[find_entities.go](file:///Users/cyin/google/mixer/internal/server/recon/find_entities.go#L59-L67)**: Replace `fmt.Errorf("empty input")` and entity count limit errors in `BulkFindEntities` with `status.Errorf(codes.InvalidArgument, ...)`. Prevents empty requests to `/v2/resolve` from returning HTTP 500.
  * **[id.go](file:///Users/cyin/google/mixer/internal/server/recon/id.go#L36-L41)**: Replace `fmt.Errorf("invalid input: ...")` in `ResolveIds` with `status.Errorf(codes.InvalidArgument, ...)`.
* **`internal/server/pagination`**:
  * **[pagination.go](file:///Users/cyin/google/mixer/internal/server/pagination/pagination.go#L26-L58)**: Wrap empty token, decompression, and proto unmarshal errors in `Decode()` and `DecodeNextToken()` with `status.Errorf(codes.InvalidArgument, ...)`. Prevents malformed pagination tokens (`nextToken`) on `/v2/node` and other paginated endpoints from returning HTTP 500.
* **`internal/server/v2/observation`**:
  * **[calculation.go](file:///Users/cyin/google/mixer/internal/server/v2/observation/calculation.go#L53)** & **[calculation_util.go](file:///Users/cyin/google/mixer/internal/server/v2/observation/calculation_util.go#L126-L368)**: Replace `fmt.Errorf(...)` with `status.Errorf(codes.InvalidArgument, ...)` for client formula syntax errors (missing variables, division by zero, invalid binary expressions, unsupported AST operators, and missing facet IDs).

---

### Testing
* **New Unit Tests**:
  * `TestBulkFindEntities_InvalidInput` in `internal/server/recon/find_entities_test.go`
  * `TestResolveIds_InvalidInput` in `internal/server/recon/id_test.go`
  * `TestDecode_InvalidInput` & `TestDecodeNextToken_InvalidInput` in `internal/server/pagination/pagination_test.go`
  * `TestEvalExpr_InvalidInput` in `internal/server/v2/observation/calculation_util_test.go`
* **Verification**:
  * Verified that all new unit tests assert `status.Code(err) == codes.InvalidArgument`.
  * Ran full unit test suites across `internal/server/recon/...`, `internal/server/pagination`, and `internal/server/v2/observation` without cache (`-count=1`). All tests compile and pass cleanly.

---

### Risks & Compatibility
* **HTTP Status Code Shift (`500` → `400`)**: External clients or probes sending empty request bodies (`{}`) or malformed `nextToken` strings will now receive `HTTP 400 Bad Request` instead of `HTTP 500 Internal Server Error`.
* **Error Message Strings Preserved**: Exact error text strings (`"empty input"`, `"formula missing variables"`, etc.) have been preserved for compatibility with client logging or regex matching.

---

### Related Alerts
* Addresses false-positive incidents on Cloud Monitoring Policy [17847086860337716121](https://pantheon.corp.google.com/monitoring/alerting/policies/17847086860337716121?project=datcom-mixer) (**Mixer API request count/latency alert** -> **Produced API - Error Count (5xx response)**).


beets:
Addressed code review feedback in fe38c0a9:
- Added `NilRequest` (`req: nil`) test cases to both `TestBulkFindEntities_InvalidInput` (`find_entities_test.go`) and `TestResolveIds_InvalidInput` (`id_test.go`).
- Verified both handlers return gRPC `codes.InvalidArgument` without panicking.

internal/server/v2/observation/calculation_util.go
```diff
@@ -295,7 +296,7 @@ func EvalExpr(
 		return nil, err
 	}
 	if intermediateResp.variableObs == nil {
-		return nil, fmt.Errorf("nil calculation response")
+		return nil, status.Errorf(codes.InvalidArgument, "nil calculation response")
```
keyurva: Is this condition triggered exclusively by user formula input , or could internal calculation edge cases reach this line? If it's purely input validation, `InvalidArgument` makes sense. Otherwise consider a different status code.

keyurva (approved):
Thanks Carolyn!

internal/server/v2/observation/calculation_util.go
beets: Good catch! This condition cannot be triggered by user formula validation—client formulas missing statistical variables (e.g. constant-only formulas like `"1 + 2"`) are already rejected with `codes.InvalidArgument` ("formula missing variables") before `EvalExpr` is called.

As long as there is at least one variable in the formula, `evalExpr` always returns a non-nil `variableObs` (even if there are 0 matching observations in the database, in which case `variableObs` is empty rather than `nil`). Reaching `intermediateResp.variableObs == nil` here would indicate an internal calculation edge case or invariant violation, so I've updated this to return `codes.Internal` instead of `codes.InvalidArgument`.

PR #2070, opened 2026-07-29
  Enable SDMX query parameter handling in staging and prod ESPv2

rohitkumarbhagat:
## Summary

- Enable `transcodingIgnoreUnknownQueryParameters` in staging and production.
- This allows ESPv2 to transcode SDMX requests containing dynamic query parameters such as `c[variableMeasured]`.
- Keep `transcodingQueryParametersDisableUnescapePlus` disabled to preserve the existing `+` handling for protobuf-backed query parameters.

## Why

SDMX query parameters such as `c[...]` are not fields on `SdmxRestRequest`. Without ignoring unknown query parameters, ESPv2 passes the HTTP request through without transcoding, and Mixer’s gRPC backend returns:

`415 invalid gRPC request content-type ""`

SDMX parses its query parameters from the original request URL, so it preserves `+` without changing ESPv2's global plus-decoding behavior. Enabling the global option would also affect APIs such as `/v2/node` and `/v2/resolve`.

## Verification

Tested the equivalent configuration on autopush with `X-Skip-Cache: true`:

- A valid SDMX data request returned `200` with SDMX CSV.
- A valid SDMX availability request returned `200` with SDMX Structure JSON.
- `Count_Person+Count_Person_Femal` returned the expected `501` SDMX error: `SDMX AND filters are not implemented yet`, rather than 415.
- `observationAbout.containedInPlace+` with `Count_Person` returned `200` for both data and availability.
- Raw `+` and encoded `%2B` produced identical responses.
- `git diff --check` passed.

### DIFF
--- a/deploy/helm_charts/envs/mixer_prod.yaml
+++ b/deploy/helm_charts/envs/mixer_prod.yaml
@@ -18,7 +18,7 @@ mixer:
 
 esp:
   transcodingIgnoreUnknownQueryParameters: true
-  transcodingQueryParametersDisableUnescapePlus: true
+  transcodingQueryParametersDisableUnescapePlus: false
 
 ingress:
   enabled: true
--- a/deploy/helm_charts/envs/mixer_prod.yaml
+++ b/deploy/helm_charts/envs/mixer_prod.yaml
@@ -16,6 +16,10 @@ mixer:
           port: "6379"
   enableOtlp: true
 
+esp:
+  transcodingIgnoreUnknownQueryParameters: true
+  transcodingQueryParametersDisableUnescapePlus: true
+
 ingress:
   enabled: true
   name: mixer-ingress-prod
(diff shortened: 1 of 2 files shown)


deploy/helm_charts/envs/mixer_prod.yaml
```diff
@@ -16,6 +16,10 @@ mixer:
           port: "6379"
   enableOtlp: true
 
+esp:
+  transcodingIgnoreUnknownQueryParameters: true
+  transcodingQueryParametersDisableUnescapePlus: true
```
keyurva: Can we disable this only for the SDMX endpoint so others are not affected?

keyurva (approved):

deploy/helm_charts/envs/mixer_prod.yaml
rohitkumarbhagat: Removed following  setting. It's not required for sdmx.

` transcodingQueryParametersDisableUnescapePlus: true`

`transcodingIgnoreUnknownQueryParameters: true` - this looks harmless for non-sdmx. Extra query params will be rejected in grpc encoding. Hence, would prefer to keep this as it is. Applying to only sdmx endpoints seems more tricky in esp.

PR #2073, opened 2026-07-30
Add V2 internal SDMX endpoints and remote data source support

keyurva:
* Defined internal V2 SDMX data and availability RPCs for inter-instance query federation.
* Implemented server handlers for internal SDMX requests.
* Added SDMX query delegation to the remote client and remote data source.

Manually tested by deploying to dev, and running mixer in dcp mode with the dev mixer as remote:

```bash
curl -X POST http://localhost:8081/v3/internal/sdmx/data \
  -d '{
    "constraints": {
      "variableMeasured": {
        "predicates": [{"value": "Count_Person"}]
      },
      "observationAbout": {
        "predicates": [{"value": "geoId/06"}]
      }
    }
  }' | jq .
```


internal/proto/sdmx/sdmx.pb.go
```diff
@@ -17,7 +17,7 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.30.0
```
rohitkumarbhagat, line 20: for uniformity should we use 1.30.0 compiler version. ./scripts/compile_protos.sh uses the same

proto/service/mixer.proto
```diff
@@ -70,6 +70,22 @@ service Mixer {
...
+    option (google.api.http) = {
+      post : "/v3/internal/sdmx/data"
+      body : "*"
+    };
+  }
+
+  rpc V3InternalSdmxAvailability(datacommons.sdmx.SdmxAvailabilityQuery)
+      returns (datacommons.sdmx.SdmxAvailabilityResult) {
+    option (google.api.http) = {
+      post : "/v3/internal/sdmx/availability"
```
rohitkumarbhagat: The new remote calls use `/v3/internal/sdmx/data` and `/v3/internal/sdmx/availability`. Standard remote configurations point to `https://api.datacommons.org`, where the Apigee policy blocks all `/v3/**` paths. Because remote errors propagate through `fetchAndMerge`, this would cause SDMX requests to fail entirely, including requests with local results. Should we allow these endpoits with api-key

internal/server/remote/client.go
```diff
@@ -139,3 +140,21 @@ func (rc *RemoteClient) FilterStatVarsByEntity(req *pb.FilterStatVarsByEntityReq
...
+	err := util.FetchRemote(rc.metadata, rc.httpClient, "/v3/internal/sdmx/data", req, resp)
+	if err != nil {
+		return nil, err
+	}
+	return resp, nil
+}
+
+func (rc *RemoteClient) SdmxAvailability(req *sdmxpb.SdmxAvailabilityQuery) (*sdmxpb.SdmxAvailabilityResult, error) {
+	resp := &sdmxpb.SdmxAvailabilityResult{}
+	err := util.FetchRemote(rc.metadata, rc.httpClient, "/v3/internal/sdmx/availability", req, resp)
```
rohitkumarbhagat: One behavior worth calling out: this change makes SDMX follow the same fail-closed `fetchAndMerge` semantics as the core APIs. If the local query succeeds but the remote query fails, `datasources.go - errGroup.Wait()` returns the remote error, the local result is discarded, and the merge is never performed. Is that the intended behavior for SDMX, or should transient remote failures fall back to local-only results?

rohitkumarbhagat (approved):
Thanks Keyur!!

proto/service/mixer.proto
keyurva: Done. Changed the route prefix to `/v2/internal/sdmx/...` and updated method names and doc comments accordingly.

internal/proto/sdmx/sdmx.pb.go
keyurva: Done. Recompiled all proto stubs using `protoc-gen-go v1.30.0` via `./scripts/compile_protos.sh`.

internal/server/remote/client.go
keyurva: If we update the fail-closed behavior, we should do it consistently across all `RemoteClient` methods rather than just SDMX. Let's address that in a separate PR.

PR #2077, opened 2026-07-31
Add ResolvePlaces API, tabular place_dcids, and Spanner topic expansion

keyurva:
- Added `ResolvePlaces` endpoint in the agent service to resolve place names to DCIDs with 1:1 row alignment (including placeholder rows for unresolved queries).
- Updated `SearchIndicators` to support tabular `place_dcids` input, preserve input place ordering, and aggregate child topic place availability prior to topic pruning.
- Updated `SpannerDataSource` vector search resolution to expand topic candidates when `expand_topics: true` is set.
- Added table-driven unit tests for place resolution edge cases and place existence filtering.

### Verification

Tested locally as follows:

1. **ResolvePlaces Alignment Test (valid and unresolved places)**:
```bash
curl -X POST http://localhost:8081/v2/agent/resolve_places \
  -d '{
    "queries": ["california", "nonexistent_place_xyz"]
  }' | jq .
```

2. **SearchIndicators with Tabular `place_dcids` & Topic Expansion**:
```bash
curl -X POST http://localhost:8081/v2/agent/search_indicators \
  -d '{
    "query": "health",
    "place_dcids": ["geoId/06"],
    "per_search_limit": 5,
    "expand_topics": true,
    "include_topics": true
  }' | jq .
```


internal/agent/search_indicators.go
```diff
@@ -556,56 +626,120 @@ func (s *Service) fetchDescendantVariables(
...
-
 	for _, c := range candidates {
+		if !isTopic(c) {
+			continue
+		}
+		if topicCount >= limit {
+			break
+		}
 		var placesWithData []string
 		if pStr, ok := c.Metadata[MetadataPlacesWithData]; ok && pStr != "" {
```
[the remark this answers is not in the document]
keyurva: Done.

internal/agent/search_indicators.go
```diff
@@ -556,56 +626,120 @@ func (s *Service) fetchDescendantVariables(
...
-			resp.DcidNameMappings[c.GetDcid()] = c.GetName()
-			translateVariableCandidate(c, placesWithData, resp)
-			varCount++
+			continue
+		}
+		if varCount >= limit {
+			break
+		}
+		var placesWithData []string
+		if pStr, ok := c.Metadata[MetadataPlacesWithData]; ok && pStr != "" {
```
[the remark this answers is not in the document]
keyurva: Done.

shixiao-coder (approved):
LGTM on the vectorSearchResolution part

proto/v2/agent.proto
```diff
@@ -58,11 +73,16 @@ message SearchIndicatorsResponse {
...
-  map<string, google.protobuf.ListValue> dcid_place_type_mappings = 3;
-  repeated Topic topics = 4;
-  repeated Variable variables = 5;
-  ResolvedPlace resolved_parent_place = 6;
+
+  // Topics Table (columns: ["dcid", "name", "placesWithData", "memberTopics", "memberVariables"])
+  Table topics = 2;
+
+  // Variables Table (columns: ["dcid", "name", "placesWithData", "observationProperties"])
+  Table variables = 3;
```
clincoln8: How do we sequence updating the corresponding skills in the mcp server with changes like these? 

https://github.com/datacommonsorg/agent-toolkit/blob/main/packages/datacommons-mcp/datacommons_mcp/instructions/agent_api/skills/data-commons-researcher/SKILL.md#a-response-structure-reference

https://github.com/datacommonsorg/agent-toolkit/blob/main/packages/datacommons-mcp/datacommons_mcp/instructions/agent_api/skills/data-commons-child-places-researcher/SKILL.md#a-response-structure-reference

proto/v2/agent.proto
```diff
@@ -58,11 +73,16 @@ message SearchIndicatorsResponse {
...
+
+  map<string, string> dcid_name_mappings = 4 [deprecated = true];
+  map<string, google.protobuf.ListValue> dcid_place_type_mappings = 5 [deprecated = true];
```
clincoln8: I think typically for protos, we'd not change the field number or types; we'd instead keep tags 2–6 marked `[deprecated = true]` and create new fields with new numbers for the tabular outputs (e.g. `Table topics = 7;` and `Table variables = 8;`). 

But since we don't really have direct grpc clients, are breaking changes like this typically okay in mixer?

internal/agent/resolve_places.go
```diff
@@ -0,0 +1,205 @@
...
+			typeList.AsSlice(),
+		})
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to build row for place %s: %v", entry.dcid, err)
+		}
+		rows = append(rows, row)
+	}
+
+	return &pbv2.Table{
+		Columns: []string{colQuery, colDcid, colName, colTypeOf},
```
clincoln8, line 202: Longer term, should we return containment information (e.g. parent state/country) in `ResolvePlacesResponse` or accept an optional `parent_place` / `ancestor` scoping parameter in `ResolvePlacesRequest`? 

Right now, if there are multiple places with the same name (like "Springfield" or "London"), `ResolvePlaces` returns multiple candidate rows, but without parent/containment info or scoping, it's hard for the LLM to differentiate between them or scope to a specific region.

clincoln8 (approved):
Thanks!!

proto/v2/agent.proto
keyurva: That is an excellent catch. I was originally under the impression that we no longer document output formats in our instructions. That is clearly not the case. So I've now made changes to keep the legacy response format for legacy requests and return tables (using new fields) only for the new place dcid requests.

PTAL.

proto/v2/agent.proto
keyurva: Yes, ok in mixer since we only ever return json. Having said that, now that I've brought back the legacy response, existing fields retain their numbers.

internal/agent/resolve_places.go
keyurva: This I'm thinking we keep as is. The agent already has context on which Springfield or London it is looking for and will choose the appropriate candidate. We can also keep the fully qualified name instruction still around which will resolve to the correct one most of the time.

clincoln8: Okay, I think this is resolved if we keep the fully qualified name instruction around. Otherwise, I don't think there's anything in the response itself that would differentiate two Springfields. The rows just look like
```     [
        "springfield",
        "geoId/1268275",
        "Springfield",
        [
          "City",
          "Place"
        ]
      ],
      [
        "springfield",
        "geoId/1372780",
        "Springfield",
        [
          "City",
          "Place"
        ]
      ],
```
so they'd need to be specific in the request

clincoln8 (approved):
Thanks!

internal/agent/resolve_places.go
keyurva: True (currently it could differentiate based on the dcid in some cases but that's ugly and brittle). We can look into including parent info like you mentioned. Need to assess how expensive it is for resolve calls. Let's chat offline.

PR #2079, opened 2026-08-03
Set V2 diversion on prod standalone mixer to 100%

n-h-diaz:
Following the release schedule, doing full ramp of Spanner diversion on standalone mixer, after successful website release


juliawu (approved):
Woohoo! Ship it!

nick-nlb (approved):
A note that following newly-instituted protocols, as a production flag change, this needs to be announced and acked in on-call.

PR #2082, opened 2026-08-05
Add download filename to SDMX-CSV responses

rohitkumarbhagat:
## Summary

- return `Content-Disposition: attachment; filename="data.csv"` for SDMX-CSV responses
- pass the negotiated content disposition from the SDMX service to the gRPC handler
- log header-setting failures while continuing to return the CSV response
- add focused coverage for successful header delivery and best-effort failure handling

### DIFF
--- a/internal/server/sdmx/service/data.go
+++ b/internal/server/sdmx/service/data.go
@@ -26,6 +26,8 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+const csvContentDisposition = `attachment; filename="data.csv"`
+
 func (s *Service) Data(ctx context.Context, request Request) (*Response, error) {
 	restRequest, err := restv2.ParseDataRequest(request.Tail, request.OriginalURI)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


internal/server/sdmx/service/data.go
```diff
@@ -26,6 +26,8 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+const csvContentDisposition = `attachment; filename="data.csv"`
```
kmoscoe, line 29: Ideally, it would be nice to reflect the query used to get the results (this is what we do for the download tool, e.g. Canada_AdministrativeArea1_LifeExpectancy_Person.csv). If that's too hard, let's make it a bit more recognizable, like datacommons.csv or dc_data.csv

rohitkumarbhagat: renamed to dc_data.csv for now. Deriving from query for sdmx would be non-trivial. can be added as a followup task.

kmoscoe: Sounds good. Thank you Rohit!

kmoscoe (approved):

keyurva (approved):

PR #2089, opened 2026-08-12
Adds the new UNData prefix to list of hardcoded for UN behavior

gmechali:


n-h-diaz (approved):
Thanks! This is fine to fix the existing UN bug, but long term maybe can have all curated svgs default to using full name (if there's a good way to detect this) vs relying on prefixes, as the prefix list was mostly for base

gmechali:
Yes this is definitely the wrong fix here. We should remove all hardcoding. Honestly it seems like no one is relying on the autogenerated SVG hierarchy at this point (other than us?). We should maybe just get rid of it to get rid of legacy code, and inthat case we won't need any of this logic to identify "curated" hierarchies

PR #2090, opened 2026-08-13
Support unresolved objects in spanner node queries

n-h-diaz:
Node federation part 1: handle dangling edges

When a destination node does not exist, return the id anyway, assuming that this is a reference to a node in another graph. For now, set a default type=Thing, which aligns with legacy custom DC behavior. 

Next steps would be to make the federated call to fetch the actual node metadata. 

Details:
Previously we used NEXT MATCH (n:Node) to fetch node metadata. However, MATCH enforces formal graph semantics, causing this to be evaluated as an Inner Join. Because we are utilizing informational foreign keys to support federated/external destination nodes, any edge pointing to a node that doesn't exist locally was being dropped.

This PR replaces the strict NEXT MATCH block with a standard SQL scalar subquery using LET dest = (SELECT AS STRUCT ...). The NEXT LET subquery bypasses GQL strictness. If the node exists locally, it retrieves the metadata into a Struct. If the node is federated (missing locally), the Struct will be NULL. 

Testing:
Added unit tests for requests with dangling edges
Note: the emulator enforces stricter graph requirements, so will not match edges with dangling objects


internal/server/spanner/dsutil.go
```diff
@@ -244,7 +247,15 @@ func nodeEdgesToLinkedGraph(edges []*Edge) (*pbv2.LinkedGraph, error) {
...
-		if len(edge.Types) == 0 { // If a node has no types, it's a terminal value.
+		if len(edge.Types) == 0 && edge.Value == "" && edge.Bytes == nil { // If a node has no types, value, bytes, treat it as unresolved.
+			// Set the node object to the edge object.
+			// Literals should be resolved locally, so unresolved nodes should be references.
+			node.Dcid = edge.ObjectID
+			// For now, set type to default "Thing".
+			// This is consistent with the behavior in legacy custom DC.
+			// TODO: Federate the request to get node metadata.
+			node.Types = []string{typeThing}
+		} else if len(edge.Types) == 0 { // Otherwise, assume the node is local. If a node has no types, it's a terminal value.
```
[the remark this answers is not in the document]
n-h-diaz: thanks, to be safe added explicit field to indicate if node is resolved

internal/server/spanner/dsutil.go
```diff
@@ -244,7 +247,15 @@ func nodeEdgesToLinkedGraph(edges []*Edge) (*pbv2.LinkedGraph, error) {
...
-		if len(edge.Types) == 0 { // If a node has no types, it's a terminal value.
+		if !edge.Resolved {
+			// Set the node object to the edge object.
+			// Literals should be resolved locally, so unresolved nodes should be references.
+			node.Dcid = edge.ObjectID
+			// For now, set type to default "Thing".
+			// This is consistent with the behavior in legacy custom DC.
+			// TODO: Federate the request to get node metadata.
+			node.Types = []string{typeThing}
+		} else if len(edge.Types) == 0 { // Otherwise, assume the node is local. If a node has no types, it's a terminal value.
```
gmechali, line 258: Is this a consequence of us not typing our string nodes (see email from me today)?

internal/server/spanner/golden/datasource_test.go
```diff
@@ -151,7 +151,7 @@ func TestSpannerResolve(t *testing.T) {
 	if client == nil {
```
gmechali, line 151: Are any of these tests able to mock a federated world?

internal/server/spanner/dsutil.go
```diff
@@ -44,6 +44,9 @@ import (
 const (
 	// Used for Facet responses with an entity expression.
 	entityPlaceholder = ""
+
+	// Default type for unresolved nodes.
+	typeThing = "Thing"
```
gmechali, line 49: Why default to Thing, rather than something like "Unknown" or just throwing error?

internal/server/spanner/statements.go
```diff
@@ -292,17 +310,26 @@ OR CreationTimestamp > (
             predicate,
             object_id,
             provenance%s
-        NEXT MATCH (n:Node)
-        WHERE
-            n.subject_id = object_id
+        NEXT LET dest = (
```
gmechali, line 313: There seems to be a lot of duplication across these files. I wonder if we could use some smaller named statements to reuse this, and make it easier to maintain longer term?

gmechali (approved):
approving from my pov - left a few minor comments. But pls wait for someone with more domain knowledge on here to review as well :)

internal/server/spanner/dsutil.go
keyurva: This is inline with Custom DC behavior and apparently how unknown types were returned traditionally. So let's go with this for this part. We can decide how we want to handle them when we do the actual federation.

internal/server/spanner/statements.go
keyurva: We've gone back and forth on this. Templates that are similar to the eventual sql have been easier to maintain. Having said that, we can consider refactors that help avoid excessive duplication. But let's address that separately.

keyurva (approved):
Thanks Natalie!

One question: Do we have data with unresolved nodes? If so, adding a golden test for that would be nice.

internal/server/spanner/dsutil.go
n-h-diaz: This is because the node doesn't have any typeOf edges

internal/server/spanner/golden/datasource_test.go
n-h-diaz: Added a test to datasource_test to mock a dangling edge and a top level datasources test to validate that unresolved node responses get merged with other datasource node responses (note this isn't actually making a second federated call at the moment, so this test just validates that the unresolved response doesn't cause any merging issues and we still return a valid response - once we add the federation, we can test the metadata hydration)

n-h-diaz:
> Thanks Natalie!
> 
> One question: Do we have data with unresolved nodes? If so, adding a golden test for that would be nice.

Spoke offline - the emulator doesn't support non strict graph data, so for now I've just been mocking the unresolved responses in unit tests. (I added some sample data temporarily to dc_graph_1 to also test this)

PR #2091, opened 2026-08-15
Use V2 for remote mixer calls for legacy V2 implementations

n-h-diaz:
Previously legacy v2 implementations for certain endpoints were directly calling the v1 implementations which had hardcoded v1 remote api requests. This meant that custom dc instances not on spanner were still making v1 calls to remote mixer. 

This PR updates these v2 endpoints to make corresponding v2 remote mixer calls instead by abstracting out the core endpoint implementation to be called by both v1 and v2 versions with the corresponding remoteAPIPath as an input

Affected endpoints
- V2BulkVariableInfo
- V2BulkVariableGroupInfo
- V2GetLocationsRankings

### DIFF
--- a/internal/server/handler.go
+++ b/internal/server/handler.go
@@ -108,7 +108,7 @@ func (s *Server) GetRelatedLocations(
 	return localResp, nil
 }
 
-// GetLocationsRankings implements API for Mixer.GetLocationsRankings.
+// fetchGetLocationsRankings fetches local location rankings, falling back to remote mixer if local data is empty.
 func (s *Server) fetchGetLocationsRankings(
 	ctx context.Context, in *pb.GetLocationsRankingsRequest, remoteAPIPath string,
 ) (*pb.GetLocationsRankingsResponse, error) {
--- a/internal/server/handler.go
+++ b/internal/server/handler.go
@@ -108,8 +108,8 @@ func (s *Server) GetRelatedLocations(
 	return localResp, nil
 }
 
-// fetchGetLocationsRankings fetches local location rankings, falling back to remote mixer if local data is empty.
-func (s *Server) fetchGetLocationsRankings(
+// getLocationsRankings fetches local location rankings, falling back to remote mixer if local data is empty.
+func (s *Server) getLocationsRankings(
 	ctx context.Context, in *pb.GetLocationsRankingsRequest, remoteAPIPath string,
 ) (*pb.GetLocationsRankingsResponse, error) {
 	localResp, err := place.GetLocationsRankings(ctx, in, s.store)
@@ -132,7 +132,7 @@ func (s *Server) fetchGetLocationsRankings(
 func (s *Server) GetLocationsRankings(
 	ctx context.Context, in *pb.GetLocationsRankingsRequest,
 ) (*pb.GetLocationsRankingsResponse, error) {
-	return s.fetchGetLocationsRankings(ctx, in, "/v1/place/ranking")
+	return s.getLocationsRankings(ctx, in, "/v1/place/ranking")
 }
 
 // GetPlaceStatVars implements API for Mixer.GetPlaceStatVars.
--- a/internal/server/handler.go
+++ b/internal/server/handler.go
@@ -109,8 +109,8 @@ func (s *Server) GetRelatedLocations(
 }
 
 // GetLocationsRankings implements API for Mixer.GetLocationsRankings.
-func (s *Server) GetLocationsRankings(
-	ctx context.Context, in *pb.GetLocationsRankingsRequest,
+func (s *Server) fetchGetLocationsRankings(
+	ctx context.Context, in *pb.GetLocationsRankingsRequest, remoteAPIPath string,
 ) (*pb.GetLocationsRankingsResponse, error) {
 	localResp, err := place.GetLocationsRankings(ctx, in, s.store)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


internal/server/handler_v2.go
```diff
@@ -608,8 +608,8 @@ func (s *Server) V2BulkVariableGroupInfo(
 
 	v2StartTime := time.Now()
 
-	// Use the V1 implementation for now.
-	v1Resp, err := s.BulkVariableGroupInfo(ctx, in)
+	// Fetch and merge local data and remote data, using the v2 API path for remote fetch.
+	v1Resp, err := s.fetchAndMergeBulkVariableGroupInfo(ctx, in, "/v2/bulk/info/variable-group")
 	if err != nil {
 		return nil, err
 	}
```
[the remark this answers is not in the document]
n-h-diaz: done

internal/server/handler_v2.go
```diff
@@ -557,8 +557,8 @@ func (s *Server) V2BulkVariableInfo(
 
 	v2StartTime := time.Now()
 
-	// Use the V1 implementation for now.
-	v2Resp, err := s.BulkVariableInfo(ctx, in)
+	// Fetch and merge local data and remote data, using the v2 API path for remote fetch.
+	v2Resp, err := s.fetchAndMergeBulkVariableInfo(ctx, in, "/v2/bulk/info/variable")
 	if err != nil {
 		return nil, err
 	}
```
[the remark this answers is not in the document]
n-h-diaz: done

internal/server/handler.go
```diff
@@ -109,8 +109,8 @@ func (s *Server) GetRelatedLocations(
 }
 
 // GetLocationsRankings implements API for Mixer.GetLocationsRankings.
```
nick-nlb: This comment is now out of date. Perhaps: 

// fetchGetLocationsRankings fetches local location rankings, falling back to remote mixer if local data is empty.

internal/server/handler_v2.go
```diff
@@ -654,15 +661,15 @@ func (s *Server) V2GetLocationsRankings(
 	ctx context.Context, in *pb.GetLocationsRankingsRequest,
 ) (*pb.GetLocationsRankingsResponse, error) {
 
-	// Use the V1 implementation for now.
-	v1Resp, err := s.GetLocationsRankings(ctx, in)
+	// Fetch and merge local data and remote data, using the v2 API path for remote fetch.
```
nick-nlb: In the case of rankings, it looks like `fetchGetLocationsRankings` doesn't fetch and merge both remote and local data, but rather fetches local data and if empty, as a fallback fetches remote data.

internal/server/handler_v1.go
```diff
@@ -108,9 +108,8 @@ func (s *Server) BulkPlaceInfo(
 	return result, nil
 }
 
-// BulkVariableInfo implements API for mixer.BulkVariableInfo.
-func (s *Server) BulkVariableInfo(
-	ctx context.Context, in *pbv1.BulkVariableInfoRequest,
+func (s *Server) fetchAndMergeBulkVariableInfo(
```
nick-nlb, line 113: The function name is descriptive, but perhaps a doc comment above this function (and fetchAndMergeBulkVariableGroupInfo) to note that  these are shared implementations for both v1 and v2 handlers?

internal/server/handler_v1_test.go
```diff
@@ -222,7 +222,7 @@ func TestBulkVariableInfo(t *testing.T) {
 		localBulkVariableInfoFunc = func(_ context.Context, _ *pbv1.BulkVariableInfoRequest, _ *store.Store) (*pbv1.BulkVariableInfoResponse, error) {
 			return tc.localResponse, nil
 		}
-		remoteBulkVariableInfoFunc = func(_ *Server, _ *pbv1.BulkVariableInfoRequest) (*pbv1.BulkVariableInfoResponse, error) {
+		remoteBulkVariableInfoFunc = func(_ *Server, _ *pbv1.BulkVariableInfoRequest, _ string) (*pbv1.BulkVariableInfoResponse, error) {
```
nick-nlb: Here we aren't testing the content of the string. Could we convert _string to `remotePathAPI`, and then inside verify that it is the string expected?

internal/server/handler_v2.go
```diff
@@ -654,15 +661,15 @@ func (s *Server) V2GetLocationsRankings(
...
 	}
-	if v1Resp == nil {
+	if v2Resp == nil {
 		return &pb.GetLocationsRankingsResponse{}, nil
 	}
-	v2Resp := proto.Clone(v1Resp).(*pb.GetLocationsRankingsResponse)
+	v2Resp = proto.Clone(v2Resp).(*pb.GetLocationsRankingsResponse)
 
 	// Set the Legacy field to true to indicate that this response is from the V1 implementation.
 	v2Resp.Legacy = true
```
nick-nlb, line 675: Could we also add unit tests to `handler_v2.go` to verify that the refactored v2 endpoints all pass their corresponding `/v2/...` remote API paths?

nick-nlb (commented):
Thank you for looking at this Natalie!

I've tested this between the two versions using local federation and this does now federate correctly on the given three endpoints with the change (I'll separately send a doc outlining the tests).

Some minor comments regarding comments and doc comments, and testing.

internal/server/handler_v1.go
n-h-diaz: done

internal/server/handler_v2.go
n-h-diaz: done

internal/server/handler_v2.go
n-h-diaz: thanks for the catch - updated the comment

internal/server/handler_v1_test.go
n-h-diaz: done

internal/server/handler.go
n-h-diaz: done

internal/server/handler.go
```diff
@@ -108,9 +108,9 @@ func (s *Server) GetRelatedLocations(
 	return localResp, nil
 }
 
-// GetLocationsRankings implements API for Mixer.GetLocationsRankings.
-func (s *Server) GetLocationsRankings(
-	ctx context.Context, in *pb.GetLocationsRankingsRequest,
+// fetchGetLocationsRankings fetches local location rankings, falling back to remote mixer if local data is empty.
+func (s *Server) fetchGetLocationsRankings(
```
nick-nlb: nit: (unless intended), `fetch` and `get` together might be duplicated (was this intended here?)

nick-nlb (approved):
Thank you Natalie!

internal/server/handler.go
n-h-diaz: good point, updated to just use "get" to match GetLocationsRankings

Thanks for the review!

PR #2092, opened 2026-08-18
Enrich unresolved nodes via federated node queries

keyurva:
- Add `UnresolvedNodeProcessor` to dispatcher pipeline to enrich responses by federating node queries for unresolved nodes to a remote data source (i.e. base DC)
- Wire processor into server startup when a remote data source is configured

### Notes
- This requires DCP spanner instances to specify an informational (`NOT ENFORCED`) foreign key on `Edge.object_id` referencing `Node.subject_id` so the query optimizer does not drop edges pointing to unresolved destination nodes:
  ```sql
  ALTER TABLE Edge
  ADD CONSTRAINT FK_Edge_Object_Node
  FOREIGN KEY (object_id) REFERENCES Node (subject_id) NOT ENFORCED;
  ```

### Testing
- Tested locally with a DCP database and remote DC instance, confirming unresolved nodes are returned and enriched with their name and types.

### DIFF
--- a/internal/server/dispatcher/unresolved_node_processor.go
+++ b/internal/server/dispatcher/unresolved_node_processor.go
@@ -170,6 +170,9 @@ func hydrateNodeResponse(resp *pbv2.NodeResponse, infoByDcid map[string]*entityI
 	for _, graph := range resp.GetData() {
 		for _, arc := range graph.GetArcs() {
 			for _, node := range arc.GetNodes() {
+				if node == nil {
+					continue
+				}
 				info, ok := infoByDcid[node.GetDcid()]
 				if !ok {
 					continue
--- a/internal/server/dispatcher/unresolved_node_processor.go
+++ b/internal/server/dispatcher/unresolved_node_processor.go
@@ -1,0 +1,185 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dispatcher
+
+import (
+	"fmt"
+	"log/slog"
+	"slices"
+
+	pb "github.com/datacommonsorg/mixer/internal/proto"
+	pbv2 "github.com/datacommonsorg/mixer/internal/proto/v2"
+	"github.com/datacommonsorg/mixer/internal/server/datasource"
+	"github.com/datacommonsorg/mixer/internal/server/datasources"
+)
+
+const (
+	unresolvedNodeType = "Thing"
+	nodeInfoProperty   = "->[name, typeOf]"
+	predicateName      = "name"
+	predicateTypeOf    = "typeOf"
+)
+
+type entityInfo struct {
+	name  string
+	types []string
+}
+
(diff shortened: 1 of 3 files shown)


internal/server/dispatcher/unresolved_node_processor.go
```diff
@@ -0,0 +1,187 @@
...
+
+	return infoByDcid
+}
+
+// hydrateNodeResponse enriches unresolved entities in the response with their fetched name and types.
+func hydrateNodeResponse(resp *pbv2.NodeResponse, infoByDcid map[string]*entityInfo) {
+	for _, graph := range resp.GetData() {
+		for _, arc := range graph.GetArcs() {
+			for _, node := range arc.GetNodes() {
+				info, ok := infoByDcid[node.GetDcid()]
```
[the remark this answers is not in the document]
keyurva: Done.

internal/server/dispatcher/unresolved_node_processor.go
```diff
@@ -0,0 +1,190 @@
...
+	for dcid, graph := range resp.GetData() {
+		info := &entityInfo{}
+
+		if nameArc, ok := graph.GetArcs()[predicateName]; ok {
+			for _, node := range nameArc.GetNodes() {
+				if val := node.GetValue(); val != "" {
+					info.name = val
+					break
+				}
+				if name := node.GetName(); name != "" {
```
n-h-diaz, line 145: just for my own understanding, when would we use name vs value?

internal/server/dispatcher/unresolved_node_processor.go
```diff
@@ -0,0 +1,190 @@
...
+					info.name = name
+					break
+				}
+			}
+		}
+
+		if typeArc, ok := graph.GetArcs()[predicateTypeOf]; ok {
+			for _, node := range typeArc.GetNodes() {
+				if typeDcid := node.GetDcid(); typeDcid != "" {
+					info.types = append(info.types, typeDcid)
```
n-h-diaz: since node metadata is just an aggregate (vs per provenance), could we filter duplicates?

n-h-diaz (approved):
Thank you! 

a few high level comments (not for this pr but just to think about)
- would we ever want to federate across all datasources or would remote be sufficient?
- re: using Thing as heuristic for unresolved - do we have a sense of how many nodes actually have type Thing? (so would be making an extra call here) not sure how often this shows up in dcp data

internal/server/dispatcher/unresolved_node_processor.go
keyurva: This is just to ensure we capture the name regardless of whether the source returns it as a terminal literal value or an `EntityInfo` object with `Name` populated.

Let me know if you think we should only look for it one way.

internal/server/dispatcher/unresolved_node_processor.go
keyurva: Good catch - done.

keyurva (commented):
Thanks for the review!

> would we ever want to federate across all datasources or would remote be sufficient?

For DCP, unresolved foreign-key nodes point to base DC, so `remoteDataSource` is sufficient for our current use case. Since `NewUnresolvedNodeProcessor` accepts a generic `DataSource`, we can pass a composite / multi-data source in the future if needed.

> re: using Thing as heuristic for unresolved - do we have a sense of how many nodes actually have type Thing? (so would be making an extra call here) not sure how often this shows up in dcp data

The heuristic checks for `Types == ["Thing"]` AND `Name == ""`. Concrete entities with `typeOf: Thing` should almost always have a populated name or more specific types. So both empty name and `Thing` should be rare.

internal/server/dispatcher/unresolved_node_processor.go
n-h-diaz: thanks for the explanation, sounds good to keep both to be safe

PR #2094, opened 2026-08-21
Make Redis protobuf cache keys deterministic

rohitkumarbhagat:
## Summary

- Use deterministic protobuf marshaling when generating Redis cache keys.
- Add coverage ensuring repeated hashing of the same map-bearing SDMX request produces a stable key.

This prevents protobuf map iteration order from producing different cache keys for the same request.

Predicate-list canonicalization is intentionally out of scope.

## Testing

- `go test ./internal/server/redis -count=1`
- `git diff --check`


n-h-diaz (approved):
Thank you!

PR #2097, opened 2026-08-25
Clean up some deprecated V0/V1 endpoints

n-h-diaz:
Includes
* /node/property-labels
* /node/property-values
* /node/triples
* /node/places-in
* /stat/value
* /place/stat-vars
* /v1/query
* /v1/recognize/entities

These endpoints have been deprecated on Apigee and have had 0 requests on prod mixer/website for >2 weeks


proto/node.proto
```diff
@@ -63,55 +56,6 @@ message PropertyLabels {
...
-// Request to get all triples linking to the given nodes.
-message GetTriplesRequest {
-  // The dcids of the nodes to query for.
-  repeated string dcids = 1;
-  // Maximum number of triples for each property and type of the neighbor.
-  int32 limit = 2;
-}
-
 // Response returned by GetTriples.
 message GetTriplesResponse {
```
keyurva: Is the response proto still needed?

keyurva: And if this is removed, can `internal/server/convert/triple.go` also be deleted?

internal/server/v0/propertyvalue/property_value.go
keyurva, line 1: If certain methods are only used in test (`filterEntities`), consider removing the test and the method.

Similarly for methods in `recognize.go` used only in `recognize_test.go`.

keyurva (approved):
Thanks!

proto/node.proto
n-h-diaz: yes thank you, done!

internal/server/v0/propertyvalue/property_value.go
n-h-diaz: good catch, removed

PR #2102, opened 2026-09-01
Resolve external schema properties in Spanner multi-entity SDMX shape

keyurva:
* Updated `resolveSdmxEntityShape` to check `ObjectID` before falling back to `Value` so external schema properties in DCP instances are not omitted from SDMX dimensions.

#### 1. DCP Mode (Multi-Entity Database with External Property `medicalCondition`)
* Verified SDMX availability returns HTTP 200 with available component values:
```bash
curl -g -i "http://localhost:8081/sdmx/v3/availability/dataflow/DC/DF_OBS/1.0.0/*/medicalCondition?c[variableMeasured]=HealthAidFunding"
```
* Verified SDMX data returns HTTP 200 with observation CSV rows:
```bash
curl -g -i "http://localhost:8081/sdmx/v3/data/dataflow/DC/DF_OBS/1.0.0/*?c[variableMeasured]=HealthAidFunding&c[medicalCondition]=HIV"
```

#### 2. Base DC Mode
* Verified SDMX data returns HTTP 200 with observation CSV rows:
```bash
curl -g -i "http://localhost:8081/sdmx/v3/data/dataflow/DC/DF_OBS/1.0.0/*?c[variableMeasured]=Median_Age_Person&c[observationAbout]=country/USA"
```
* Verified SDMX availability returns HTTP 200:
```bash
curl -g -i "http://localhost:8081/sdmx/v3/availability/dataflow/DC/DF_OBS/1.0.0/*/observationAbout?c[variableMeasured]=Median_Age_Person&c[observationAbout]=country/USA"
```

### DIFF
--- a/internal/server/spanner/query_multientity.go
+++ b/internal/server/spanner/query_multientity.go
@@ -891,7 +891,7 @@ func resolveSdmxEntityShape(
 			if edge.Predicate != "observationProperties" {
 				continue
 			}
-			property := strings.TrimSpace(edge.Value)
+			property := edgeObservationProperty(edge)
 			if property == "" {
 				continue
 			}
@@ -1003,4 +1003,16 @@ func sortedUniqueStrings(values []string) []string {
 		seen[value] = struct{}{}
 	}
 	return slices.Sorted(maps.Keys(seen))
+}
+
+// edgeObservationProperty returns the observation property name from an edge,
+// preferring ObjectID for external reference nodes and falling back to Value.
+func edgeObservationProperty(edge *Edge) string {
+	if edge == nil {
+		return ""
+	}
+	if id := strings.TrimSpace(edge.ObjectID); id != "" {
+		return id
+	}
+	return strings.TrimSpace(edge.Value)
 }
--- a/internal/server/spanner/query_multientity.go
+++ b/internal/server/spanner/query_multientity.go
@@ -1006,13 +1006,14 @@ func sortedUniqueStrings(values []string) []string {
 }
 
 // edgeObservationProperty returns the observation property name from an edge,
-// preferring ObjectID for external reference nodes and falling back to Value.
+// preferring Value for locally resolved nodes and falling back to ObjectID
+// for external reference nodes not materialized in the local Node table.
 func edgeObservationProperty(edge *Edge) string {
 	if edge == nil {
 		return ""
 	}
-	if id := strings.TrimSpace(edge.ObjectID); id != "" {
-		return id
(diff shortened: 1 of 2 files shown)


internal/server/spanner/query_multientity.go
```diff
@@ -1004,3 +1004,15 @@ func sortedUniqueStrings(values []string) []string {
...
+// edgeObservationProperty returns the observation property name from an edge,
+// preferring ObjectID for external reference nodes and falling back to Value.
+func edgeObservationProperty(edge *Edge) string {
+	if edge == nil {
+		return ""
+	}
+	if id := strings.TrimSpace(edge.ObjectID); id != "" {
+		return id
+	}
+	return strings.TrimSpace(edge.Value)
```
clincoln8: I don't think we'd ever have a case where we fallback to value here, every edge would have an ObjectID. 

We can either remove this and update the tests, or keep it in order to not change the tests, but leave a comment that we won't actually hit this case.

internal/server/spanner/query_multientity.go
```diff
@@ -1004,3 +1004,15 @@ func sortedUniqueStrings(values []string) []string {
 	}
 	return slices.Sorted(maps.Keys(seen))
 }
+
+// edgeObservationProperty returns the observation property name from an edge,
+// preferring ObjectID for external reference nodes and falling back to Value.
+func edgeObservationProperty(edge *Edge) string {
+	if edge == nil {
+		return ""
+	}
```
clincoln8, line 1014: consider skipping literal values as a defense to not accidentally return hash string values.


```suggestion
	}
	// If the target node was resolved locally in Spanner but has no types,
	// it is a terminal literal value rather than a Property node reference.
	if edge.Resolved && len(edge.Types) == 0 {
		return ""
	}
```

however unit tests would have to be updated to include the Type in edge.

clincoln8 (approved):
Thanks! A couple thoughts, but non blocking. I'll follow up in the thread with a few more questions about how we're handling these fields more broadly.

internal/server/spanner/query_multientity.go
keyurva: Chatted offline to check for Value first and fallback to ObjectID to preserve existing behavior.

internal/server/spanner/query_multientity.go
keyurva: Makes sense but keeping it as-is for now to avoid updating multiple unit tests.

PR #2104, opened 2026-09-03
Add CLI flag and service option to configure agent default topic expansion

keyurva:
Added a CLI flag to control the default topic expansion behavior in the agent `SearchIndicators` API when `expand_topics` is not explicitly provided in the request payload.

### Context & Motivation
Previously, `SearchIndicators` hardcoded topic expansion to `true` when omitted from the request. In custom Data Commons instances with extensive topic ontologies, recursive topic expansion causes response payloads to balloon into multiple megabytes and exhaust client token limits. This change allows instances to configure whether topic expansion occurs by default, while preserving existing behavior for standard deployments.

### Changes
* Added `--agent_default_expand_topics` CLI flag (defaults to `true`) and wired it through `MixerServerOptions` and `agent.Service`.
* Updated `SearchIndicators` to use the configured service default when `expand_topics` is not specified in the request (request-level overrides continue to take precedence).
* Added unit tests for service options and request overrides.

### Verification & Testing
Verified against a test database:

1. **Default (`--agent_default_expand_topics=true` or omitted):**
   * Request without `expand_topics`:
     ```bash
     curl -s -X POST "http://localhost:8081/v2/agent/search_indicators" \
       -d '{"query": "health", "places": ["World"], "include_topics": true}' | wc -c
     ```
     Returned `4,684,981` bytes.

   * Explicit unexpanded override:
     ```bash
     curl -s -X POST "http://localhost:8081/v2/agent/search_indicators" \
       -d '{"query": "health", "places": ["World"], "include_topics": true, "expand_topics": false}' | wc -c
     ```
     Returned `3,280` bytes.

2. **Configured (`--agent_default_expand_topics=false`):**
   * Default request without `expand_topics` in body:
     ```bash
     curl -s -X POST "http://localhost:8081/v2/agent/search_indicators" \
       -d '{"query": "health", "places": ["World"], "include_topics": true}' | wc -c
     ```
     Returned `3,280` bytes.

   * Explicit expanded override:
     ```bash
     curl -s -X POST "http://localhost:8081/v2/agent/search_indicators" \
       -d '{"query": "health", "places": ["World"], "include_topics": true, "expand_topics": true}' | wc -c
     ```
     Returned `4,684,981` bytes.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -54,27 +54,22 @@ type Service struct {
 	defaultExpandTopics bool
 }
 
-// ServiceOption configures a Service instance during construction.
-type ServiceOption func(*Service)
-
-// WithDefaultExpandTopics sets the default topic expansion behavior for search indicators.
-func WithDefaultExpandTopics(expand bool) ServiceOption {
-	return func(s *Service) {
-		s.defaultExpandTopics = expand
-	}
+// ServiceOptions holds optional configuration for agent.Service.
+type ServiceOptions struct {
+	DefaultExpandTopics *bool
 }
 
 // NewService constructs a new Service instance backed by the provided Mixer and Cache.
-func NewService(mixer Mixer, cache *Cache, opts ...ServiceOption) *Service {
-	s := &Service{
+func NewService(mixer Mixer, cache *Cache, opts *ServiceOptions) *Service {
+	defaultExpandTopics := true
+	if opts != nil && opts.DefaultExpandTopics != nil {
+		defaultExpandTopics = *opts.DefaultExpandTopics
+	}
+	return &Service{
 		mixer:               mixer,
 		cache:               cache,
-		defaultExpandTopics: true,
+		defaultExpandTopics: defaultExpandTopics,
 	}
-	for _, opt := range opts {
-		opt(s)
-	}
-	return s
 }
 
 // Reset flushes all cached state inside the service L1 Cache atomically,
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -49,16 +49,32 @@ type Mixer interface {
 // Service orchestrates the aggregation and formatting of lower-level API data in response
 // to conversational assistant/agent queries.
 type Service struct {
-	mixer Mixer
-	cache *Cache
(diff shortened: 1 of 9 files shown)


internal/server/server.go
```diff
@@ -306,6 +306,7 @@ type MixerServerOptions struct {
 	EmbeddingsServiceClient           *resolve.EmbeddingsServiceClient
 	UseSpannerGraph                   bool
 	TopicExpander                     resolve.TopicExpander
+	AgentDefaultExpandTopics          *bool
```
clincoln8: why is this *bool and not bool?

internal/agent/agent.go
```diff
@@ -49,16 +49,32 @@ type Mixer interface {
...
+	mixer               Mixer
+	cache               *Cache
+	defaultExpandTopics bool
+}
+
+// ServiceOption configures a Service instance during construction.
+type ServiceOption func(*Service)
+
+// WithDefaultExpandTopics sets the default topic expansion behavior for search indicators.
+func WithDefaultExpandTopics(expand bool) ServiceOption {
```
clincoln8: can you point kAI at https://github.com/datacommonsorg/mixer/blob/eed881100060abcb74ed2a26ecf3e0d1ff9d7f04/docs/architecture.md?plain=1#L4 and ask it if it should use the Options Struct pattern instead here? 

it might look like 

```
// ServiceOptions holds optional configuration for agent.Service.
type ServiceOptions struct {
	DefaultExpandTopics *bool
}

// NewService constructs a new Service instance backed by the provided Mixer and Cache.
func NewService(mixer Mixer, cache *Cache, opts *ServiceOptions) *Service {
	defaultExpandTopics := true
	if opts != nil && opts.DefaultExpandTopics != nil {
		defaultExpandTopics = *opts.DefaultExpandTopics
	}
	return &Service{
		mixer:               mixer,
		cache:               cache,
		defaultExpandTopics: defaultExpandTopics,
	}
}

```

keyurva: Done.

clincoln8 (approved):
Thank you!

PR #2109, opened 2026-09-11
Skip node data sources that were exhausted on an earlier page

keyurva:
The merged pagination token drops a data source once that source stops returning a cursor, but the next page still fans out to every source. A source with no entry in the token fell back to first-page behavior:

- Spanner re-queried at offset 0.
- The remote client cleared its token and refetched its first page.

Those rows then repeated on every subsequent page.

## Changes

- Changed `getOffset` and `updateNodeRequestNextToken` to report whether the pagination token carries an entry for the data source.
- `SpannerDataSource.Node` and `RemoteClient.Node` return an empty response when it does not.

## Testing

Paged through `/v2/node` for `<-typeOf` on `StatisticalVariable` against a custom instance whose Spanner database holds two statistical variables, so the Spanner source returns both rows on page 1 and issues no cursor. Compared a local server carrying this change against a deployed instance running the current code.

| | Current | This change |
|---|---|---|
| page 1 | 502 rows | 502 rows |
| pages 2-6 | 502 rows, 2 repeated each | 500 rows, 0 repeated |
| 6-page totals | 3012 returned / 3002 unique | 3002 returned / 3002 unique |

- The repeated rows were the two Spanner variables, re-read at offset 0 on every page.
- Added unit tests covering the exhausted-source path for both sources; each fails against the unguarded code.


internal/server/remote/datasource.go
```diff
@@ -40,6 +40,11 @@ func (rds *RemoteDataSource) Type() datasource.DataSourceType {
 }
 
 // Id returns the id of the data source.
+//
+// Note this is not the id RemoteClient keys its pagination token entries by,
+// which is the bare remote mixer domain. The two are self-consistent today
+// because the client both writes and reads that entry, but the divergence
+// means token entries cannot be matched against DataSource.Id().
```
clincoln8, line 47: I don't fully understand what this is saying 😓 can you clarify?

clincoln8 (approved):
Thanks Keyur (and kAI)!

internal/server/remote/datasource.go
keyurva: It's basically saying that the ids of the client and the datasource are different. If we converge then it can use it to stop propagating the calls to the exhausted datasource itself at a higher level (datasources). Regardless of convergence, I'm not in favor of this coupling. So I rejected that but kept the comment.

PR #2111, opened 2026-09-14
Add 20MB size limit to API responses

n-h-diaz:
Setting up a limit to protect against potentially large responses overwhelming memory.

Currently setting the limit to 20MB - from looking at Mixer metrics, the largest endpoint (V2Observation) has P99 ~10MB, so this is a very generous limit (can be adjusted). I did see a few spikes approaching 100MB, but unclear if we'd like to support responses this big, so this warns users to make their request smaller. (Long term, we should consider paginating other endpoints.)

In the past week there have been 2070 requests over 20mb, 1982 of which are from V2 observation (the others are from the legacy V1 observation). Added some logging to better understand large requests.

This is to help prepare for limiting the unzip size to protect against decompression attacks - unzip is used by both pagination and our redis cache, so this will set an effective upper bound on what we store in redis. (This is intended last resort stop gap against large responses, as follow up we can add additional filtering at the spanner level and/or at the request level, as ideally we'd catch these sooner)


internal/util/util.go
```diff
@@ -916,3 +921,32 @@ func SortedStringKeys[V any, M ~map[string]V](m M) []string {
...
+					"sizeBytes", size,
+					"limitBytes", limit,
+				)
+				return nil, status.Errorf(codes.ResourceExhausted, maxResponseSizeErrorMsg)
+			}
+		}
+
+		return resp, nil
+	}
+}
```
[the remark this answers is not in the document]
n-h-diaz: updated the message to calculate MB limit based on input

cmd/main.go
```diff
@@ -508,6 +509,10 @@ func main() {
 			processors = append(processors, &redisProcessor)
 		}
 
+		// Response Size Limiter Processor.
+		var sizeLimiter dispatcher.Processor = dispatcher.NewResponseSizeLimiterProcessor(util.MaxResponseSize)
+		processors = append(processors, &sizeLimiter)
```
keyurva, line 514: Should we put this behind a flag in case we want to turn it down due to any issues (perf, memory, etc.)?

cmd/main.go
```diff
@@ -508,6 +509,10 @@ func main() {
 			processors = append(processors, &redisProcessor)
 		}
 
+		// Response Size Limiter Processor.
+		var sizeLimiter dispatcher.Processor = dispatcher.NewResponseSizeLimiterProcessor(util.MaxResponseSize)
```
keyurva, line 513: nit and optional: should the size be configurable?

internal/server/dispatcher/response_size_processor.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	responseSizeHints = map[RequestType]string{
+		TypeObservation: "Try requesting a specific date or reducing the number of variables or entities.",
+	}
+)
+
+// ResponseSizeLimiterProcessor checks the serialized protobuf size of the
+// response and aborts post-processing with codes.InvalidArgument if it exceeds
+// limitBytes.
```
keyurva, line 47: nit: Let's add a comment to explain why we're doing this even though we're not saving any computation cost.

internal/server/dispatcher/response_size_processor.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// PostProcess checks the size of rc.CurrentResponse and returns an
+// InvalidArgument error if it exceeds p.limitBytes.
+func (p *ResponseSizeLimiterProcessor) PostProcess(rc *RequestContext) (Outcome, error) {
+	if rc == nil || rc.CurrentResponse == nil || p.limitBytes <= 0 {
+		return Continue, nil
+	}
+
+	size := proto.Size(rc.CurrentResponse)
```
keyurva, line 70: Here's an edge condition jetski found. Feel free to ignore since it's a corner case:

> In the PR description you mentioned this sets an upper bound on what we store in Redis to prepare for limiting util.Unzip size. Note that RedisCacheClient.CacheResponse wraps response in an anypb.Any (anypb.New(response)) before marshaling and calling util.Zip, which adds ~50–60 bytes of type_url overhead on top of proto.Size(rc.CurrentResponse).<br/>
If the follow-up PR sets the util.Unzip limit to the same util.MaxResponseSize constant (20 MB), a response within ~60 bytes of 20 MB will pass this check, get written to Redis as a slightly >20 MB anypb.Any, and then fail to unzip on cache hits. Should we either leave a small buffer between MaxResponseSize and the future MaxUnzipSize, or account for the wrapper margin?

internal/server/dispatcher/response_size_processor.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// PreProcess is a no-op for ResponseSizeLimiterProcessor.
+func (p *ResponseSizeLimiterProcessor) PreProcess(rc *RequestContext) (Outcome, error) {
+	return Continue, nil
+}
+
+// PostProcess checks the size of rc.CurrentResponse and returns an
+// InvalidArgument error if it exceeds p.limitBytes.
+func (p *ResponseSizeLimiterProcessor) PostProcess(rc *RequestContext) (Outcome, error) {
```
keyurva, line 65: If the primary motivation here is bounding the payload size for zip/unzip in Redis, wondering if we should do the size check directly in the Redis cache processor (or `RedisCacheClient.CacheResponse` before `util.Zip`) and just skip caching responses over the limit instead of failing the request?

By the time `PostProcess` runs, we've already built the full response in memory anyway, so failing here breaks the ~2K requests/week over 20MB without saving the upstream Spanner/Mixer memory cost. Skipping the cache write (and logging the large request there) would still give us the upper bound for `Unzip` without breaking existing callers, while we work on catching large requests earlier in `PreProcess` or Spanner. WDYT?

keyurva (approved):
Thanks! Think about the larger size check in redis processor / client comment. OK to move whichever way you decide.

internal/server/dispatcher/response_size_processor.go
n-h-diaz: thanks this makes sense! I've updated this to avoid caching in the cache processor instead and just log, so this will not change any existing behavior but can help inform additional filtering

internal/server/dispatcher/response_size_processor.go
n-h-diaz: thanks for the call out - when unzipping I can add a small buffer to account for the overhead

internal/server/dispatcher/response_size_processor.go
n-h-diaz: updated to no longer drop response

cmd/main.go
n-h-diaz: since this switched from failing requests to just skipping the cache write (and logging), this no longer changes user-facing api behavior - large responses are just remain uncached (same as a cache miss) while saving the anypb marshaling and gzip allocations. given the low risk, keeping it always-on without a flag seems safe

cmd/main.go
n-h-diaz: kept maxCacheResponseBytes as a package-private default in redis (with newCacheProcessorWithLimit for unit tests) since nothing outside redis needs it right now, but can update to expose it via config/flags later if we'd like to curate it per environment

internal/server/redis/processor.go
```diff
@@ -60,20 +83,56 @@ func (processor *CacheProcessor) PreProcess(rc *dispatcher.RequestContext) (disp
...
+				"sizeBytes", size,
+				"limitBytes", processor.maxResponseBytes,
+				"request", formatRequestForLog(rc.OriginalRequest),
+			)
+			return dispatcher.Continue, nil
 		}
 	}
+	if skipCache(rc.Context) {
+		return dispatcher.Continue, nil
+	}
```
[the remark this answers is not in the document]
n-h-diaz: moved if skipCache(rc.Context) before the proto.Size check so we don't compute size or log a cache-skip warning when the caller already passed X-Skip-Cache: true.

kept the truncated (1KB) formatRequestForLog(rc.OriginalRequest) field in the warning log: rc.OriginalRequest only contains public query parameters (DCIDs, stat vars, dates—no headers or PII, and PreProcess on line 65 already logs rc.OriginalRequest on cache hits), and seeing which query parameters produced >20MB responses is the main signal we need to add upstream request/Spanner filtering

n-h-diaz:
Thank you Keyur for the many rounds of review/iteration on this! I think it's a much cleaner change now

PR #2112, opened 2026-09-15
Fetch all pages in node calls from agent APIs

keyurva:
Node calls from agent APIs ignored `NextToken`, so requests with large entity
lists returned empty names and types for entities past the first page. A
`get_observations` request for all countries returned 252 entities with only
154 names populated.

Updated the agent to fetch all pages for every node call. Moved the node
chunking helpers from `datasource` into `nodefetcher` so the agent can use them
through the `Mixer` facade.

## Manual verification

Same request against prod and a local server running this change, printing the
tail of the entity metadata:

```bash
BODY='{"variable_dcid":"Count_Person","entities":{"observationAbout":{"parent_dcid":"Earth","child_type":"Country"}},"date":"latest"}'

echo "=== PROD ==="
curl -s -X POST "https://api.datacommons.org/v2/agent/get_observations" \
  -H "X-API-Key: $DC_API_KEY" -H "Content-Type: application/json" -d "$BODY" \
  | jq -c '.entityMetadata.rows[250:252]'

echo "=== LOCAL ==="
curl -s -X POST "http://localhost:8081/v2/agent/get_observations" \
  -H "Content-Type: application/json" -d "$BODY" \
  | jq -c '.entityMetadata.rows[250:252]'
```

```
=== PROD ===
[["country/ZWE","",[]],["nuts/FI2","",[]]]
=== LOCAL ===
[["country/ZWE","Zimbabwe",["Country"]],["nuts/FI2","Åland",["AdministrativeArea1","Country","EurostatNUTS1","Region"]]]
```

Both return the same 252 entities in the same order, and every name prod
returned matches local. Blank names go from 98 to 0.


internal/nodefetcher/nodefetcher.go
```diff
@@ -83,6 +90,89 @@ func NodeFetchAllFunc(ctx context.Context, fetch func(ctx context.Context, req *
...
+		"chunkSize", fetchAllChunkSize,
+		"totalChunks", len(chunks),
+	)
+	responses, err := fetchChunksParallel(ctx, fetch, req, chunks)
+	if err != nil {
+		return nil, err
+	}
+
+	return mergeDisjointResponses(responses), nil
+}
```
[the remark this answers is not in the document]
keyurva: This is simply a function that was moved from datasource to here. So keeping it as-is.

internal/nodefetcher/nodefetcher.go
```diff
@@ -83,6 +90,89 @@ func NodeFetchAllFunc(ctx context.Context, fetch func(ctx context.Context, req *
...
+
+// fetchChunksParallel fetches node metadata for partitioned chunks in parallel.
+func fetchChunksParallel(
+	ctx context.Context,
+	fetch func(ctx context.Context, req *pbv2.NodeRequest) (*pbv2.NodeResponse, error),
+	req *pbv2.NodeRequest,
+	chunks [][]string,
+) ([]*pbv2.NodeResponse, error) {
+	g, groupCtx := errgroup.WithContext(ctx)
```
clincoln8, line 123: long term: we should consider setting a concurrency limit 
`g.SetLimit(N)`
and/or discuss limits on number of nodes requested in the /v2/node api

internal/agent/observations_test.go
```diff
@@ -1569,3 +1572,92 @@ func TestFetchEntityProperties_Deduplication(t *testing.T) {
...
+	sorted := slices.Clone(req.GetNodes())
+	sort.Strings(sorted)
+
+	start := 0
+	if token := req.GetNextToken(); token != "" {
+		if _, err := fmt.Sscanf(token, "offset:%d", &start); err != nil {
+			return nil, fmt.Errorf("malformed test token %q: %w", token, err)
+		}
+	}
+	end := min(start+m.nodesPerPage, len(sorted))
```
clincoln8, line 1602: **Test coverage suggestion: simulating Spanner's mid-entity edge split across pages**

`truncatingNodeMixer` is a great reproduction of Spanner returning empty `LinkedGraph{}` structs for out-of-page nodes. One subtle detail worth noting about how Spanner's `nodeEdgesToNodeResponse` (`internal/server/spanner/dsutil.go`) actually paginates:

Spanner paginates by **SQL edge offset** (`OFFSET @offset LIMIT 501`), not by entity offset. When the 500th edge lands in the middle of an entity's properties, `edges = edges[:len(edges)-1]` truncates that entity mid-node. For example, if `geoId/06037` straddles the 500-edge boundary:
* **Page 1** returns `geoId/06037` with `Arcs: {"name": ["Los Angeles County"], "typeOf": ["AdministrativeArea2"]}`
* **Page 2** (`OFFSET 500`) returns `geoId/06037` with `Arcs: {"typeOf": ["County"]}` (and no `name` arc)

Right now, `truncatingNodeMixer` slices cleanly on entity boundaries (`nodesPerPage`), so every entity is either 100% populated on Page 1 and empty on Page 2, or vice versa. While `mergeNodeResponse` in `nodefetcher.go` currently handles deep property/slice merging properly, if someone ever simplified `mergeNodeResponse` in the future to overwrite `accumulated.Data[dcid] = newGraph` whenever `newGraph.Arcs` is non-empty, this test would still pass while production queries would silently lose `name` (or partial `typeOf` values) on every boundary entity.

It would be great to have `truncatingNodeMixer` (or a test case in `nodefetcher_test.go`) split at least one boundary entity's arcs across Page 1 and Page 2 (e.g., `name` on Page 1 and `typeOf` on Page 2, or splitting two `typeOf` values across pages) to lock in that deep-merge guarantee.

clincoln8 (approved):
**Follow-up / Long-term architectural discussion (non-blocking for this PR):**

While this PR is the right immediate fix for the Agent API, having to set `fetchAllChunkSize = 200` in `nodefetcher` is a design smell that points to a broader layer leak in our storage abstraction:

1. **`v2/node` should be agnostic to backend datasource changes:**
   Per `architecture.md`, the `DataSource` interface exists to shield higher layers from database specifics. Switching the backend from Bigtable to Spanner should not have changed the effective page size or response structure of `v2/node`. Previously, Bigtable enforced a limit of 500 values per `(subject, predicate, object_type)` group, meaning a `v2/node` request for `->[name, typeOf]` across 3,000 counties returned all 3,000 names and types in a single page. Because `SpannerDataSource` replaced this with a global 500-edge limit (`MaxPageSize = 500`) across the entire query, the exact same `v2/node` request now truncates after ~100 to 165 entities and returns empty `{}` graphs for the remaining ~2,900 entities.
2. **Client-side chunking is working around a storage-layer leak:**
   Because `SpannerDataSource` changed the page semantics instead of matching the prior response contract, callers asking `v2/node` for 3,000 nodes now require 16 sequential `NextToken` round-trips (~1.3s total latency). Splitting requests into parallel chunks of 200 nodes in `nodefetcher` is a client-side workaround to parallelize what would otherwise be 16 serial Spanner queries.

**Long-term ideal architecture:**
`v2/node` and its callers (`agent.Service`, `nodefetcher`, and external API clients) should be blissfully unaware of whether the underlying datasource is Bigtable or Spanner. Callers should pass all 3,000 DCIDs in one request, and `SpannerDataSource` (or the `datasources` scatter-gather layer) should handle internal batching/parallelization to match the expected page contract without requiring higher layers to know that 200 is the magic number that fits inside a single Spanner 500-edge query.

internal/nodefetcher/nodefetcher.go
keyurva: Agree. Kai had a similar suggestion but decided to leave it out for now - reasons: current behavior does not have a limit and need to think holistically about it including in spanner tier.

internal/agent/observations_test.go
keyurva: Added a test for this. We actually hit it in this PR's repro - `country/MSR` came back from prod with a `name` but empty `typeOf`.

The overwrite case you mention is already caught by `TestNodeFetchAllFunc/MultiPage`. What wasn't covered is an entity picking up a new arc key on a later page - added `EntitySplitAcrossPages` for that.

keyurva:
> Follow-up / Long-term architectural discussion

Agree on the spanner-level paging.

On 200 being tied to spanner - it was. The old comment justified it as "200 nodes * 3 properties = 600 edges", a truncation guard. But that rationale was wrong. Spanner does emit `NextToken` correctly - the truncation happened because callers ignored it, not because chunks were too big. Now that we follow the token, each chunk pages to exhaustion, so chunk size only affects parallelism. 200 could be 50 or 500 and the results would be identical. That's why I rewrote the comment in this PR.

On BT - not sure the 3K-in-one-call behavior is an invariant or just what that request happened to hit. Can dig if useful.

Your `/v2/node` point stands and is separate from this PR - external clients asking for 3K nodes still pay the serial round-trips and the datasource layer is the right place to fix that.

For agent apis I'd go further: they shouldn't return large responses at all. Large responses overwhelm the context window and degrade agent quality however fast we fetch them. The fix there is pruning / aggregating rather than paging - what we've started discussing. Let's think more in that direction.

PR #2114, opened 2026-09-17
Adds a new resolver type to resolve on non-place-entities

shixiao-coder:
- Add `EnableNonPlaceEntityResolver` feature flag.
- Refactor `SpannerSearchConfig` to store `SearchConfigs` as a map keyed by config identifier (`indicator`, `non_place_entity`).
- Configure `non_place_entity` embedding label (`non_place_entity_experimental_embedding`) in Spanner search configuration.
- Support `non_place_entity` resolver in `v2/resolve`:
  - Validate and normalize types: `indicator` defaults to `[StatisticalVariable, Topic]`, while `non_place_entity` excludes statistical variables and topics if resolve on specific type.
- Update Spanner vector search query to dynamically apply `nodeTypeExistence` when `node_types` filter is provided, or `TRUE` when empty/unspecified.
- Add unit and golden tests for `non_place_entity` resolution and query building.


internal/server/spanner/datasource.go
```diff
@@ -510,16 +510,28 @@ func (sds *SpannerDataSource) NodeSearch(ctx context.Context, req *pbv2.NodeSear
...
+	}
 	normalizedResolveRequest, err := resolvev2.ValidateAndParseResolveInputs(req)
 	if err != nil {
 		return nil, err
 	}
 
-	if resolver := normalizedResolveRequest.Request.GetResolver(); resolver == resolvev2.ResolveResolverIndicator {
-		slog.Info("SpannerDataSource: Starting resolution", "resolver", resolver, "num_nodes", len(req.GetNodes()), "inProp", normalizedResolveRequest.InProp)
-		return sds.vectorSearchResolution(ctx, normalizedResolveRequest)
+	if resolver := normalizedResolveRequest.Request.GetResolver(); resolver == resolvev2.ResolveResolverIndicator || resolver == resolvev2.ResolveResolverNonPlaceEntity {
```
keyurva, line 528: nit: Consider abstracting this conditional into a helper method, say `useVectorSearchResolution`

internal/server/spanner/datasource.go
```diff
@@ -566,34 +578,66 @@ func loadSpannerSearchConfig(path string) (*SpannerSearchConfig, error) {
...
 	}
 
+	searchConfig, ok := cfg.SearchConfigs[spannerConfigKey]
+	if !ok {
+		return nil, fmt.Errorf("no matching SearchConfig found for spanner config key: %s", spannerConfigKey)
+	}
+
+	typeOfs, ok := validateAndNormalizeVectorSearchTypes(req.Request.GetResolver(), req.TypeOfValues)
+	if !ok {
+		return &pbv2.ResolveResponse{}, nil
```
keyurva, line 635: Should we be returning an error vs failing silently?

internal/server/spanner/datasource.go
```diff
@@ -510,16 +510,28 @@ func (sds *SpannerDataSource) NodeSearch(ctx context.Context, req *pbv2.NodeSear
...
 		return nil, err
 	}
 
-	if resolver := normalizedResolveRequest.Request.GetResolver(); resolver == resolvev2.ResolveResolverIndicator {
-		slog.Info("SpannerDataSource: Starting resolution", "resolver", resolver, "num_nodes", len(req.GetNodes()), "inProp", normalizedResolveRequest.InProp)
-		return sds.vectorSearchResolution(ctx, normalizedResolveRequest)
+	if resolver := normalizedResolveRequest.Request.GetResolver(); resolver == resolvev2.ResolveResolverIndicator || resolver == resolvev2.ResolveResolverNonPlaceEntity {
+		spannerConfigKey, ok := resolverToSpannerConfigKey[resolver]
+		if !ok {
+			return nil, status.Errorf(codes.InvalidArgument, "unsupported resolver for Spanner search: %s", resolver)
```
keyurva, line 531: can just say `"unsupported resolver: %s"`. Spanner search is an implementation detail.

internal/server/v2/resolve/resolve.go
```diff
@@ -48,6 +48,8 @@ const (
 	ResolveResolverIndicator = "indicator"
 	// ResolveResolverTopic is the resolver name for explicit topic tree navigation.
 	ResolveResolverTopic = "topic"
+	// ResolveResolverNonPlaceEntity is the resolver name for non-place entity resolution.
+	ResolveResolverNonPlaceEntity = "non_place_entity"
```
keyurva, line 52: Should this just be `non_place` inline with `place`, `indicator`, `topic`?

TBH, even `non_place` does not sound right but we can change later before launching it needed.

keyurva (approved):
Thanks Xiao!

internal/server/spanner/datasource.go
```diff
@@ -566,34 +578,66 @@ func loadSpannerSearchConfig(path string) (*SpannerSearchConfig, error) {
...
+	spannerConfigKey string,
 ) (*pbv2.ResolveResponse, error) {
 	cfg := sds.searchConfig
 	if cfg == nil {
 		return nil, status.Errorf(codes.Internal, "failed to load search config")
 	}
 
+	searchConfig, ok := cfg.SearchConfigs[spannerConfigKey]
+	if !ok {
+		return nil, fmt.Errorf("no matching SearchConfig found for spanner config key: %s", spannerConfigKey)
```
keyurva, line 630: `dcp_default.yaml` only defines `indicator` under `search_configs`. On a DCP custom instance where target defaults to `base_and_custom`, `datasources.Resolve` queries both the local `SpannerDataSource` and `RemoteDataSource` (Base DC) in an `errgroup`. Because `cfg.SearchConfigs["non_place_entity"]` is missing in `dcp_default.yaml`, the local `SpannerDataSource` will return an error here and fail the entire federated request even if Base DC returns valid candidates.

Should we either add `non_place_entity` (with `base_text_embedding`) to `dcp_default.yaml`, or have `vectorSearchResolution` log a warning and return an empty `&pbv2.ResolveResponse{}` when a `spannerConfigKey` is not configured for that instance?

Overall, it'll be good to test this on a DCP instance.

internal/server/spanner/statements.go
```diff
@@ -784,6 +786,9 @@ OR CreationTimestamp > (
 		FROM Node
 		WHERE subject_id IN UNNEST(@nodes)
 			AND EXISTS (SELECT 1 FROM UNNEST(types) t WHERE t IN UNNEST(@type_filters))`,
+	nodeTypeExistence: `EXISTS (
```
keyurva, line 789: nit: if we include the leading `\n\t\t\tAND` inside `nodeTypeExistence` (similar to how `filterDescendentStatVars` / `filterDescendentStatVarsByImport` are structured above), we can default `nodeTypeCondition` (in `query_builder.go`) to "" when `nodeTypes` is empty and avoid emitting `AND TRUE` in the final sql.

internal/server/spanner/datasource.go
shixiao-coder: Sure, using a useVectorSearchResolution with flattened condition

internal/server/spanner/datasource.go
shixiao-coder: yep, returned error

internal/server/spanner/datasource.go
shixiao-coder: updated

internal/server/v2/resolve/resolve.go
shixiao-coder: Might be `entitiy` itself is good? considering we may also embed something like developing countries in Asia, updated to non place for now

internal/server/spanner/datasource.go
shixiao-coder: I think a warning should work better, updated to empty response with warning

internal/server/spanner/statements.go
shixiao-coder: sure, updated accordingly

